PR #433, opened 2020-05-23
Fix test suite and update Travis config

tbm:


.travis.yml
```diff
@@ -8,6 +8,8 @@ addons:
       - zip
 install:
   - pip install -r requirements.txt
+  - pip install pylint
```
xuhcc, line 11: It seems that `pytest<4` is needed as well.

I suggest adding the `requirements_dev.txt` file:

```
-r requirements.txt
pylint==2.5.2
pytest==3.10.1
```

tbm: pylint is already provided in the Travis build environment for Python.  I tested the Travis file and it works.

We can add a requirements_dev.txt as another improvement and travis.yml can use that, but the patch I submitted works as is.

tbm:
Travis also needs to be enabled for beancount/beancount.

I think Martin can go to https://travis-ci.org/ to enable it.

.travis.yml
xuhcc: >We can add a requirements_dev.txt as another improvement and travis.yml can use that

Yes, then it can be used in both local and CI environments.

blais:
> Travis also needs to be enabled for beancount/beancount.
> 
> I think Martin can go to https://travis-ci.org/ to enable it.

Done.

PR #439, opened 2020-05-23
bean-query: make run custom accept query name prefixes

### LINKED ISSUE #334: prefix matching when running custom queries
**[Original report](https://bitbucket.org/blais/beancount/issue/334) by Stefano Zacchiroli (Bitbucket: [zacchiro](https://bitbucket.org/zacchiro), GitHub: [zacchiro](https://github.com/zacchiro)).**

BQL "run QUERY_NAME" command currently expects a precise match on QUERY_NAME.

It would be nice to make it select the best match for QUERY_NAME, based on prefix matching. That is, any prefix of existing queries would work, as long as it is non-ambiguous.

The attached patch implements this.

zacchiro:
as long as the given prefix unambiguously identifies a single query, running
the query will work

Closes #334

### DIFF
--- a/beancount/query/shell.py
+++ b/beancount/query/shell.py
@@ -541,14 +541,19 @@ class BQLShell(DispatchingShell):
                 print()
                 print()
         else:
-            try:
+            query = None
+            if name in custom_query_map:
                 query = custom_query_map[name]
-            except KeyError:
-                print("ERROR: Query '{}' not found".format(name))
-            else:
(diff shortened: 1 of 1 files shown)


zacchiro:
note: with this PR I'm just reviving on GH the patch that was at https://bitbucket.org/blais/beancount/issues/334/prefix-matching-when-running-custom , nothing new here other than that

PR #444, opened 2020-05-24
ignore generated flex/bison output

zacchiro:
rationale: each time make/setup.py is rerun these files changes, making the
local copy dirty. Adding them to .gitignore alone is not enough, as they have
been versioned in the past. So here we both gitignore and remove them from the
repo.


dnicolodi:
Why not do the same for the generated lexer (`lexer.c` and `lexer.h`)?

zacchiro:
> Why not do the same for the generated lexer (`lexer.c` and `lexer.h`)?

I didn't notice, because apparently they didn't get rebuilt on my machine upon make/setup. So, yeah, good point, I agree we should ignore those too. I'll update the PR accordingly.

dnicolodi:
I think the rationale is that, having those files in the repository, flex and yacc are not needed to compile beancount. I don't know how easy is to install them on Windows. Those files could be generated when the release tarball uploaded on PyPI is prepared, but this is not done at the moment, and releases have been infrequent in the past.

blais:
These four files are intended to be kept around to that users don't have to have flex/bison installed on their computer and also to minimize variation across versions. Bisons options change often and I've wanted to shield development from this occasional breakage. I've always checked them in, generated from my box, for that specific purpose, and it helps with portability.

There's little point to change them at this moment; the next version will use a Bazel build and download its own re/flex and bison third-party programs and auto-generate the files on demand.

zacchiro:
> There's little point to change them at this moment; the next version will use a Bazel build and download its own re/flex and bison third-party programs and auto-generate the files on demand.

Sounds fair. Thanks for your feedback.

PR #452, opened 2020-05-25
Fix error message in IEX price source

tbm:
Fix the following issue spotted by pylint:

    beancount/prices/sources/iex.py:35: [E1305(too-many-format-args), fetch_quote] Too many arguments for format string


tbm:
Fixed in cf76eee55bf4f536468e8fdff

PR #454, opened 2020-05-29
D function be able to parse comma number

pfrancois:
i change a bit the D function in order to be able to parse comma as decimal point in a number.
i add also some tests.


beancount/core/number.py
```diff
@@ -81,27 +87,28 @@ def D(strord=None):
     ignored, as they are assumed to be thousands separators (the French comma
     separator as decimal is not supported). This function just returns the
```
maxalbert: The remark "the French comma separator as decimal is not supported" should be removed, now that this new option is introduced.

maxalbert: Also, for users from countries where the comma is used as the decimal separator, it would be useful to clarify in the docstring what Beancount's default decimal separator is.

beancount/core/number.py
```diff
@@ -58,21 +60,25 @@ def is_fast_decimal(decimal_module):
...
 NUMBER_RE = r"[+-]?\s*[0-9,]*(?:\.[0-9]*)?"
 
-_CLEAN_NUMBER_RE = re.compile('[, ]')
+_CLEAN_NUMBER_RE = re.compile("[, ]")
+_CLEAN_NUMBER_RE_COMMA = re.compile("[. ]")
+
 
 # pylint: disable=invalid-name
-def D(strord=None):
+def D(strord=None, decimal_separator_comma=False):
```
maxalbert: Probably a matter of taste, but personally I like if the name of an option makes it unambigiously clear that it is a boolean option. How about renaming it as follows: `use_comma_as_decimal_separator=False`?

pfrancois: made the changes and updated the docstring

maxalbert: Thanks! :) Btw, I should have mentioned that I am not a Beancount developer and the decision of what to name options is down to @blais, of course. I just happened to stumble across your PR and thought I'd comment. I hope I didn't accidentally step on anyone's toes in doing so.

dnicolodi:
I think adding any form of locale support should use a locale specification rather than special cases and cascaded substitutions. There must be a package out there that offers this functionality.

Other than that, please do not touch unrelated code in your patches, and squash the commits into a coherent set of patches (in this case a single patch) to make it easier to review.

blais:
I'm reluctant to change this core function, for just the benefit of the importers. It might also slow down everything (should be evaluated).  Furthermore, this doesn't make the syntax work from the parser.

I think this should be left for v3, when a lot of other things will change. 
For now you can just make the conversions in your importer with a utility function.

matz-e:
I just added my take on this in #456, which I had sitting around for a little bit, since I parse CSV in 3 different locales. I had a previous attempt using `locale.setlocale` temporarily, but the documentation advises strongly against doing so in libraries.

blais:
Locale support may be added in v3, not even certain, would require change to parser. If you need locale support in your importers, do that in  your code, it doesn't have to be done here.

PR #456, opened 2020-06-01
Add optional locale to parse numbers.

matz-e:
Useful when parsing localized CSV files with different number grouping
and decimal points.  Can be adjusted by manually importing `babel`:

    from babel import Locale
    Locale("de_CH").number_symbols["group"] = "'"  # For ASCII imports, default is UTF-8 quote

in importers.

Also works better than setting `LC_NUMERIC` via the `locale` module (see
warnings there).

### DIFF
--- a/beancount/core/number.py
+++ b/beancount/core/number.py
@@ -12,6 +12,7 @@ About Decimal usage:
 __copyright__ = "Copyright (C) 2015-2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
+from babel.numbers import parse_decimal
 import types
 import warnings
 import re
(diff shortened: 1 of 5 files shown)


blais:
About D(): I'm sorry but D is not going to support locales in v2. Adding locale-specific support for Beancount is a much larger task which also involves changing the lexer as well. D's called about a million times in the current system and would require serious testing effort to change it (and again, it's not useful to do this without a corresponding change to the grammar which may only occur in v3).

If you need locale-specific support for your importer, do that in your code.

matz-e:
Sure thing, I'll keep this in my local tree (switching to git has made that a lot easier for me, thanks!). I didn't find too many hooks to do these modifications in the importers only. I already overwrote `get_amounts`, but the balance field is read/converted before then, and I did not see a possibility to manipulate it beforehand.

It was a nice exercise to dig a little into the locale support, I'm eager to see how you would solve that.

PR #460, opened 2020-06-04
Fix formatting issues of some docstrings

tbm:


tbm:
This is a start but I ran out of steam... and I'm not sure how useful it's going to be since you're rewriting parts of beancount.

tbm:
The CI failures are not related to this change. They already exist on master.

beancount/core/account.py
```diff
@@ -41,9 +41,9 @@ def is_valid(string):
     This does not check for the root account types, just the general syntax.
 
     Args:
-      string: A string, to be checked for account name pattern.
+      string (str): To be checked for account name pattern
```
blais: I prefer not to adopt this style for documenting types, especially as it would require changing the entire set of docstrings to be consistent. In v3 we'll be using pytype everywhere and that might change too.
Can you remove this from the PR?

blais:
This is great; I would only like to contest changes to style of the
Args/Returns adding types in parens on the left of the colon. If you can
remove those I'll merge the rest of your PR.
I remember proposing this style elsewhere and rolling that back. We'd have
to change the entire codebase (not worth it) and also with pytype and types
in v3 we'll have something better. I think I might relax the Args/Returns
style as well once we have types.

If you'd like to help adding types, that would definitely be welcome, but I
don't have a fast setup with pytype yet. I'm almost done making a Bazel
build and the next thing I need to add to it is this (globally, with a
custom flag to disable temporarily).
Finally, since I'll be deleting or moving a lot of code out to other repos
in v3, I prefer to wait until then to invest in adding types everywhere
(e..g if I'll be deleting reports there's no point wasting time decorating
it with types).

beancount/core/account.py
tbm: done

tbm:
> I prefer not to adopt this style for documenting types, especially as it would require changing the entire set of docstrings to be consistent

Yeah. I started fixing genuine formatting issues and while I was looking at the generated docs I noticed that the types were missing so started adding those...

Anyway, the current patch is formatting issues only.

PR #461, opened 2020-06-04
Fix two test suite failures

tbm:
Commit d9a70473 added a call to functools.partial() without
importing functools.


tbm:
This only leaves this pylint issue:

```
beancount/ingest/file.py:291: [C0415(import-outside-toplevel), run] Import outside toplevel (__main__)
```

I'll leave that for @blais.

blais:
Hmmm, ... about the functools imports, I ran the tests (I always run the tests), but I think what's happening is that I have an auto-imports hook for debugging in personal sitecustomize that I keep on permanently (automatically imports missing stdlibs on first attribute lookup from them and issues a warning). I never realized this jeopardizes my tests (I disabled it for lint, see ENABLE_AUTOIMPORTS= in the Makefile). I'll remove this from my environment.

PR #467, opened 2020-06-07
tools: Skip sheet_upload_test if oauth2client is not available

dnicolodi:
oauth2client is not part of benacount dependencies


blais:
I think there's another lint error (thanks for fixing the previous one).
Also, can't you just install the required dependency?

dnicolodi:
The pylint error should be fixed. `oauth2client` has a ton of dependencies so it is annoying to have to install it in virtualenvs just to run this test. I think that, in general, it is nice to be able to run the tests without the optional dependencies, and the `pytest` feature makes it trivial to support.

PR #470, opened 2020-06-11
Modernize setup.py and get rid of the cdecimal import fallback

dnicolodi:


setup.py
```diff
@@ -224,66 +95,114 @@ def get_git_changeset():
...
+
+      # Include the Emacs support for completeness, for packagers not to have to
+      # check out from the repository.
+      data_files = [
+          ('elisp', ['editors/emacs/beancount.el']),
+      ],
+
+      install_requires = [
+          # Testing support now uses the pytest module.
+          'pytest',
```
xuhcc, line 139: Could we move pytest to development requirements? As in this SO answer https://stackoverflow.com/a/28842733/1868395

blais: +1

blais:
Re. setuptools: you did see the big note I wrote in the code, right?
The workaround is to include the files in MANIFEST.in?
Here's a question: I will have no time to handle issues with this. V3 is going to be purely Bazel built (all this setup.py will go away).
If this breaks installation for some users, do you commit to stepping up and handling those?

re. Decimal: I'm fine with it, but again, if some users report uber-slow startup times, will you handle those?

Thank you,

blais:
@dnicolodi BTW your parser work hasn't been lost; it's stored in a branch.
Check out the branches if you'd like to revive it to a PR.
In V3 the entire scanner/parser gets a full rewrite, so this may or may not make sense.

setup.py
dnicolodi: Sure, this was one of the reasons for reworking `setup.py`. However, this is a behavior change and I was thinking to do it in a separate pull request. With @blais blessing it, I'll add it to this one.

dnicolodi:
> Re. setuptools: you did see the big note I wrote in the code, right?
> The workaround is to include the files in MANIFEST.in?

Yes. The `setup_test.py` passes just fine adding the files to `MANIFEST.in`. I agree that it should be handled better by `setuptools`. As we are using `MANIFEST.in` for other files to include in the source distribution, I don't see any drawback in including the C source files there too (including the *.c is redundant, but I think it makes the intention of those two lines more clear).

> Here's a question: I will have no time to handle issues with this. V3 is going to be purely Bazel built (all this setup.py will go away).

How do you `pip install beancount3` without a `setup.py`? Will this imply that all users will have to install Bazel?

> If this breaks installation for some users, do you commit to stepping up and handling those?

Sure, but I don't think this is likely.

> re. Decimal: I'm fine with it, but again, if some users report uber-slow startup times, will you handle those?

As I wrote in one of the commit messages, I think the `cdecimal` fallback was broken in a few cases: some parts of beancount already imported the Decimal class from decimal without the indirection. I don't think mixing the two implementations works. If no bugs have been reported, I think the fallback has not been used in quite some time. I hope there are no more broken Python 3 installs out there.

I commit to handle bugs introduced by these changes, or other changes I propose.

dnicolodi:
> @dnicolodi BTW your parser work hasn't been lost; it's stored in a branch.
> Check out the branches if you'd like to revive it to a PR.
> In V3 the entire scanner/parser gets a full rewrite, so this may or may not make sense.

I know, thanks. While I am looking forward to something better than flex and bison, I still think that work should land in v2. It fixes some real issues and cleans up the code a bit. Also, I think that having the parsing module accept a Python `file` object instead than a file name is much nicer and should probably be used in v3 too. I'll rebase the code to current master and open and pull request.

blais:
re. cdecimal OK. Let's see what happens.
re. setup.py OK, merging this. Hopefully nothing bad will come out of it.

re. parser accepting file object-- don't care much, I'm rewriting all the parser anyway.

re. pip install buijld: it won't be supported anymore. We'll have to build with Bazel and then construct binary artifacts (wheels?). I'm going to detail this in a couple of docs I have in progress - but essentially the core of Beancount will become more or less a C++ project, with lots of Python bindings, the query engine a more general Python-based query tool (beyond Beancount) and a bunch of other stuff is going to go away. Check out the "bazel" branch for a falvour (it's pretty much done, I'm deciding where/how to merge things over the next cuople of days).

blais:
In porting the bazel branch to this, I realized maybe the bin/ scripts aren't needed anymore for the v2 version. Is this correct? I'll still need them for v3.

dnicolodi:
> In porting the bazel branch to this, I realized maybe the bin/ scripts aren't needed anymore for the v2 version. Is this correct? I'll still need them for v3.

The installed scripts are indeed generate by setuptools, however, unless I'm missing something, they are used in `beancount/scripts/tutorial.py`. I think getting rid of them and refactoring that code to work differently would be a nice cleanup. Why do you still need them for v3?

blais:
Bazel binary targets require an entry point.

Issue #222, opened 2018-02-26
Re-opening NamedTemporaryFile is not valid on Windows


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/222) by Jeff Brantley (Bitbucket: [jsbmsu](https://bitbucket.org/jsbmsu), GitHub: [jsbmsu](https://github.com/jsbmsu)).**

----------------------------------------

Both test and production code are using `tempfile.NamedTemporaryFile`, which on non-Windows can be re-opened (by name) while the original handle is still open. For example, `beancount.utils.test_utils.docfile` decorator extracts a test's docstring, writes it to a file, and then---still holding the file open---passes the filename to the test, to be read by that test.

Searching the repo for NamedTemporaryFile turns up 10 files with "test" in the name, as well as `beancount.parser.lexer` and `beancount.utils.text_utils`.

It's unfortunate that this doesn't just work. Assuming all instances are in a `with` statement (not true, but perhaps that is a bug?), then it may be possible to workaround with a wrapper that passes `delete=False` and then manually deletes the file on `__exit__`. I'm not sure what flaws (if any) this may exhibit over the real thing.

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

Great bug find. We'll have to look at this indeed.

blais:
**Original comment by Bob Jordan (Bitbucket: [bmjjr](https://bitbucket.org/bmjjr), GitHub: [bmjjr](https://github.com/bmjjr)).**

----------------------------------------

Yes, I had to add delete=False, but I had to get it done and learn what I could about the library within a few day holiday break, so I didn't think about anything fancy like deleting the file on `__exit__`. Looks like I just made another function to clean up the temp file and call it a few places in the test suite and at the end of the tests.  


```
#!python
def cleantmpdir():
    """clean tmp directory"""

    del_dir = r'c:\tmp'
    return subprocess.Popen('rmdir /S /Q %s' % del_dir, shell=True,
                            stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

Ev2geny:
I think there was another issue open for the same problem https://github.com/beancount/beancount/issues/550

Any way, in python **3.12**   there is now [new functionality](https://docs.python.org/3/whatsnew/3.12.html), which allows to re-write the tests in a way that they will work on Windows as well: 

[...]

PR #471, opened 2020-06-12
Parser rework

dnicolodi:
This makes the parser accept a Python `file` object as input, makes the parser reentrant, and generally cleans up the parsing code. This should fix bugs related to non ASCII file names on Windows (was reported as a bug but I am not able to find it now) and help fixing #222 (I need to check if some more work is required to completely fix the issue).

### DIFF
--- a/beancount/parser/grammar.c
+++ b/beancount/parser/grammar.c
@@ -66,12 +66,9 @@
 
 
 /* First part of user prologue.  */
-#line 11 "beancount/parser/grammar.y"
+#line 49 "beancount/parser/grammar.y"
 
 
-#include <stdio.h>
-#include <assert.h>
-#include "parser.h"
 #include "grammar.h"
 #include "lexer.h"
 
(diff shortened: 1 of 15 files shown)


dnicolodi:
If you provide the code that instantiates the objects you would like the v3 parser to generate (or something that gives the feeling of how the interface would look like) I would be happy to rewrite the parsing code to work with it as an alternative to the current Python object factory.

blais:
Thanks Daniele. I'll be sending an update of what's coming soon, and rewriting some of the parser will be the first thing after that.

Here's something else in the meantime that I think would have higher impact in the parser code: convert lexer and parser bodies to C code. Let me offer this suggestion as a project.

Right now lexer.py and grammar.py are both implemented in Python. I think they are pretty static at this stage. lexer.py has little state, and most of its code could be implemented directly in C in the lexer definition itself, calling functions from the Python C API. I think that one's an obvious win. Ultimately we want to be able to delete most (all?) of lexer.py, but keep the lexer_test.py working.

For the parser (grammar.py), it's a little more complex and it would require some data structures to rewrite in C and therefore maybe it's best to wait for C++ support (which will only happen in v3, never in v2), maybe. One way to do it in C - again, without introducing C++ in v2, which we need to avoid - is to use the Python data structures from C using the Python C API. Same goals here... ultimately, complete removal of grammar.y translated to a grammar.c module would be a worthwhile goal. The grammar.y would directly call functions in grammar.c/.h instead of grammar.py.

Both of these would be valuable for two reasons:
- Immediate performance improvements that would benefit everyone.
- If I already had C versions of these codes, I could probably start the v3 parser AND maintain the v2 parser at the same time (wrapping up a v2 C functions in a v2-specific driver with the same interface as the v3 driver).

These tasks:
- Can be done incrementally (start with the lexer, separate PRs)
- Require you to work a little deeper with the Python C API (in particular its data structures). But it's easy IMO, it translates pretty closely to the Python APIs for data structures.
- Are pretty straightforward.

Let me know what you think,

blais:
We can VC if that's easier, I'm in the EST timezone

dnicolodi:
I was already wondering how much of a performance improvement could come from this approach. Having the most basic parsing nodes do not require calling into Python could have a significant performance impact and seems straightforward indeed. Working with the Python C API is not difficult, but can get tedious and getting the reference counting right in all error paths is error prone. In my experience, Cython is much easier to maintain and helps a great deal in avoiding pitfalls. How do you feel about using Cython?

dnicolodi:
A first try at implementing the first part of the plan is in #475. I haven't performed any becnhmarking to see if it has any substantial impact on performance.

PR #472, opened 2020-06-14
parser: Drop options['commodities']

dnicolodi:
The lexer updates a set with all the currencies seen during
parsing. This set is then assigned to the automatic 'currencies'
option. Despite what is reported in a comment, this information is not
used by beancount.core.getters.get_commodity_map(). Besides tests, the
only user is in the beancount.plugins.check_commodity plugin. The
plugin can however be rewritten in a simpler form relaying exclusively
on the information returned by get_commodity_map().

It may be argued that the set in the options dictionary is more efficient than recomputing the set with get_commodity_map(), however, the only user of the options calls this function anyway, so keeping the set around is actually more work than not doing it.

### DIFF
--- a/beancount/reports/price_reports.py
+++ b/beancount/reports/price_reports.py
@@ -167,8 +167,8 @@ class TickerReport(base.TableReport):
     names = ['tickers', 'symbols']
 
     def generate_table(self, entries, errors, options_map):
-        commodity_map = getters.get_commodity_map(entries)
-        ticker_info = getters.get_values_meta(commodity_map, 'name', 'ticker', 'quote')
+        commodities = getters.get_commodity_directives(entries)
+        ticker_info = getters.get_values_meta(commodities, 'name', 'ticker', 'quote')
 
         price_rows = [
             (currency, cost_currency, ticker, name)
--- a/beancount/core/getters.py
+++ b/beancount/core/getters.py
@@ -10,10 +10,7 @@ from collections import OrderedDict
 from beancount.core.data import Transaction
 from beancount.core.data import Open
 from beancount.core.data import Close
-from beancount.core.data import Balance
-from beancount.core.data import Price
 from beancount.core.data import Commodity
-from beancount.core import data
 from beancount.core import account
 
 
@@ -325,59 +322,15 @@ def get_account_open_close(entries):
     return dict(open_close_map)
 
 
-def get_commodity_map(entries, create_missing=True):
+def get_commodity_directives(entries):
     """Create map of commodity names to Commodity entries.
 
     Args:
       entries: A list of directive instances.
-      create_missing: A boolean, true if you want to automatically generate
-        missing commodity directives if not present in the output map.
     Returns:
       A map of commodity name strings to Commodity directives.
     """
-    if not entries:
-        return {}
-
-    commodities_map = {}
-    for entry in entries:
-        if isinstance(entry, Commodity):
-            commodities_map[entry.currency] = entry
-
-        elif isinstance(entry, Transaction):
-            for posting in entry.postings:
-
-                # Main currency.
-                units = posting.units
-                commodities_map.setdefault(units.currency, None)
-
-                # Currency in cost.
-                cost = posting.cost
-                if cost:
-                    commodities_map.setdefault(cost.currency, None)
-
-                # Currency in price.
-                price = posting.price
-                if price:
-                    commodities_map.setdefault(price.currency, None)
-
-        elif isinstance(entry, Balance):
-            commodities_map.setdefault(entry.amount.currency, None)
-
-        elif isinstance(entry, Price):
-            commodities_map.setdefault(entry.currency, None)
-
-    if create_missing:
-        # Create missing Commodity directives when they haven't been specified explicitly.
-        # (I think it might be better to always do this from the loader.)
-        date = entries[0].date
-        meta = data.new_metadata('<getters>', 0)
-        commodities_map = {
-            commodity: (entry
-                        if entry is not None
-                        else Commodity(meta, date, commodity))
-            for commodity, entry in commodities_map.items()}
-
-    return commodities_map
+    return { entry.currency: entry for entry in entries if isinstance(entry, Commodity) }
 
 
 def get_values_meta(name_to_entries_map, *meta_keys, default=None):
--- a/beancount/core/getters_test.py
+++ b/beancount/core/getters_test.py
@@ -200,34 +200,26 @@ class TestGetters(unittest.TestCase):
                                'Cash', 'Coffee', 'Expenses', 'Credit-Card'}
         self.assertEqual(sorted(expected_components), components)
 
-    def test_get_commodities_map(self):
+    def test_get_commodity_directives(self):
         entries, _, options_map = loader.load_string(TEST_INPUT)
-        commodity_map = getters.get_commodity_map(entries)
-        self.assertEqual({'HOOL', 'PIPA', 'USD'}, commodity_map.keys())
+        commodities = getters.get_commodity_directives(entries)
+        self.assertEqual({'HOOL', 'PIPA'}, commodities.keys())
         self.assertTrue(all(isinstance(value, data.Commodity)
-                            for value in commodity_map.values()))
-        self.assertEqual(commodity_map['HOOL'],
-                         next(entry
-                              for entry in entries
-                              if isinstance(entry, data.Commodity)))
+                            for value in commodities.values()))
 
     def test_get_values_meta__single(self):
         entries, _, options_map = loader.load_string(TEST_INPUT)
-        commodity_map = getters.get_commodity_map(entries)
-        values = getters.get_values_meta(commodity_map, 'name', default='BLA')
-        self.assertEqual({'USD': 'BLA',
-                          'PIPA': 'Pied Piper',
-                          'HOOL': 'Hooli Corp.'},
-                         values)
+        commodities = getters.get_commodity_directives(entries)
+        values = getters.get_values_meta(commodities, 'name', default='BLA')
+        self.assertEqual({'PIPA': 'Pied Piper',
+                          'HOOL': 'Hooli Corp.'}, values)
 
     def test_get_values_meta__multi(self):
         entries, _, options_map = loader.load_string(TEST_INPUT)
-        commodity_map = getters.get_commodity_map(entries)
-        values = getters.get_values_meta(commodity_map, 'name', 'ticker')
+        commodities = getters.get_commodity_directives(entries)
+        values = getters.get_values_meta(commodities, 'name', 'ticker')
         self.assertEqual({'HOOL': ('Hooli Corp.', 'NYSE:HOOLI'),
-                          'PIPA': ('Pied Piper', None),
-                          'USD': (None, None)},
-                         values)
+                          'PIPA': ('Pied Piper', None)}, values)
 
 
 if __name__ == '__main__':
--- a/beancount/loader.py
+++ b/beancount/loader.py
@@ -464,10 +464,6 @@ def aggregate_options_map(options_map, src_options_map):
         if currency not in op_currencies:
             op_currencies.append(currency)
 
-    commodities = options_map["commodities"]
-    for currency in src_options_map["commodities"]:
-        commodities.add(currency)
-
 
 def _load(sources, log_timings, extra_validations, encoding):
     """Parse Beancount input, run its transformations and validate it.
--- a/beancount/loader_test.py
+++ b/beancount/loader_test.py
@@ -538,25 +538,6 @@ class TestOptionsAggregation(unittest.TestCase):
 
             self.assertEqual({'USD', 'EUR', 'CAD'}, set(options_map['operating_currency']))
 
-    def test_aggregate_commodities(self):
-        with test_utils.tempdir() as tmp:
-            test_utils.create_temporary_files(tmp, {
-                'apples.beancount': """
-                  include "oranges.beancount"
-                  include "bananas.beancount"
-                  option "operating_currency" "USD"
-                """,
-                'oranges.beancount': """
-                  2015-12-12 open Assets:CA:Checking  CAD
-                """,
-                'bananas.beancount': """
-                  2015-12-13 open Assets:FR:Checking  EUR
-                """})
-            top_filename = path.join(tmp, 'apples.beancount')
-            entries, errors, options_map = loader.load_file(top_filename)
-
-            self.assertEqual({'EUR', 'CAD'}, options_map['commodities'])
-
 
 if __name__ == '__main__':
     unittest.main()
--- a/beancount/ops/holdings.py
+++ b/beancount/ops/holdings.py
@@ -211,11 +211,11 @@ def get_commodities_at_date(entries, options_map, date=None):
                         for holding in holdings_list}
 
     # Add in the associated ticker symbols.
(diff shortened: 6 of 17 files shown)


blais:
SGTM
You forgot one use in loader.py

grep --exclude-from=$HOME/p/conf/etc/grep-excludes -srn  options.*commodities  /home/blais/p/beancount/beancount
/home/blais/p/beancount/beancount/plugins/check_commodity.py:30:    commodities_used = options_map['commodities']
/home/blais/p/beancount/beancount/parser/grammar.py:202:        self.options['commodities'] = self.commodities
/home/blais/p/beancount/beancount/loader_test.py:558:            self.assertEqual({'EUR', 'CAD'}, options_map['commodities'])
/home/blais/p/beancount/beancount/loader.py:467:    commodities = options_map["commodities"]
/home/blais/p/beancount/beancount/loader.py:468:    for currency in src_options_map["commodities"]:

blais:
Actually, reviewing this change some more, let me turn that around, I think it's probably best to keep the full list produced from the lexer. It's the code in get_commodity_map() that might be better off checking against the full list of commodities seen during parsing. For instance, what if a commodity is present only as metadata? I think we would want an error to be issued. Fetching that full list during parsing ensures we get all of the seen commodities.

dnicolodi:
I didn't think about the "currency in metadata field" case, all other cases seem to be covered, although not in the most efficient way, by `get_commodity_map()`. The problem is that `get_commodity_map()` currently does not have access to the options dictionary. Can we change the API? Or get rid of this function all together and directly store the info returned by this call in the options dict at parsing time? The only issue I see is that `get_commodity_map()` can work work on a list of entries not coming from the parser (ie from an importer), while populating the map at parsing time does not allow this. BTW, in `experiments/prices/price-attributes.py`, `get_commodity_map()` is called with the wrong arguments...

dnicolodi:
Hi @blais , it turns out that with a simple change to the parser `get_commodity_map()` can be completely eliminated. This approach is simpler and more efficient. Let me know what you think.

dnicolodi:
Pushed a new version fixing pylint errors. I really should pick up the habit of running the pylint before pushing...

blais:
I see two problems with this proposal:

1. options["commodities"] really becomes data, not an option. Prior to this the data really lives in "entries" and having been a stickler about this has paid off in many other ways in the past. I'm worried about the slippery slope of relying on more and more non-entries data, even if structured.

2. this ties the check_commodity plugin to how the core provides a mapping of commodities. Before that it only depended on the entries. 

How much time does this save? Is it worth it adding another data to depend on?

dnicolodi:
I have the same reserves, and this maybe the main reason why my first proposal was to get rid it. But I interpreted your previous comment (have `get_commodity_map()` consume a list of commodities stored in `options["commodities"]`) as you suggesting that this is fine. Having `options["commodities"]` being a set does not make it less data-like than it being a dictionary, IMO.

I don't know how much time this saves, but `get_commodity_map()` is called in quite a few places.

The alternative is to ditch `options["commodities"]` and teach `get_commodity_map()` to look also in metadata to find all commodities mentioned in a file. I think that this would not need to be done for most call sites, but would be quite expensive when it needs to be done.

blais:
Let's measure it.

Because in a sense it's so "pure" - the whole data is just a single list - the step to changing that would have to be well motivated. How much time does it save? I'm happy to eventually consider the data to be a more complex data structure (this wouldn't go into options_map, which shouldn't become a kitchen-sink for non-options data), but that's a pretty big change that would affect a lot of the code. If it doesn't look like a big change it's because this buries data about the entries into the options map...
One could argue in some sense I was already doing that with the commodities list.

dnicolodi:
> One could argue in some sense I was already doing that with the commodities list.

And with `options['dcontext']` :)

blais:
Yep... indeed. Though that's used in booking and interpolation, not so much thereafter (other than for rendering).  Maybe for v3 I should commit to some more complex structured data?  With protos I easily could...

dnicolodi:
I had another look at this. The "good" thing is that the current form of `get_currency_map()` does a lot of work that is not necessary and can be simplified: it looks for mention of commodities in a lot of places, while all callers except the `check_commodity` plugin only care about explicit `commodity` directives. The "bad" thing is that there is no easy way to look for commodities in metadata values, as commodities are simply string objects. This requires the plugin to check the value of metadata fields against `beancount.core.amount.CURRENCY_RE`, not only their type. It may be a bit slow. I'm working on a benchmark.

dnicolodi:
The benchmark gave thumbs up: scanning a 5 years ledger generated by bean-example for all the commodities mentioned anywhere, metadata included, takes less than 20 ms. I think we can afford that to get rid of options['commodities']. @blais , please have a look if I forgot anywhere a commodity can appear in a ledger (I don't think I did).

blais:
That's still not right.  I don't think that's right. The new function does something rather different than the current one.
The current function, when called with create_missing = False, will still include the key for all used commodities, just with None values. With create_missing = True (the default), it will make sure all values of the map are Commodity directives.

There are remaining uses of get_commodity_map() which assume this. See in beancount/ops/holdings.py for example.

blais:
Hmm I'm reviewing this in much detail now, this is probably right.

blais:
A full review of impact:


/home/blais/p/beancount/beancount/core/getters.py:328:def get_commodity_map(entries, create_missing=True):
/home/blais/p/beancount/beancount/core/getters_test.py:205:        commodity_map = getters.get_commodity_map(entries)
/home/blais/p/beancount/beancount/core/getters_test.py:216:        commodity_map = getters.get_commodity_map(entries)
/home/blais/p/beancount/beancount/core/getters_test.py:225:        commodity_map = getters.get_commodity_map(entries)

This is the actual function.
(Note that you could turn it into a one-liner using a dictionary comprehension.)
I would actually rename the function to get_commodity_directives() so that
people with scripts and associated assumptions won't break with a difficult bug
to track down. I'd rather we break their script and force them to examine the
change in semantics in that way.


cd /home/blais/p/beancount && grep -srn -E 'get_commodity_map' $PWD  | grep -v Binary
/home/blais/p/beancount/beancount/reports/price_reports.py:170:        commodity_map = getters.get_commodity_map(entries)

This is fine, because get_values_meta() will get empty metadata on the
dynamically created entries.


/home/blais/p/beancount/beancount/reports/export_reports_test.py:50:            commodities_map = getters.get_commodity_map(entries)
/home/blais/p/beancount/beancount/reports/export_reports_test.py:160:        commodities_map = getters.get_commodity_map(entries)
/home/blais/p/beancount/beancount/reports/export_reports.py:139:    commodities_map = getters.get_commodity_map(entries)

Same thing here; this ends up calling get_values_meta() from classify_holdings_for_export().
Fine.


/home/blais/p/beancount/beancount/plugins/check_commodity.py:35:    commodity_map = getters.get_commodity_map(entries, create_missing=False)

This is fine too, since we call .get(..., None) on it and it's called with
create_missing = False.


/home/blais/p/beancount/beancount/plugins/merge_meta.py:58:    comm_map = getters.get_commodity_map(entries, False)
/home/blais/p/beancount/beancount/plugins/merge_meta.py:59:    ext_comm_map = getters.get_commodity_map(ext_entries, False)

This is fine, since we check for both values to not be None.


/home/blais/p/beancount/beancount/ops/holdings.py:214:    commodities_map = getters.get_commodity_map(entries)

Okay, this one is fine actually, it's called with create_missing = True, but the
resulting lookups for 'ticker' and 'quote' will produce Nones. Instead, now
it'll trigger KeyError and both ticker and quote_currency will be set to None.
Fine.


/home/blais/p/beancount/beancount/query/query_execute.py:206:    context.commodity_map = getters.get_commodity_map(entries)

I think that one is NOT correct. The problem is that the
'context.commodity_meta' is being used in the CurrencyMeta function, and since
not all the commodities are present in the map (if some directives are missing),
they will fail on this line:

    commodity_entry = context.commodity_map[args[0]]

Prior to that, it would output an empty metadata.
This one needs to be addressed if we're not going to change the behavior.


/home/blais/p/beancount/beancount/projects/export.py:74:    commodities_map = getters.get_commodity_map(entries)

Same problem as above in theory; get_commodities_table() calls this function
with the default create_missing = True, and the list of commodities produced
will exclude those without a directive. However, this only gets called if
--output_commodities is requested, and given I'm probably the only person using
that script and I'm not even using the option (it's really the final joined
table that matters), it probably won't matter that the full list isn't produced.

Fine.


/home/blais/p/beancount/experiments/prices/price-timeline-checks.py:34:    # commodity_map = getters.get_commodity_map(entries, options_map)
/home/blais/p/beancount/experiments/prices/price-attributes.py:22:    commodity_map = getters.get_commodity_map(entries, options_map)

That's fine, it's calling get_values_meta().



Okay, so all in all, it's only really affecting SQL queries that would use
CURRENCY_META() on them. If you can fix that one, we can merge this.

- Please handle the special case in query_env.CurrencyMeta: if the commodity is
  not in the map, don't fail with a KeyError, instead, return a None.

- Please rename the function to avoid mystery behavior in user scripts. Call it
  get_commodity_directives() which is pretty clear.

blais:
Another unfortunate effect of this change: your function that looks for currencies in metadata issues me errors for some of my string flags as if they were currencies (they match the CURRENCY_RE):

<check_commodity>:0:       Missing Commodity directive for 'IGNORE'
<check_commodity>:0:       Missing Commodity directive for 'PRETAX'
<check_commodity>:0:       Missing Commodity directive for 'SPDR'
<check_commodity>:0:       Missing Commodity directive for 'AFTERTAX'
<check_commodity>:0:       Missing Commodity directive for 'ROTH'

Maybe that's fine, I'm not sure. Maybe we can leave it out and add this detection back in v3. Currencies in metadata values is probably rare anyway, but in v3, the parse tree will provide different data types for currencies and strings, so it'll be fine. I would say, comment out that block and insert a comment with the spirit of this text in front of it, with a `TODO(dnicolodi)` tag.

dnicolodi:
Thanks for the detailed review. I went through the same, but I misinterpreted the code around `CurrencyMeta`, I should have included my notes in the PR or in the commit messages, but they felt too long. I'll apply the proposed changes ASAP.

dnicolodi:
I also have an half-finished commit that reports the correct file name and line number for an undeclared commodity. Now that the commodities listing is self contained in the plugin it is easy to implement. If that can be of interest I can look into finishing it.

blais:
I ran some timings; this PR actually shaves a bit of time (~4%) on my personal ledger.

About the CurrencyMeta; if you can make the change I'll merge this right away.
(And also the renaming of the function.)

About the file name reporting: that would be awesome, feel free to send in another PR, doesn't matter.

(Thank you for your contributions, these are great, I love your attention to detail!)

dnicolodi:
>  Please handle the special case in query_env.CurrencyMeta: if the commodity is not in the map, don't fail with a KeyError, instead, return a None.

Is it ok to return None there? The old code returned a metadata object...

blais:
> > Please handle the special case in query_env.CurrencyMeta: if the commodity is not in the map, don't fail with a KeyError, instead, return a None.
> 
> Is it ok to return None there? The old code returned a metadata object...

Return an empty dict. IIRC this was intended to be called with GETITEM().

No testing... I'm really looking forward to rewriting all this code from scratch and throwing this version away...

beancount/reports/price_reports.py
```diff
@@ -167,8 +167,8 @@ class TickerReport(base.TableReport):
     names = ['tickers', 'symbols']
 
     def generate_table(self, entries, errors, options_map):
-        commodity_map = getters.get_commodity_map(entries)
-        ticker_info = getters.get_values_meta(commodity_map, 'name', 'ticker', 'quote')
+        commodities = getters.get_commodity_directive(entries)
```
blais: Hmm... typo?

dnicolodi: Yes, I wonder how I managed to fat finger this between the moment in which I run the tests and the one in which I committed the changes... Fixup on its way

dnicolodi: Ah! No tests cover this code, that's why.

blais: Current tests don't pass either.

```
__________________________________________________________________________________ TestCheckCommodity.test_get_commodity_map_ex ___________________________________________________________________________________ 
                                                                                                          
self = <beancount.plugins.check_commodity_test.TestCheckCommodity testMethod=test_get_commodity_map_ex>   
entries = [Open(meta={'filename': '<string>', 'lineno': 4}, date=datetime.date(2011, 1, 1), account='Expenses:Test', currencies=...king=None), Commodity(meta={'filename': '<string>', 'lineno': 2}, date=datetime.date(2011, 1, 1), currency='BAZ'), ...]
_ = []                                                                                                    
__ = {'account_current_conversions': 'Conversions:Current', 'account_current_earnings': 'Earnings:Current', 'account_previous_balances': 'Opening-Balances', 'account_previous_conversions': 'Conversions:Previous', ...}
                                                                                                          
    @loader.load_doc()                                                                                                                                                                                              
    def test_get_commodity_map_ex(self, entries, _, __):                                                                                                                                                            
        """                                                                                                                                                                                                         
            2011-01-01 commodity BAZ                                                                      
                                                                                                                                                                                                                    
            2011-01-01 open Expenses:Test  TEST1, TEST2, TEST7                                                                                                                                                      
            2011-01-01 open Expenses:Other                                                                                                                                                                          
            2011-01-01 open Assets:Test                                                                                                                                                                             
            2011-01-01 open Assets:Investments                                                                                                                                                                      
            2011-01-01 open Equity:Opening-Balances                                                                                                                                                                 
                                                                                                                                                                                                                    
            2011-05-17 * "Something"                                                                                                                                                                                
              foo: TEST3                                                                                                                                                                                            
              Expenses:Other  1.00 TEST4                                                                  
                bar: TEST5                                                                                
              Assets:Test                                                                                                                                                                                           
                                                                                                          
            2011-05-17 * "Something"                                                                      
              Assets:Investments  5.00 FOO @ 1.00 TEST6                                                   
              Assets:Test                                                                                 
                                                                                                          
            2011-05-18 pad Assets:Test Equity:Opening-Balances                                            
            2011-05-19 balance Assets:Test -1.00 TEST7                                                    
                                                                                                          
            2011-05-20 price BAR 1.00 TEST6                                                               
        """                                                                                               
        commodities = check_commodity.get_commodity_map_ex(entries)                                       
>       self.assertEqual(commodities.keys(), {'TEST1', 'TEST2', 'TEST3', 'TEST4', 'TEST5',                
                                              'TEST6', 'TEST7', 'FOO', 'BAR', 'BAZ'})                     
E       AssertionError: dict_keys(['TEST1', 'TEST2', 'TEST7', 'BA[31 chars]AR']) != {'TEST1', 'TEST5', 'FOO', 'TEST3', 'TEST7[38 chars]BAZ'}                                                                        

beancount/plugins/check_commodity_test.py:39: AssertionError                                              

```

blais: self = <beancount.plugins.check_commodity_test.TestCheckCommodity testMethod=test_get_commodity_map_ex>

dnicolodi: Ops. Should be fixed.

blais:
Test is still failing on my end...

dnicolodi:
Sorry, I forgot to squash the commits... working on two branches at the same time...

blais:
I think you had just forgotten to add the metadata=False check.

dnicolodi:
Indeed, that hunk was left on behind

Issue #446, opened 2020-05-25
0.005 rounded down instead of up


tbm:
I reported the following issue [to ledger](https://github.com/ledger/ledger/issues/1663) but beancount also gets it wrong.  Apparently (recent) hledger do it right.

In daily financial life, 0.005 is typically rounded up to 0.01.  But:

```
2018-01-01 open Assets:Investments
2018-01-01 open Assets:Cash

2018-01-02 * "Bought FOO"
    Assets:Investments           3 FOO @ 1.595 EUR
    Assets:Cash
```

`3 * 1.595` is `4.785` which should imho be rounded to `4.79` and not `4.78`.  But:

```
bean-report d holdings | grep Assets:Cash
Assets:Cash         -4.78       EUR         EUR          -4.78         -4.78
```

Surely that should be `-4.79`.

In Python, it can be done with `decimal.ROUND_HALF_UP`:

```
>>> import decimal
>>> decimal.Decimal('0.015').quantize(D('.01'), rounding=decimal.ROUND_HALF_UP)
Decimal('0.02')
>>> decimal.Decimal('0.014999').quantize(D('.01'), rounding=decimal.ROUND_HALF_UP)
Decimal('0.01')
```

dnicolodi:
Which version of Python and which OS? On my system (macOS 10.14):
```
daniele@black:~$ python3.6 -c "from decimal import Decimal as D; print(D('0.015').quantize(D('0.01')))"
0.02
daniele@black:~$ python3.7 -c "from decimal import Decimal as D; print(D('0.015').quantize(D('0.01')))"
0.02
daniele@black:~$ python3.8 -c "from decimal import Decimal as D; print(D('0.015').quantize(D('0.01')))"
0.02
daniele@black:~$ python3.9 -c "from decimal import Decimal as D; print(D('0.015').quantize(D('0.01')))"
0.02
```

dnicolodi:
I investigated more and the above example is not the right one to demonstrate the rounding behavior. The default rounding mode is `ROUND_HALF_EVEN` which "rounds to nearest with ties going to nearest even integer" https://docs.python.org/3/library/decimal.html#decimal.ROUND_HALF_EVEN. So 0.015 rounds to 0.02 as does 0.025. I think this is the rounding rule with the least bias.

Why do you believe that the `ROUND_HALF_UP` is the expected behavior for financial applications? I've done some research and I haven't found a no clear indication that one rounding rule is universally accepted as the one to use.

[...]

PR #473, opened 2020-06-14
parser: Allow to specify commodities precision

dnicolodi:
Use metadata on `commodity` entries to explicitly set the precision to
be used when reporting amounts in the given currency, overriding the
automagic determination. Example:

```
1993-11-01 commodity EUR
  precision: 0.01
```

I need to write a few more tests for this and implement a simple optimization, but I would like to know if this is a direction in which we want to go or not before.

### DIFF
--- a/beancount/core/display_context.py
+++ b/beancount/core/display_context.py
@@ -99,9 +99,10 @@ class _CurrencyContext:
         self.has_sign = False
         self.integer_max = 1
         self.fractional_dist = distribution.Distribution()
+        self.precision = None
 
     def __str__(self):
-        fmt = ('sign={:<2}  integer_max={:<2}  '
+        fmt = ('sign={:<2}  precision={:}  integer_max={:<2}  '
                'fractional_common={:<2}  fractional_max={:<2}  '
                '"{}" "{}"')
         dist = self.fractional_dist
@@ -126,8 +127,7 @@ class _CurrencyContext:
             example_max = example + '.' + ('0' * fractional_max)
 
         return fmt.format(
-            int(self.has_sign),
-            self.integer_max,
+            int(self.has_sign), self.precision, self.integer_max,
             '_' if dist.empty() else dist.mode(),
             '_' if dist.empty() else dist.max(),
             example_common, example_max)
@@ -157,6 +157,9 @@ class _CurrencyContext:
         Returns:
           An integer for the number of fractional digits, or None.
         """
+        if self.precision is not None:
+            return -self.precision.as_tuple().exponent
+
         if self.fractional_dist.empty():
             return None
         if precision == Precision.MOST_COMMON:
@@ -183,6 +186,11 @@ class DisplayContext:
     def set_commas(self, commas):
         """Set the default value for rendering commas."""
         self.commas = commas
+
+    def set_precision(self, precision, currency='__default__'):
+        if precision is None:
+            return
+        self.ccontexts[currency].precision = precision
 
     def __str__(self):
         oss = io.StringIO()
--- a/beancount/parser/grammar.py
+++ b/beancount/parser/grammar.py
@@ -97,6 +97,23 @@ def valid_account_regexp(options):
                                                account.ACC_COMP_NAME_RE))
 
 
+def valid_precision(value):
+    """Check if the argument represents a valid precision.
+
+    Precision is specified with a Decimal representing the resolution
+    with which the amounts are reported. In practice this sets the
+    number of digits used for amout representation. To avoid
+    confusion, accept only values with mantissa equal to one: 0.1 or
+    0.01 but not 0.2.
+
+    Argument:
+      value: The value to check.
+    Returns:
+      A boolean.
+    """
+    return isinstance(value, Decimal) and value.as_tuple().digits == (1, )
+
+
 # A temporary data structure used during parsing to hold and accumulate the
 # fields being parsed on a transaction line. Because we want to be able to parse
 # these in arbitrary order, we have to accumulate the fields and then unpack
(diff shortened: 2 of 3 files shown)


blais:
No can do. Metadata is specifically designed as verboten for Beancount to interpret and is intended for users / plugins only. It's too unstructured and starting to add dependencies to any values stashed in a dictionary would get messy.

In v3's core the reports code / web code and SQL code all go away, so I think possibly the concept of precision will be located elsewhere.

dnicolodi:
I understand. However, currently options are even less structured: they can only accept string values that would need to be parsed ad-hoc. Do you have an idea about how to extend syntax to allow to attach properties to commodities? I don't think adding more fields to the directive is that nice. Should the parsing of options be extended to allow other values than strings? A fairly easy thing to do would to introduce another syntax element beside tags, links an matadata that can be attached to entries. Something like
```
1993-11-01 commodity EUR
  !precision: 0.01
  !name: "Euro"
```
looks nice to me (pick your favorite prefix character). I thought about an "inline" versions:
```
1993-11-01 commodity EUR precision: 0.01 name: "Euro"
```
but I suspect it would be confusing to tell it apart from regular metadata.

The concept of precision definitely needs to be revisited, however, it is not only required for presentation but for computation as well. For example, in the example in #446:
```
2018-01-02 * "Bought FOO"
    Assets:Investments           3 FOO @ 1.595 EUR
    Assets:Cash
```
I think the correct think to do is to complete the transaction like this:
```
2018-01-02 * "Bought FOO"
    Assets:Investments           3 FOO @ 1.595 EUR
    Assets:Cash  4.79 EUR ;; note the rounding of 4.785 to 4.79
```
because that is what is prescribed by accounting rules for the Euro: AFAIU a financial transaction can be only registered with 0.01 Euro resolution. I think that what beancount currently does (AFAIK keeping all digits till the amounts are displayed, possibly after aggregation) is not correct.

blais:
On Sun, Jun 21, 2020 at 10:54 PM Daniele Nicolodi <notifications@github.com>
wrote:

> I understand. However, currently options are even less structured: they
> can only accept string values that would need to be parsed ad-hoc. Do you
> have an idea about how to extend syntax to allow to attach properties to
> commodities? I don't think adding more fields to the directive is that
> nice. Should the parsing of options be extended to allow other values than
> strings? A fairly easy thing to do would to introduce another syntax
> element beside tags, links an matadata that can be attached to entries.
> Something like
>
> 1993-11-01 commodity EUR
>   !precision: 0.01
>   !name: Euro
>
> looks nice to me.
>
Yes, I'm doing something like this already in mine, like this:

2000-05-09 commodity EWY
  name: "iShares MSCI South Korea ETF"
  issuer: "iShares"
  assetcls: "Stocks"
  strategy: "IntlHedge"
  price: "USD:yahoo/EWY"
  export: "NYSEARCA:EWY"
  quote: USD

but only the export scripts (point being, it's not the core) is using
those. The core never assumes anything in the metadata.

There's an interesting twist coming up: I'll be removing the entire SQL
query tool and moving that to another project. In that sense, if the core
itself never renders numbers (requiring precision), having a "precision"
metadata field could be something for upstream (i.e., the query tool not
being part of the core maybe it's okay).

The concept of precision definitely needs to be revisited
>
Oh yes.


> , however, it is not only required for presentation but for computation as
> well. For example, in the example in #446
> <https://github.com/beancount/beancount/issues/446>:
>
> 2018-01-02 * "Bought FOO"
>     Assets:Investments           3 FOO @ 1.595 EUR
>     Assets:Cash
>
> I think the correct think to do is to complete the transaction like this:
>
> 2018-01-02 * "Bought FOO"
>     Assets:Investments           3 FOO @ 1.595 EUR
>     Assets:Cash  4.79 EUR ;; note the rounding of 4.785 to 4.79
>
> That's right. Currently I think I'm using the tolerance for this.


> because that is what is prescribed by accounting rules for the Euro: AFAIU
> a financial transaction can be only registered with 0.01 Euro resolution. I
> think that what beancount currently does (AFAIK keeping all digits till the
> amounts are displayed, possibly after aggregation) is not correct.
>
I agree!
There are really three concepts:
- tolerance for balance checks
- rounding for interpolation (what you describe), and
- precision for rendering numbers.
I'm not sure the names for all these are great, but in v3 we need distinct
concepts and a way to configure those by currency.

I also agree that configuring those on the Commodity directive makes most
sense.
Keep in mind that in v2 those aren't required; I believe v3 will be more
pedantic and require the full set of directives (.e.g. Commodity, Open,
Close and maybe some more).

Let's think about syntax, "!" is a little funky, what else can we do
that'll clearly not be metadata?
And if so, should we also consider that syntax to move tags and links into
that same place? I've had that in the back of my mind for a while.

dnicolodi:
> Let's think about syntax, "!" is a little funky, what else can we do that'll clearly not be metadata? 

`!` was the first character that came to mind that didn't look too much out of place, but `$`, `%`, `&` would make as much sense as `!` and I don't have any reason to argue for any of them over the others.

> And if so, should we also consider that syntax to move tags and links into that same place? I've had that in the back of my mind for a while.

I don't understand what you are suggesting here. I think using a character prefix for marking links and metadata is actually nice and concise. My reasoning in coming up with the `!` for configuration attached to entries was inspired by this.

blais:
On Mon, Jun 22, 2020 at 12:38 AM Daniele Nicolodi <notifications@github.com>
wrote:

> Let's think about syntax, "!" is a little funky, what else can we do
> that'll clearly not be metadata?
>
> ! was the first character that came to mind that didn't look too much out
> of place, but $, %, & would make as much sense as ! and I don't have any
> reason to argue for any of them over the others.
>
> And if so, should we also consider that syntax to move tags and links into
> that same place? I've had that in the back of my mind for a while.
>
> I don't understand what you are suggesting here. I think using a character
> prefix for marking links and metadata is actually nice and concise. My
> reasoning in coming up with the ! for configuration attached to entries
> was inspired by this.
>
I'd like to keep #tag and ^link working, but I think we there was some new
syntax for metadata-like non-metadata fields, I'd add support for
specifying them like that too, e.g.

2018-01-02 * "Bought FOO"
    !tag: "annual-rebalancing"
    !link: "092a099dee51ec8c3"
    Assets:Investments           3 FOO @ 1.595 EUR
    Assets:Cash  4.79 EUR ;; note the rounding of 4.785 to 4.79

The reason is that tags and links aren't really used by any of the code in
Beancount other than for filtering or grouping subsets of transactions. In
that sense, they really could be swallowed by metadata and I wouldn't be
too troubled (and the schema further simplified).

PR #475, opened 2020-06-16
Move instantation of lexer token values to compiled code

dnicolodi:
This implements the first part of what @blais suggested in #471. This build upon that pull request as the work done there makes the rewriting here much easier. Lexer errors handling keeps `lexer.py` alive. I am hesitant to move it to compiled code in the same form.


beancount/parser/Makefile
```diff
@@ -0,0 +1,4 @@
+PYTHON := python-3.7
```
blais, line 1: This file should be removed and its contents inserted in the Makefile at the root.

beancount/parser/decimal.h
```diff
@@ -0,0 +1,22 @@
+#ifndef BEANCOUNT_DECIMAL_H
```
blais, line 1: Remove this file; make the PyDecimal_IMPORT code in the initialization of the module (along with the comment), and place PyDec_FromCString in the lexer.l file./

beancount/parser/grammar.h
```diff
@@ -43,6 +43,48 @@
...
+    int first_line;
+    int first_column;
+    int last_line;
+    int last_column;
+    PyObject* file_name;
+} YYLTYPE;
+
+#define YYLTYPE_IS_DECLARED 1
+
+/* Extend defult location action to copy file name over. */
```
blais, line 65: default

beancount/parser/grammar.y
```diff
@@ -8,119 +8,135 @@
...
+    int first_line;
+    int first_column;
+    int last_line;
+    int last_column;
+    PyObject* file_name;
+} YYLTYPE;
+
+#define YYLTYPE_IS_DECLARED 1
+
+/* Extend defult location action to copy file name over. */
```
blais, line 29: default

blais:
Hey Daniele, thank you, but I started reviewing this, it's just way, way too big and it changes way  too many things at one. I just don't have that kind of available time to properly review and merge this at once.

Please break this apart in individual changes with a single focus at a time that I can merge piecemeal. Maybe I should look at each individual commit, I'm not sure, I don't have the hours available to just merge this as is and I won't merge it without a proper review.

dnicolodi:
I was already thinking to break this up as it is definitely a too long list of patches. Each patch in the series is independent of the subsequent ones (although the Github interface shows some of them in a different order...) with all tests passing at each point. Thus, in principle, patches can be applied in batches. However, having the series broken up by focus will definitely help the review.

beancount/parser/decimal.h
dnicolodi: I mostly organize the code like this because of the global variable that holds a reference to the `Decimal` type. I was too lazy to have it defined `static` in one place and `external` in an header file. But I see the point. I will rework this.

beancount/parser/Makefile
dnicolodi: I didn't do this because this Makefile is mostly an hack that work locally, but making it work reliably on a wide range of systems would be more complex. I don't know of a reliable way to get the right argument to `pkg-config` from a python executable path that is accepted by the project Makefile. I'll see if I can come up with something that does not look too complex.

dnicolodi:
I implemented a simple benchmark. Moving the lexer token semantic values generation to compiled code results is a factor ~2 speed-up in the lexing of ledgers generated by `bean-example`. I haven't measured how this impacts parsing overall yet.

blais:
I had another go at this in more detail.

This PR does the following orthogonal changes:
- Normalizing the functions from grammar.py to all pass around (filename, lineno) context
- Changes the DECREF and BUILD macros (should be two distinct and separate changes), introducing macros.h magic (needs documentation BTW).
- Adds explicit grammar rules for "account" and "currency"
- Removes token name map to use the yytranslate table instead.
- Explicitly passes the builder and scanner objects.
... I stopped here, reviewed all the fiels, there's probably another 10 different sets of changes.

This is just too much to ensure nothing broke. This looks like really great work but these changes need to be pushed independently of each other, in small increments, I cannot review this full set of changes without having a lot of blind trust in your changes. It will also make it more difficult to track down where something might have broken (silently). Also, this seems like it's changing, renaming or moving things around of a lot of various things for no particularly good reason...  you're basically coming in and changing the very core of the parser and moving a lot of other stuff around in the process.

Looking at it another way: the size and scope of this change is that which I might accept blindly if I had worked with you in the same cubicle for years and we had reviewed each other's code by working closely for a long time and had complete trust in your discipline. Can you start slower?

Are the intermediate commits in your branch tested and mergeable by themselves?
If not, can you cut this up to bite-size changes?
I want to accept the changes, but you have to realize renaming things, creating a bunch of new files, adding fancily macro optimizations, etc. are obscuring the focus of the changes. Ideally they would be very narrow and just do the one thing you're trying to do.

I'm in a bind now, I don't want to lose your changes, but I'm about to start working on the parser myself for v3... If I start working on it my changes would make it difficult for you to upgrade.

Not so sure what to do. I'd need another couple of hours to inspect this, I don't have it.
I could try to cut it up myself, but I can't, have more important changes I want to make.
I could merge it and go change what I think needs to be changed, but I'd have to do that in short order so as not to have the pending code (I could move it to a branch I suppose).

When I mentioned the lexer work, I imagine you'd do *just* that... can you isolate that work as much as possible?

dnicolodi:
> Are the intermediate commits in your branch tested and mergeable by themselves?
> If not, can you cut this up to bite-size changes?

Yes, each commit leaves the code in a good state that passes the testsuite. I tried to break up the changes in a set of patches where each patch does only one thing and the commit message is an attempt at explaining why. 

I'll break this up.

beancount/parser/lexer.l
```diff
@@ -573,3 +577,27 @@ void build_lexer_error_from_exception()
                         "Internal error: No exception");
     }
 }
+
+int PyFile_Read(PyObject *file, char *buf, size_t max_size)
```
blais, line 484: Avoid naming functions that we define ourselves using the Python C API convention, it'll get confusing. Prefer to use lowercase convention used in our C code, e.g., read_from_file().

beancount/parser/lexer.l
```diff
@@ -573,3 +577,27 @@ void build_lexer_error_from_exception()
...
+{
+     PyObject* dest = NULL;
+     PyObject* read = NULL;
+     int ret = 0;
+
+     dest = PyMemoryView_FromMemory(buf, max_size, PyBUF_WRITE);
```
blais, line 490: Hmm, on every single block read, this will (a) allocate an object, and (b) copy the buffer into it.
Right now the lexer just uses yy_scan_string() to set the buffer (there's one copy overall).
Frankly I don't think it matters... but where does this save a copy? (From the CL description)
Do you mean that instead of copying the entire string once and then scanning, the copy will only occur block-by-block?

Also, do you think there's an easy way to save and reuse the memoryview instead of recreating and deleting on every call to yyinput?

beancount/parser/parser.py
```diff
@@ -214,9 +218,12 @@ def parse_string(string, report_filename=None, **kw):
...
     if kw.pop('dedent', None):
         string = textwrap.dedent(string)
-    builder = grammar.Builder(report_filename or '<string>')
-    _parser.parse_string(string, builder, report_filename=report_filename, **kw)
-    return builder.finalize()
+    if isinstance(string, str):
+        string = string.encode('utf8')
+    if report_filename is None:
+        report_filename = '<string>'
+    file = io.BytesIO(string)
```
blais, line 225: Does this make a copy?

beancount/parser/parser.py
```diff
@@ -192,9 +193,12 @@ def parse_file(filename, **kw):
         list of errors that were encountered during parsing, and
         a dict of the option values that were parsed from the file.)
     """
-    abs_filename = path.abspath(filename) if filename else None
-    builder = grammar.Builder(abs_filename)
-    _parser.parse_file(filename, builder, **kw)
+    if file == '-':
+        file = sys.stdin.buffer
+    elif not isinstance(file, io.IOBase):
```
blais, line 197: The Args description needs to be updated for the filename arg.

beancount/parser/parser.c
```diff
@@ -162,45 +111,62 @@ PyObject* get_yylineno(PyObject *self, PyObject *args)
...
     /* Initialize the lexer. */
-    yylex_initialize(filename, encoding);
-    yyin = fp;
+    yylex_initialize(report_filename ?: "", encoding);
+    yyset_in((void*)file);
+
+    /* Initialize the parser. */
+    yy_firstline = report_firstline;
+
+    /* We need to keep those objects alive after we live this function. */
```
blais: leave

beancount/parser/parser.c
```diff
@@ -162,45 +111,62 @@ PyObject* get_yylineno(PyObject *self, PyObject *args)
...
+    Py_INCREF(file);
+    Py_INCREF(builder);
 
     Py_RETURN_NONE;
 }
 
 /* Inititalize the lexer to start running in debug mode. */
 PyObject* lexer_finalize(PyObject *self, PyObject *args)
 {
+    /* Now we can let those objects go. */
```
blais: Oh was I leaking these... *sigh* Thx

beancount/parser/parser.c
```diff
@@ -54,102 +49,56 @@ PyObject* handle_yyparse_result(int result)
...
-    yy_switch_to_buffer(yy_scan_string(input_string));
+    yylex_initialize(report_filename ?: "", encoding);
+    yyset_in((void*)file);
 
     /* Initialize the parser. */
     yy_firstline = report_firstline;
 
     /* Parse! This will call back methods on the builder instance. */
-    result = yyparse();
+    rv = yyparse();
```
blais: google style: avoid abbreviations... prefer "result" over "rv"

beancount/parser/lexer.l
dnicolodi: Indeed this implementation is not the most efficient I can think of. I implemented an optimized one but it did not make any difference in a simple benchmark. Creating the memoryview is probably very cheap (I have benchmarked this but not profiled it). What I thought would have been expensive is looking up the "readinto" method and build the argument tuple. However removing those steps did not make any difference. I'll profile it and see.

It is a long time since I wrote this commit message and I am note sure anymore about saving a copy. With the current code:

1. the docstring is copied into a `char*` by the Python function argument unpacking,
2. that is copied into the lex buffer by `yy_scan_string()`,

thus the data is copied twice. With this code:

1. we docstring is wrapped into an `io.BytesIO` object (no copy),
2. the `io.BytesIO` object is copied into the lex buffer,

thus the data is copied once. However, the docstrings are unicode and not utf8, thus there they need to be encoded in utf8 before being wrapped into the `io.BytesIO`. I don't know if Python is smart enough to elide the copy if the strings are ASCII (like most of the test data). What would surely save one copy would be have the docstrings actually being bytes objects (place a `b` in front of the docstrings used as input data for the tests).

beancount/parser/parser.py
dnicolodi: Constructing the `io.BytesIO` object (line 225) does not. Encoding the docstring in utf8 (line 222) does, unless Python is very smart.

beancount/parser/lexer.l
dnicolodi: I just checked and it is not possible to have bytes as docstrings. Thus, there are no copies saved in these cases.

PR #483, opened 2020-06-23
Enhance csv categorizer

floriskruisselbrink:
Add an optional second argument ‘row’ to the categorizer callback so the categorizer can use all the info from the csv-file that isn’t already available in the generated transaction.

Possible solution for #419

### DIFF
--- a/beancount/ingest/importers/csv.py
+++ b/beancount/ingest/importers/csv.py
@@ -3,21 +3,20 @@
 __copyright__ = "Copyright (C) 2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
+import collections
 import csv
 import datetime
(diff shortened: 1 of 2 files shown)


beancount/ingest/importers/csv.py
```diff
@@ -333,6 +331,16 @@ def get(row, ftype):
 
         return entries
 
+    def call_categorizer(self, txn, row):
```
blais, line 336: Please avoid magic. Just update the interface to require two arguments.
Clients will break; adapting their inputs should be trivial (add an argument, ignore it).

blais: I'll merge this now, will remove inspection later.

PR #484, opened 2020-06-24
Reentrant parser part 1

dnicolodi:
Hi @blais , I re-organized the commits in #471 to be more logic. Unfortunately there is not an easy way to make them independent. Thus I'll submit a few PRs in which the later ones will contain also the patches in the earlier ones. I hope this will make reviewing them easier (you just need to focus on the patches you have not seen yet). I'll rebase and force push as you merge them.

This first set of patches ("parser: Step 1 toward a reentrant parser", "parser: Step 2 toward a reentrant parser")contain only mechanical work, it should be straigthforward.

### DIFF
--- a/beancount/parser/grammar.c
+++ b/beancount/parser/grammar.c
@@ -55,7 +55,7 @@
 #define YYSKELETON_NAME "yacc.c"
 
 /* Pure parsers.  */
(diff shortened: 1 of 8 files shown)


dnicolodi:
Rebased.

PR #485, opened 2020-06-24
Reentrant parser preparatory work

dnicolodi:
This set of patches ("parser: Modify builder methods calling convention", "parser: Simplify error handling in the lexer", "parser: Extend location tracking") introduces some preparatory work to make the parser reentrant. The intent is to decouple the scanner, the grammar, and the builder classes. The only patch that introduces changes in behavior is "parser: Simplify error handling in the lexer" which changes slightly the lexer output in case of error.

### DIFF
--- a/beancount/parser/lexer.py
+++ b/beancount/parser/lexer.py
@@ -34,13 +34,16 @@ class LexBuilder:
         # A regexp for valid account names.
         self.account_regexp = re.compile(account.ACCOUNT_RE)
 
+        # A regexp for valid numbers.
+        self.number_regexp = re.compile(r'(\d+|\d{1,3}(,\d{3})+)(\.\d+)?$')
+
         # A set of all the commodities that we have seen in the file.
         self.commodities = set()
 
         # Errors that occurred during lexing and parsing.
         self.errors = []
 
-        # Default number of lines as threshold to warn over long strings.
+        # Default number of lines in string literals.
         self.long_string_maxlines_default = 64
 
     def get_invalid_account(self):
@@ -132,12 +135,7 @@ class LexBuilder:
         if '\n' in string:
             num_lines = string.count('\n') + 1
             if num_lines > self.long_string_maxlines_default:
-                # This is just a warning; accept the string anyhow.
-                self.errors.append(
-                    LexerError(
-                        self.get_lexer_location(),
-                        "String too long ({} lines); possible error".format(num_lines),
-                        None))
+                raise ValueError("String too long ({} lines)".format(num_lines))
         return string
 
     def NUMBER(self, number):
@@ -151,24 +149,10 @@ class LexBuilder:
         # Note: We don't use D() for efficiency here.
         # The lexer will only yield valid number strings.
         if ',' in number:
-            # Extract the integer part and check the commas match the
-            # locale-aware formatted version. This
-            match = re.match(r"([\d,]*)(\.\d*)?$", number)
-            if not match:
-                # This path is never taken because the lexer will parse a comma
-                # in the fractional part as two NUMBERs with a COMMA token in
-                # between.
-                self.errors.append(
-                    LexerError(self.get_lexer_location(),
-                               "Invalid number format: '{}'".format(number), None))
-            else:
-                int_string, float_string = match.groups()
-                reformatted_number = r"{:,.0f}".format(int(int_string.replace(",", "")))
-                if int_string != reformatted_number:
-                    self.errors.append(
-                        LexerError(self.get_lexer_location(),
-                                   "Invalid commas: '{}'".format(number), None))
-
+            # Check for a number with optional commas as thousands separator.
+            if not self.number_regexp.match(number):
+                raise ValueError("Invalid number format: '{}'".format(number))
+            # Remove commas.
             number = number.replace(',', '')
         return Decimal(number)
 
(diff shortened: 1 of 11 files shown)


blais:
Please update and I'll try to review today.

dnicolodi:
Rebased. There are going to be a couple more of PRs of this size or smaller. I'll try to push them out today.

blais:
BTW, I'm a little bit concerned about passing the impact of passing in (filename, lineno) on every calls to the most frequent callbacks, like amount, cost_spec, handle_list...
I get that it's clean, but I wonder about performance.
So I benchmarked it:

bergamot [git|master]:~/p/beancount$ ~/benchmark.py master reentrant-parser-preparatory
Already on 'master'
Benchmark #1: bean-check --no-cache $L
  Time (mean ± σ):      4.646 s ±  0.046 s    [User: 4.576 s, System: 0.068 s]
  Range (min … max):    4.557 s …  4.734 s    30 runs
 
Switched to branch 'reentrant-parser-preparatory'
Benchmark #1: bean-check --no-cache $L
  Time (mean ± σ):      4.691 s ±  0.154 s    [User: 4.615 s, System: 0.073 s]
  Range (min … max):    4.517 s …  5.292 s    30 runs


No problem..

dnicolodi:
I was worried it too, but it was already done for most methods anyway. I also benchmarked it and it didn't have an impact. It will also get a bit cheaper in a future patch when the filename will be stored as a Python string object instead than a char*, avoiding having to copy it at every call.

beancount/parser/lexer.py
```diff
@@ -132,12 +131,7 @@ def STRING(self, string):
         if '\n' in string:
             num_lines = string.count('\n') + 1
             if num_lines > self.long_string_maxlines_default:
-                # This is just a warning; accept the string anyhow.
-                self.errors.append(
-                    LexerError(
-                        self.get_lexer_location(),
-                        "String too long ({} lines); possible error".format(num_lines),
-                        None))
+                raise ValueError("String too long ({} lines)".format(num_lines))
```
blais, line 134: We cannot raise an exception from the parser. We need to accumulate errors for rendering later.

beancount/parser/lexer.py
```diff
@@ -151,24 +145,10 @@ def NUMBER(self, number):
...
-            else:
-                int_string, float_string = match.groups()
-                reformatted_number = r"{:,.0f}".format(int(int_string.replace(",", "")))
-                if int_string != reformatted_number:
-                    self.errors.append(
-                        LexerError(self.get_lexer_location(),
-                                   "Invalid commas: '{}'".format(number), None))
-
+            # Check for a number with optional commas as thousands separator.
+            if not self.number_regexp.match(number):
```
blais, line 149: Ditto here. We cannot raise an exception from the parser. Create an error to accumulate.

beancount/parser/lexer.py
```diff
@@ -151,24 +145,10 @@ def NUMBER(self, number):
...
-            if not match:
-                # This path is never taken because the lexer will parse a comma
-                # in the fractional part as two NUMBERs with a COMMA token in
-                # between.
-                self.errors.append(
-                    LexerError(self.get_lexer_location(),
-                               "Invalid number format: '{}'".format(number), None))
-            else:
-                int_string, float_string = match.groups()
-                reformatted_number = r"{:,.0f}".format(int(int_string.replace(",", "")))
```
blais, line 166: This was a way to check that the commas were input as per the user locale.
I'm not attached to it, you version is fine, thought I'd let you know why it was there though.

blais (changes requested):
Reviewed; LGTM apart from the two ValueError exception raised from lexer.py.
The parser should not raise any exceptions, instead all errors are to be accumulated for rendering.
So far that includes lexer errors.

If you can change those

beancount/parser/lexer.py
dnicolodi: The exception is raised here and catch in the `BUILD_LEX` where it is translated into a error accumulated like before. This was already done for "unexpected" exceptions raised by `lexer.LexBuilder` methods. Now it is done for all exceptions, simplifying error handling a bit and removing the need to be able to access the scanning location from `lexer.LexBuilder` methods, which in turn allows some decoupling of the scanner from the rest of the code. From the outside the code works exactly as before (I didn't have to modify any test). The same decoupling could be achieved  passing filename and lineno to all methods (in a similar way as what I have done for the `grammar.Builder` methods) but I thought is was a overkill.

beancount/parser/lexer.py
dnicolodi: Same explanation as above.

beancount/parser/lexer.py
dnicolodi: I don't remember now if I changed this willingly to have the parsing independent of the locale or if I didn't understand why it was done like this. I like the fact that the parsing is locale independent but I can revert to the old behavior if you prefer.

beancount/parser/lexer.py
dnicolodi: Also, this simplification comes very handy when moving the construction of token semantic values to compiled code.

blais: SGTM. We should comment in both lexer.py and on BUILD_LEX() that these are coupled.

beancount/parser/lexer.py
blais: I also prefer parsing as locale independent - but I fear this might break some clients.
Let's see what happens.

PR #489, opened 2020-06-28
Reentrant parser part 2

dnicolodi:
This concludes the refactoring of the parser code to make it reentrant. It builds on #485. I'll rebase this PR once that is merged. After these I have a few more cleanup and small improvement patches and the conversion most of `lexer.LexBuilder` into compiled code.

Commit "parser: Step 3 toward a reentrant parser" is mechanical and straightforward. Most of the work is done in commit "parser: Step 4 toward a reentrant parser", which completely changes how the C parser is exposed to Python code. I think it is a nice improvements (especially cleaning up the interface to the lexer to be a proper iterator). I don't understand the UTF-16 tests, thus commit "parser: Mark UTF-16 lexer tests as known fail". @blais , if you can explain what is the expected behavior, I can implement it. I doubt it is what the tests are testing for. Commit "parser: Allow to enable Bison debugging facility" brings back functionality that was present before and that got lost in the large refactoring. It could be squashed into "parser: Step 4..."

### DIFF
--- a/beancount/parser/parser.c
+++ b/beancount/parser/parser.c
@@ -15,191 +15,270 @@ extern YY_DECL;
 
 extern const char* getTokenName(int token);
 
-/* TODO(blais,dnicolodi): Convert those to locals eventually. */
-yyscan_t _scanner;
+/* Placeolder object for missing cost specifications. */
+PyObject* missing_obj;
 
-/* The current builder during parsing (as a global variable for now). */
-PyObject* _builder = 0;
+typedef struct {
+    PyObject_HEAD
+    yyscan_t scanner;
+    PyObject* builder;
+} Parser;
 
-/* A reference to a Python-defined constant object used as a placeholder for
-   missing cost specifications. */
-PyObject* missing_obj = 0;
+PyDoc_STRVAR(parser_doc,
+             "Parser(builder)\n"
+             "\n"
+             "Class exposing the interface to the Beancout parser C code.\n"
+             "\n"
+             "Construct a new parser object using the builder Builder object to\n"
+             "intantiate and store parsing results. This class is not intended to be\n"
+             "used directly. See the beancount.parser moduel instead.");
 
-
-PyDoc_STRVAR(parse_file_doc,
-"Parse the filename, calling back methods on the builder.\n\
-Your builder is responsible to accumulating results.\n\
-If you pass in '-' for filename, stdin is parsed.");
-
-/* Handle the result of yyparse() {459018e2905c}. */
-PyObject* handle_yyparse_result(int result)
+static PyObject* parser_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
 {
-    /* Check for internal errors during parsing (which would be the result of
-     * calling YYABORT, which we don't call), and this should raise an
-     * exception. */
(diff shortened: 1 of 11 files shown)


dnicolodi:
Tests fail because of a pylint error on master.

dnicolodi:
Rebased. I don't know why the Github interface shows the commits in the wrong order.

beancount/parser/lexer_test.py
```diff
@@ -732,13 +733,15 @@ def test_bytes_encoded_latin1(self):
...
+    @pytest.mark.xfail
     def test_bytes_encoded_utf16_invalid(self):
         utf16_bytes = self.test_utf8_string.encode('utf16')
         builder = lexer.LexBuilder()
         with self.assertRaises(SystemError):
             tokens = list(lexer.lex_iter_string(utf16_bytes, builder))
 
     # Test providing utf16 bytes to the lexer with an encoding.
+    @pytest.mark.xfail
     def test_bytes_encoded_utf16(self):
```
blais: Hmm, I don't see my two tests being different here... I think it was intended to show failure when an explicitly specified other encoding was provided. I just removed it (update).

beancount/parser/lexer_test.py
```diff
@@ -732,13 +733,15 @@ def test_bytes_encoded_latin1(self):
         self.assertEqual(self.expected_latin1_string, str_tokens[0][3])
 
     # Test providing utf16 bytes to the lexer when it is expecting utf8.
+    @pytest.mark.xfail
     def test_bytes_encoded_utf16_invalid(self):
```
blais: The intention is for the test to fail on UTF16 encoding, not accept it silently:
Looks like this:

            while 1:
>               token_tuple = _parser.lexer_next()
E               SystemError: <class 'UnicodeDecodeError'> returned a result with an error set

beancount/parser/lexer.py:225: SystemError

The error occurs in build_lexer_error(), and I just pushed a fix that will let the proper exception bubble out of the lexer (a UnicodeDecodeError). Can you make sure the exception also triggers in your version?

beancount/parser/parser.c
```diff
@@ -15,189 +15,248 @@ extern YY_DECL;
...
+typedef struct {
+    PyObject_HEAD
+    yyscan_t scanner;
+    PyObject* builder;
+} Parser;
 
-/* A reference to a Python-defined constant object used as a placeholder for
-   missing cost specifications. */
-PyObject* missing_obj = 0;
+PyDoc_STRVAR(parser_doc, "");
```
blais: Why no docstring?

beancount/parser/lexer_test.py
dnicolodi: Understood. I was a bit surprised by the kind of expected error. I can surely fix it. Actually, I think your most recent commit will fix it.

beancount/parser/parser.c
dnicolodi: Because I am lazy and I forgot to fill them in before submitting the PR. I'll add them.

dnicolodi:
Indeed the tests have been fixed by your last commit. I added the docstrings, rebased, and pushed.

PR #492, opened 2020-06-30
Reentrant parser cleanup

dnicolodi:
Some code simplification and cleanups on top of #491. These are the last patches I had from the large parser rework. The patch count is larger in the broken up PRs because I decided to split the changes differently, hopefully in an easier to review way. I actually dropped some useless changes (renames and the like...). I will push the parches with the move of the token semantic values construction to compiled code next.

### DIFF
--- a/beancount/parser/hashsrc.py
+++ b/beancount/parser/hashsrc.py
@@ -18,6 +18,7 @@ from os import path
 PARSER_SOURCE_FILES = [
     'lexer.l',
     'grammar.y',
+    'macros.h',
     'parser.h',
     'parser.c',
 ]
--- a/beancount/parser/lexer_test.py
+++ b/beancount/parser/lexer_test.py
@@ -120,9 +120,9 @@ class TestLexer(unittest.TestCase):
             ('EOL', 2, '\n', None),
             ('KEY', 2, 'abc1:', 'abc1'),
             ('COLON', 2, ':', None),
-            ('LEX_ERROR', 2, 'abc1', None),
+            ('error', 2, 'abc1', None),
             ('ACCOUNT', 2, 'ΑβγⅠ:ΑβγⅠ', 'ΑβγⅠ:ΑβγⅠ'),
-            ('LEX_ERROR', 2, 'ابجا:ابجا', None),
+            ('error', 2, 'ابجا:ابجا', None),
             ('EOL', 3, '\n', None),
             ('EOL', 3, '\x00', None)
             ], tokens)
@@ -229,7 +229,7 @@ class TestLexer(unittest.TestCase):
           2013-12-98
         """
         self.assertEqual([
-            ('LEX_ERROR', 1, '2013-12-98', None),
+            ('error', 1, '2013-12-98', None),
             ('EOL', 2, '\n', None),
             ('EOL', 2, '\x00', None),
         ], tokens)
@@ -242,7 +242,7 @@ class TestLexer(unittest.TestCase):
           2013-12-228
         """
         self.assertEqual([
-            ('LEX_ERROR', 1, '2013-12-228', None),
+            ('error', 1, '2013-12-228', None),
             ('EOL', 2, '\n', None),
             ('EOL', 2, '\x00', None),
             ], tokens)
@@ -271,7 +271,7 @@ class TestLexer(unittest.TestCase):
             ('EOL', 2, '\n', None),
             ('ACCOUNT', 2, 'Assets:99Test', 'Assets:99Test'),
             ('EOL', 3, '\n', None),
-            ('LEX_ERROR', 3, 'Assets:signals', None),
+            ('error', 3, 'Assets:signals', None),
             ('EOL', 4, '\n', None),
             ('EOL', 4, '\x00', None)
         ], tokens)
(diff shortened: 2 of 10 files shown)


dnicolodi:
I didn't test the Bazel build and it fails miserably. Working on a fix.

beancount/parser/macros.h
```diff
@@ -0,0 +1,33 @@
...
+#define _CC_FUNC_03(FN, X, ...) FN(X); _CC_FUNC_02(FN, __VA_ARGS__)
+#define _CC_FUNC_04(FN, X, ...) FN(X); _CC_FUNC_03(FN, __VA_ARGS__)
+#define _CC_FUNC_05(FN, X, ...) FN(X); _CC_FUNC_04(FN, __VA_ARGS__)
+#define _CC_FUNC_06(FN, X, ...) FN(X); _CC_FUNC_05(FN, __VA_ARGS__)
+#define _CC_FUNC_07(FN, X, ...) FN(X); _CC_FUNC_06(FN, __VA_ARGS__)
+#define _CC_FUNC_08(FN, X, ...) FN(X); _CC_FUNC_07(FN, __VA_ARGS__)
+#define _CC_FUNC_09(FN, X, ...) FN(X); _CC_FUNC_08(FN, __VA_ARGS__)
+#define _CC_FUNC_10(FN, X, ...) FN(X); _CC_FUNC_09(FN, __VA_ARGS__)
+
+#define _CC_FUNC_SEQ(_01, _02, _03, _04, _05, _06, _07, _08, _09, _10, NAME, ...) NAME
```
blais, line 15: How does this work? I keep staring at it, don't grok it.
You call it below, passing __VA_ARGS__ as the first argument, I'm assuming this gets spliced into the call to __CC_FUNC_SEQ(), and then ... well that just eats up the arguments and expands to NAME... which is...what? 
This is simultaneously magic and beautiful yet incomprehensible.
Can you add a comment explaining how this is expanded?

beancount/parser/hashsrc.py
```diff
@@ -18,6 +18,7 @@
 PARSER_SOURCE_FILES = [
     'lexer.l',
     'grammar.y',
+    'macros.h',
```
blais, line 21: Do we need to add anything to setup.py?

beancount/parser/lexer_test.py
```diff
@@ -429,13 +427,11 @@ def test_null_true_false(self, tokens, errors):
...
         TRUE FALSE NULL
         '''
-        # Note that this test contains an _actual_ newline, not an escape one as
-        # in the previous test. This should allow us to parse multiline strings.
         self.assertEqual([
             ('EOL', 2, '\n', None),
             ('BOOL', 2, 'TRUE', None),
             ('BOOL', 2, 'FALSE', None),
-            ('NULL', 2, 'NULL', None),
+            ('NONE', 2, 'NULL', None),
```
blais, line 434: Why this change? I'm calling them "NULL" in Beancount.

beancount/parser/hashsrc.py
dnicolodi: If you mean to include the new file in the distribution, I don't think so: we use a wildcard expansion in MANIFEST.in

beancount/parser/lexer_test.py
dnicolodi: The token is called NONE in the grammar definition https://github.com/beancount/beancount/blob/master/beancount/parser/grammar.y#L211 , its name is translated into "NULL" in the function that this commits drops https://github.com/beancount/beancount/blob/master/beancount/parser/grammar.y#L941 I didn't fee like this is important enough to rename the token or special case it in the `token_to_string()` function. Either can be done if you think the name NULL is more appropriate.

beancount/parser/macros.h
dnicolodi: Indeed it takes a while to see how this works, but once you see it, it is obvious. It would be easier to find an article that describes the technique, but I haven't found one in a rapid search. I'll either add some comments in the form of an explanation or a link to one.

blais: Gimme some comment on this and we're good to go. Ideally a link...

beancount/parser/lexer_test.py
blais: Fair enough.

blais:
The error from CI will show you that the quoting from your macro tricks is probably not working:

```
beancount/parser/hashsrc.py:62
  /home/travis/build/beancount/beancount/beancount/parser/hashsrc.py:62: UserWarning: The Beancount parser C extension module is out-of-date ('PARSER_SOURCE_HASH' != '86a5881b897b0b11eb4e001ca1ebd274'). You need to rebuild.
    warnings.warn(

```

beancount/parser/macros.h
dnicolodi: I didn't find a link to an explanation I liked. Thus I wrote one. Let me know if it is understandable.

dnicolodi:
> The error from CI will show you that the quoting from your macro tricks is probably not working:

That's what you get for trying to be too smart... I think a more elegant way could have been found, but I decided to keep the code simple. Please have a look.

PR #495, opened 2020-07-06
Enabled historical price fetching for Coinbase

edejong:
I added functionality to fetch historical prices from Coinbase. It's supported through a `date` parameter as seems to have been prophesied in the previous NotImplementedError.

### DIFF
--- a/beancount/prices/sources/coinbase.py
+++ b/beancount/prices/sources/coinbase.py
@@ -25,17 +25,22 @@ class CoinbaseError(ValueError):
     "An error from the Coinbase API."
 
 
-def fetch_quote(ticker):
+def fetch_quote(ticker, time=None):
     """Fetch a quote from Coinbase."""
     url = "https://api.coinbase.com/v2/prices/{}/spot".format(ticker.lower())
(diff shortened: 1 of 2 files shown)


beancount/prices/sources/coinbase.py
```diff
@@ -25,17 +25,23 @@ class CoinbaseError(ValueError):
...
+    options = {}
+    if time is not None:
+        options['date'] = time.astimezone(tz.tzutc()).date().isoformat()
+
+    response = requests.get(url, options)
     if response.status_code != requests.codes.ok:
         raise CoinbaseError("Invalid response ({}): {}".format(response.status_code,
                                                                response.text))
     result = response.json()
+    print(type(time))
```
blais: Remove forgotten trace.

blais (changes requested):

beancount/prices/sources/coinbase.py
edejong: Oops

edejong: Removed the line

PR #496, opened 2020-07-07
makefile: Do not clobber the Bazel build definition on make clean

dnicolodi:
As far as I can tell there is no reason to remove a 'build' directory
on 'make clean' and this removes the Bazel BUIL definition on my
system: macOS uses case insensitive filesystem by defaul. Yes, it
sucks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ all: build
 # Clean everything up.
 clean:
 	rm -f core
-	rm -rf build
 	rm -f $(CROOT)/grammar.h $(CROOT)/grammar.c
 	rm -f $(CROOT)/lexer.h $(CROOT)/lexer.c
 	rm -f $(CROOT)/*.so


Makefile
```diff
@@ -15,7 +15,6 @@ all: build
 # Clean everything up.
 clean:
 	rm -f core
-	rm -rf build
```
blais, line 18: The setup.py build creates this directory and grepping through the code recursively you'll get a bunch of hits from the copies of the source that are made there... it's a bit of a pain, that's why I delete it.

Now... can we delete it conditionally if some specific file in its contents is present? That would work

dnicolodi: I use `setup.py develop` so I think I don't run into this, or I missed it completely. Actually I think that spelling it `rm -rf build/` may be enough. Let me test and come back

dnicolodi: I was wrong, `setup.py build_ext` creates the `build` directory and I have a name clash if I don't remove the `BUILD` file before. This is unfortunate. I'll research a different solution.

blais:
There might be a better solution: I think bazel honors BUILD and BUILD.bazel files.
Check it out.

PR #498, opened 2020-07-07
Preparatory work to move construction of the tokens semantic values to compiled code

dnicolodi:
The string literals parsing changes allows to avoid a copy for string that do not contain escape sequences. This optimization is not implemented yet.


beancount/parser/parser.py
```diff
@@ -122,14 +122,8 @@
...
-from beancount.parser.grammar import ParserError
-from beancount.parser.grammar import ParserSyntaxError
-from beancount.parser.grammar import DeprecatedError
-
-
-# disable pyflakes
-# pylint: disable=pointless-statement
-ParserError, ParserSyntaxError, DeprecatedError
+# pylint: disable=unused-import
+from beancount.parser.grammar import ParserError, ParserSyntaxError, DeprecatedError
```
blais: Google style: one symbol per line. Please revert that hunk.

beancount/scripts/doctor_test.py
```diff
@@ -30,29 +30,29 @@ def test_dump_lexer(self, filename):
...
+            ACCOUNT           2 b'Expenses:Restaurant'
+            EOL               3 b'\\n'
+            DATE              3 b'2013-01-01'
+            OPEN              3 b'open'
+            ACCOUNT           3 b'Assets:Cash'
+            EOL               4 b'\\n'
+            EOL               5 b'\\n'
+            DATE              5 b'2014-03-02'
+            ASTERISK          5 b'*'
+            STRING            5 b'"Something"'
```
blais, line 45: Nice.

beancount/parser/parser.c
```diff
@@ -223,7 +223,7 @@ static PyObject* parser_iternext(Parser* self)
     }
 
     /* Yield a (token name, line, matched string, token value) tuple. */
-    return Py_BuildValue("(sis#O)",
+    return Py_BuildValue("(siy#O)",
```
blais, line 227: Why change the iterator to return bytes?

beancount/parser/grammar.y
```diff
@@ -409,22 +410,28 @@ price_annotation : incomplete_amount
                      $$ = $1;
                  }
 
-posting : INDENT optflag ACCOUNT incomplete_amount cost_spec eol
+account : ACCOUNT
```
blais, line 410: There are a lot of accounts... this is adding another call for each IIUC.
Did you benchmark the difference?

beancount/parser/parser.c
dnicolodi: Because Flex `yyget_text()` returns a view into the lexer buffer, and those are bytes. This has been working as is so far because the lexer was not returning the matched text for string literals. In a the next commit this will change and thus this will break if the strings are not UTF-8 (and there are tests checking that other encodings are accepted). The easiest fix is to return byte objects here. Another possibility is to accept UTF-8 only, but you are worried about breaking compatibility there. Returning strings would require decoding the matched text for string literals and other token differently. Returning bytes is also (marginally) faster.

beancount/parser/grammar.y
dnicolodi: Combined with the changes in the other PR, the call is moved from the lexer to the grammar, thus there should be no performance penalty. I haven't benchmarked this change alone. With both PRs applied, parsing gets 50% faster IIRC.

dnicolodi:
Fixed the one import symbol per line thing.

beancount/parser/lexer.l
```diff
@@ -388,41 +342,28 @@ NULL		{
...
+
+    r = cunescape(yytext + 1, yyleng - 2, false, &unescaped);
+    if (r < 0) {
+	build_lexer_error(yylloc, builder, "invalid string escape.", 22);
+	yylval->pyobj = Py_None;
+	Py_INCREF(Py_None);
+	return YYerror;
     }
 
-    /* Escape sequences. */
```
blais, line 416: Do these still work when seen within strings?

beancount/parser/lexer.l
```diff
@@ -552,43 +491,68 @@ void yylex_initialize(PyObject* file, PyObject* filename, int lineno, const char
...
-    assert(b->ptr != NULL);
-
-    size_t size = b->end - b->buf + 1;
-    size_t cur_size = b->ptr - b->buf;
-    size_t new_size = cur_size + num_new_chars;
+    const char* src;
+    char* buffer;
+    char* dst;
+
+    /* Make sure we hace space for the string terminator. */
```
blais: have

beancount/parser/lexer.l
```diff
@@ -552,43 +491,68 @@ void yylex_initialize(PyObject* file, PyObject* filename, int lineno, const char
     extra->n_line_tokens = 0;
     extra->line = lineno;
     extra->encoding = encoding ? encoding : "utf-8";
-    buffer_init(&extra->str, 1024);
 
     Py_XDECREF(yyget_in(scanner));
     Py_INCREF(file);
     yyrestart((void *)file, scanner);
     yybegin(scanner);
 }
```
blais, line 499: Please add a documentation comment for this function, in particular describing the special error codes it can return.

beancount/parser/lexer.l
```diff
@@ -552,43 +491,68 @@ void yylex_initialize(PyObject* file, PyObject* filename, int lineno, const char
...
+        if (*src != '\\') {
+            *dst = *src;
+            dst++;
+            continue;
+        }
 
-    while (size < new_size) {
-        size <<= 1;
+        if (string + len - src < 2)
+            return -EINVAL;
```
blais, line 532: If you return, you're leaking buffer.

beancount/parser/lexer.l
```diff
@@ -552,43 +491,68 @@ void yylex_initialize(PyObject* file, PyObject* filename, int lineno, const char
...
+            break;
+        case 'f':
+            *dst++ = '\f';
+            break;
+        case 'b':
+            *dst++ = '\b';
+            break;
+        default:
+            if (strict)
+                return -EINVAL;
```
blais, line 559: Ditto, returning an error here leaks the buffer.

beancount/parser/lexer.l
blais: Also please add comment about ownership of returned buffer.

beancount/parser/lexer.l
```diff
@@ -388,41 +342,28 @@ NULL		{
...
     }
 
-    /* Escape sequences. */
-    \\n       buffer_push(strbuf, '\n');
-    \\t       buffer_push(strbuf, '\t');
-    \\r       buffer_push(strbuf, '\r');
-    \\b       buffer_push(strbuf, '\b');
-    \\f       buffer_push(strbuf, '\f');
-    \\(.|\n)  buffer_push(strbuf, yytext[1]);
+    PyObject* string = PyUnicode_Decode(unescaped, r, yy_encoding, "ignore");
```
blais, line 353: I think you forgot to free "unescaped".

blais:
I like the removal of a state in the lexer, but note that your solutions causes a malloc() for each string - and there are many strings in these files - while the previous one reused the buffer. I'm not sure how much of a difference that'll make. Given that this is single-threaded, can you find a way to reuse the buffer?

dnicolodi:
I thought about it and decided it was not worth optimizing it. The strings are mostly short and should hit the fast path in `malloc`. Since the unescaped is at most as long as the escaped one (if there are no escape characters) the easy thing would be to simply override the input string. However, doing so we loose the possibility to return the raw input in `lexer.lex_iter()` (I implemented it like that at the beginning, thus the obvious bugs you found). Reusing a buffer would require bringing back most of the buffer management code this removed.

blais:
SGTM.

dnicolodi:
I run a quick benchmark lexing 10 times 10 years of transactions generated by bean-example.

These are the timings with the code as in this PR (after fixing the bugs you found):
```
[1.070451189, 1.0652146020000002, 1.0898541430000002, 1.0894032449999997, 1.1069556669999994]
```
and these are replacing the malloc with a static buffer big enough for all the strings in there:
```
[1.0353750679999996, 1.0699426470000004, 1.0674283330000005, 1.0733579679999998, 1.0852542340000007]
```
If there is a difference, it is negligible.

dnicolodi:
Updated and force pushed.

blais:
BTW you can use hyperfine for benchmarks in the future, it's pretty neat.

dnicolodi:
> BTW you can use hyperfine for benchmarks in the future, it's pretty neat.

Those numbers are form `timeit.repeat()`. I had to write a script to do the lexing only anyway, using`timeit.repeat()` didn't take much longer. I'll look at hyperfine tho.

Give me one minute to rebase #505.

PR #502, opened 2020-07-08
CI fix

dnicolodi:
The most recent Ununtu release supported by Travis-CI is Focal which
ships with Bison 3.5. We now require Bison 3.6. This worked because
the Makefile decided not to rebuild the checked in generted files when
the commit requiring Bison 3.6 was introduced. When the Makefile
attempts to regenerate the grammar it fails.

Install a more recent Bison as part of the CI setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ python:
 
 addons:
   apt:
-    sources:
-      - ubuntu-toolchain-r-test
     packages:
       - gnupg
       - zip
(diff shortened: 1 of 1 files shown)


blais:
Hmm we shouldn't have to do that. The v2 build uses the pregenerated files checked into the repository.

dnicolodi:
That is what was done before, thus I was trying to fix it properly. If using the generated files is OK, I can drop this experiment. v2 is built and tested with Bazel and the simple Makefile, thus the Bison and Flex generation is tested anyhow, I believe.

PR #506, opened 2020-07-11
Remove some abandoned code

dnicolodi:
Probably from an attemot to better support UTF-8 input files.


blais:
Could you instead move all those files to a new "tools/utf8/" directory for preservation?
I feel there's a very slim change this might come in handy in the future.

dnicolodi:
The files in `beancount/parser/lexer/` are generated by `generate_unicode_category_regexps.py` and currently unused and not updated (the Makefile rules that generate and uses them is commented out). I think all this is not relevant switching to RE/flex and I don't feel like we want to start making use of this in v2. I can leave the script alone and just remove the generated files and the obsolete Makefile rules.

blais:
Leaving the script alone and removing the generated files SGTM.
Thanks

dnicolodi:
Done. I piled some more cleanups on top, they are all fairly uneventful.

PR #507, opened 2020-07-16
Adding US Thrift Savings plan as a price source.

jmoles:
The US thrift savings plans (TSP) prices aren't posted like a typical stock (say like through Yahoo! Finance API). With the roll out of the new TSP web site, they seem to have an API to get the data that is on a user interface here (https://www.tsp.gov/fund-performance/share-price-history/). This adds a price source to get the latest and historic prices from the TSP web site.


jmoles:
Fixed the issues that were causing Travis CI to fail. Should be good for review now.

PR #511, opened 2020-07-25
makefile: Tiny cleanup

dnicolodi:
The ENABLE_AUTOIMPORTS environment variable does not seem to have any
effect on the execution of pylint. If it does it is not documented
anywhere.


blais:
Yeah, I used to have autoimports in my personal setup and I disabled it eventually because it messes with pylint and now with something else.

PR #514, opened 2020-07-27
Cast float to string before casting to Decimal

dumbPy:
float to Decimal gives approximation errors as below example
 
But casting it to `str` and then to `D`/`Decimal` would solve this problem I suppose.

```python
In [1]: from beancount.core.number import D                                                                                                                                                    

In [2]: D(22.5)                                                                                                                                                                                
Out[2]: Decimal('22.5')

In [3]: D(22.50)                                                                                                                                                                               
Out[3]: Decimal('22.5')

In [4]: D(22.52)  # these shouldn't exist                                                                                                                                                                               
Out[4]: Decimal('22.519999999999999573674358543939888477325439453125')

In [5]: D(str(22.52)) # cast to str first                                                                                                                                   
Out[5]: Decimal('22.52')

In [6]: D(22.52)/D(11.26) # I see the operations have already been handled                             
Out[6]: Decimal('2')
```

I understand that the `D` documentation clearly mentions about passing strings. But since `D` is accepting float, I think it should be handled here.

### DIFF
--- a/beancount/core/number.py
+++ b/beancount/core/number.py
@@ -55,8 +55,10 @@ def D(strord=None):
             return Decimal(_CLEAN_NUMBER_RE.sub('', strord))
         elif isinstance(strord, Decimal):
             return strord
-        elif isinstance(strord, (int, float)):
+        elif isinstance(strord, int):
             return Decimal(strord)
+        elif isinstance(strord, float):
+            return Decimal(str(strord))
         else:
             assert strord is None, "Invalid value to convert: {}".format(strord)
     except Exception as exc:


dnicolodi:
The rounding error is introduced in the float representation, not in the conversion to Decimal. 22.52 cannot be represented exactly as a float. Converting the float to a string only introduces some extra rounding that, in this specific case, happens to give you the result you expect. Floats should not be used when representing exact fractional numbers.

blais:
Yes I agree with Danielle casting to string is not the right thing to do.
Perhaps disallowing (by raising an exception) would be more of a guardrail but I see no reason to protect people from misuse if the decimal library doesn't

dumbPy:
Decimal is a standard library and still explain this issue in the documentation. Since it's beancount (presumably) that stringifies the Decimal object while writing it to file, I think the onus of handling this issue is on beancount and not on Decimal. I think we should at least be issuing a warning for this if not an exception.

blais:
@dumbPy This is not a Beancount issue, this is a fundamental representation issue.
We could constrain the creation from floats, but then we deviate from the Decimal implementation.
But let's not do this, because eventually D() will probably go away in favor of just using vanilla Decimal.

PR #515, opened 2020-07-28
lexer: Fix subtle bug

dnicolodi:
This typo would cause yy_skip_line() to loop forever if it does not encounter newline before the end of the input.


dnicolodi:
I introduced the bug when I rewrote the function into a macro during the refactoring of the lexer code. Sorry!

blais (approved):

Issue #192, opened 2017-09-13
Make it possible to use single letter as commodity name


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/192) by Zhuoyun Wei (Bitbucket: [wzyboy](https://bitbucket.org/wzyboy), GitHub: [wzyboy](https://github.com/wzyboy)).**

----------------------------------------

Hi Martin,

this is a similar issue to #138. I recently found out that Beancount currently does not allow single-letter commodity name. A minimal reproducible example:

```
1970-01-01 commodity USD
1970-01-01 commodity V

2017-09-01 open Assets:SingleLetter:X
2017-09-01 open Assets:Trade:Cash USD
2017-09-01 open Assets:Trade:Positions "FIFO"

2017-09-13 * "Buy 10 shares of Visa"
  Assets:Trade:Cash                         -1000.00 USD
  Assets:Trade:Positions                      +10.00 V { 100.00 USD }
```

A bean-check shows:

```
single-letter.beancount:2:       Invalid token: 'V'
single-letter.beancount:10:      Invalid token: 'V'
```

It seems that single-letter account names do not trigger errors any more (line 4) since #138 is resolved, but single-letter commodity names are still not allowed (line 2 and line 10).

Could you please make single-letter commodity possible? I would like to make pull request but I do not know how to write C code...

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

Unfortunately this won't be possible. This would result in ambiguity between the flag - that character which comes right after a date - and a currency. Add a _ or other characters to your currency.

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

Not feasible.

blais:
**Original comment by Zhuoyun Wei (Bitbucket: [wzyboy](https://bitbucket.org/wzyboy), GitHub: [wzyboy](https://github.com/wzyboy)).**

----------------------------------------

@martin Sorry but I am a little bit confused: in which case does a commodity come right after a date?

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

[...]

Issue #516, opened 2020-07-28
Handling of flags in the lexer


dnicolodi:
The lexer recognizes the `!&#?%PSTCURM` characters as flags. Despite the comment above the code this is a different set of characters than the one in `beancount.core.flags`, which is restricted to `*!PSTCURM`. Should the two be unified?

The lexer rule includes `#` but it cannot ever match as `#` is matched by a previous rule that return the HASH token. Thus, for clarity, should be removed from the FLAG rule.

The current code also ignores any line that starts with a character in this character set. I see where this is coming from in the history of the parser, although I think it is a fairly surprising behavior (why should lines starting with `P` be allowed, but not lines starting with `A`?) Should it be removed? The handling of lines starting with `#` is doubly strange, as they are ignored but only if the character following `#` cannot be part of a tag, in which case they are parsed as tags (but rejected in the grammar, I guess).

cgevans:
There should be no case where a `#` can only be distinguished as a flag, line-skip-indicator, or non-flag (eg, beginning of a tag or token in a compound cost) on the basis of what comes after it, right? As far as I can tell, flags should only be possible after a DATE or after an INDENT, and no other use of `#` should be valid there.  Similarly, there should be no possibility of having a tag at the beginning of a line, right?

PR #517, opened 2020-07-28
Modify lexer to allow single letter currencies (patch from #192)

cgevans:
I still need to check that this works after rebasing, and add tests, which is why I am marking it as a draft.

### DIFF
--- a/beancount/parser/lexer.l
+++ b/beancount/parser/lexer.l
@@ -166,7 +166,6 @@ int pyfile_read_into(PyObject *file, char *buf, size_t max_size);
         yylloc->last_column = yycolumn + yyleng - 1;                    \
         yylloc->file_name = yy_filename;                                \
         yycolumn += yyleng;                                             \
-        if (YY_START == AFTERDATE) BEGIN(INITIAL);                      \
     }
 
 /* Skip the rest of the input line.  This needs to be implemented as a
@@ -197,8 +196,8 @@ int strtonl(const char* buf, size_t nchars);
 /* Exclusive start condition for parsing escape sequences in string literals. */
 %x STRLIT
 
-/* A start condition triggered when a date is found, to allow flags to be recognized. */
-%s AFTERDATE
+/* A start condition triggered when a date or indent is found, to allow flags to be recognized. */
+%x POSSIBLEFLAG
 
 
 ASCII           [\x00-\x7f]
@@ -222,9 +221,7 @@ ACCOUNTNAME     ([A-Z0-9]|{UTF-8-ONLY})([A-Za-z0-9\-]|{UTF-8-ONLY})*
     return EOL;
 }
 
-<AFTERDATE>[ \t\r]+ {
-    BEGIN(AFTERDATE);
-}
+
 
  /* Ignore whitespace, except when found at the beginning of a line
     and followed by a regular character. This is how we detect an
@@ -239,6 +236,7 @@ ACCOUNTNAME     ([A-Z0-9]|{UTF-8-ONLY})([A-Za-z0-9\-]|{UTF-8-ONLY})*
             return SKIPPED;
         }
         else {
+            BEGIN(POSSIBLEFLAG);
             return INDENT;
         }
     }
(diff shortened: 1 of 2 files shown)


beancount/parser/lexer.l
```diff
@@ -218,6 +222,10 @@ ACCOUNTNAME     ([A-Z0-9]|{UTF-8-ONLY})([A-Za-z0-9\-]|{UTF-8-ONLY})*
     return EOL;
 }
 
+<AFTERDATE>[ \t\r]+ {
+    BEGIN(AFTERDATE);
+}
```
dnicolodi: This looks funny and it seems to be required only because the `BEGIN(INITIAL)` added to `YY_USER_ACTION`. I think that getting rid of of both these would make the thing more understandable. I think the `BEGIN(INITIAL)` in `YY_USER_ACTION` can be placed in a default rule for the `<AFTERDATE>` state instead.

beancount/parser/lexer.l
```diff
@@ -166,6 +166,7 @@ int pyfile_read_into(PyObject *file, char *buf, size_t max_size);
         yylloc->last_column = yycolumn + yyleng - 1;                    \
         yylloc->file_name = yy_filename;                                \
         yycolumn += yyleng;                                             \
+        if (YY_START == AFTERDATE) BEGIN(INITIAL);                      \
```
dnicolodi: The plan is to move the lexer to RE/flex where `YY_USER_ACTION` should go away. I would avoid to add more to it. Also, having a state transition in it does not make the mechanism really obvious. See below.

beancount/parser/lexer.l
```diff
@@ -301,15 +309,13 @@ ACCOUNTNAME     ([A-Z0-9]|{UTF-8-ONLY})([A-Za-z0-9\-]|{UTF-8-ONLY})*
...
-        return SKIPPED;
-    }
+<AFTERDATE>[!&#?%PSTCURM]  {
+    yylval->character = yytext[0];
+    return FLAG;
+}
+^[!&#?%PSTCURM]    {
+    yy_skip_line();
+    return SKIPPED;
 }
```
dnicolodi, line 287: I like this. But I think `*` should be added to the flag characters. Also, it requires adjusting the grammar too: in a few places looks for an `HASH` or `ASTERISK` token (see the `txn` grammar rule).

dnicolodi (commented):
How is this handling flags that can be prepended to postings? Like
```
2020-07-28 * "Test"
  Expenses:Foo  1.00 USD
  C Assets:Cash
```

beancount/parser/lexer.l
cgevans: Looking into this, `*`, `#`, and `:` were all handled separately before this line, with their own beginning-of-line checks. In fact, it's not clear to me that the previous code *ever* triggered for `#`.  Adding all three to this arrangement seems like it would be fine: I don't think there are places where any of them are used after a DATE, or after an INDENT, to mean anything other than a FLAG.  Then rules for separate HASH, ASTERISK, and COLON tokens can be moved to after this handling, with the separate beginning-of-line check removed.

However, I'm a bit uncomfortable here, because looking through the grammar, it appears that HASH is actually used in what appears to be an undocumented but implemented compound cost statement, and ASTERISK is used in an undocumented and apparently not yet implemented cost merging statement.  While I think my new changes preserve these behaviours, I'm hoping that they don't break any other potential behaviour I'm not aware of.

beancount/parser/lexer.l
```diff
@@ -382,7 +389,7 @@ NULL		{
 
  /* Currencies. These are defined as uppercase only in order to disambiguate the
   * syntax. This is kept in sync with beancount.core.amount.CURRENCY_RE. */
-[A-Z][A-Z0-9\'\.\_\-]{0,22}[A-Z0-9]	{
```
cgevans, line 385: I actually need to add one to the maximum length here in order to preserve the 24 character limit that is noted in the documentation.  But is there a reason for that limit?

beancount/parser/lexer.l
cgevans: Hmm... there seems to be a wider problem here.  `beancount.core.flags` has `[*!PSTCURM]` as valid flags, as I now notice you pointed out in #516.

cgevans:
> How is this handling flags that can be prepended to postings? Like
> 
> ```
> 2020-07-28 * "Test"
>   Expenses:Foo  1.00 USD
>   C Assets:Cash
> ```

It doesn't, because I completely forgot about that.  The commit fixes this.  However, I need to have Bison 3.6 installed to test the grammar changes, which will take some time.

beancount/parser/lexer.l
blais: "#" is used in compound costs, to specify unit and total cost, as {unit # total currency}.

blais:
A bit reluctant to make this change in v2 and break things (unless I know you'll be there to support it); I think it may be wise to do this in v3 along with the RE/flex conversion.

blais:
We want to fix this more generally in v3, with a scanner rewrite. I'm really reluctant to change the v2 lexer's state at this moment.
Let's hold off this feature for v3.

beancount/parser/lexer.l
cgevans: Yes.  Before this change, "#" was tokenized early on as HASH if not the first token, and otherwise the line was skipped.  "#" does not appear to ever trigger line 304, as it is caught by line 257.  In the grammar, it can be parsed as a flag, but that doesn't fit with beancount.core.flags (see also https://github.com/beancount/beancount/issues/516).  

After the change, "#" is never parsed as a flag, per beancount.core.flags.  The line-skipping and "#" in a compound cost is not affected by the change.  Alternatively, "#" could be added to line 275 to be parsed as a flag, without affecting either of hose.

cgevans:
@blais, while this could be fixed more generally with a rewrite, it seems from the mailing list that v3 is in planning stages, involves considerable changes, and has a relatively long timescale.  This change is relatively small, and while it currently also addresses https://github.com/beancount/beancount/issues/516, could be changed to not address this, and would then involve minimal changes: in that case, the only change would be that a single capital letter would only be interpreted as a flag if it came after a date or indent, while flag and first-token behaviour would not change.  Would this smaller change be better?

PR #519, opened 2020-07-29
lexer: Fix subtle bug

dnicolodi:
This typo would cause yy_skip_line() to loop forever if it does not encounter newline before the end of the input.

This was merged into master, but not into v2.


blais:
> This needs to go into v2 too. Should I open a new PR for this?

After untangling v2 & master, it's in, just checked.

PR #520, opened 2020-07-29
parser: Test indentation handling

dnicolodi:
This test is currently failing. Namely, this syntax
```
2020-07-28 open Assets:Foo

    2020-07-28 open Assets:Bar
```
and variations where an entry is indented, is allowed.

This is PR is only to demonstrate the issue. I have a fix for it ready to be pushed, if it is decided that it needs to be fixed.

### DIFF
--- a/beancount/parser/grammar_test.py
+++ b/beancount/parser/grammar_test.py
@@ -227,6 +227,28 @@ class TestParserEntryTypes(unittest.TestCase):
                           (D('23'), Decimal)],
                          txns[0].values)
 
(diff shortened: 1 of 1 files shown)


blais:
Yes, there are a few quirks like that.

blais:
Can you add your fix to the PR? Is it a big change?

dnicolodi:
The fix is intertwined with a set of cleanups I came up with during the port to RE/flex. I am planning to submit all that, but I need to split the patch series up a bit.

blais:
Will merge test and skip manually.

Issue #523, opened 2020-07-31
python-magic needs libmagic on Windows


dnicolodi:
I am trying to run beancount tests on Windows on AppVeyor. They fail because python-magic has some problem loading libmagic:
```
platform win32 -- Python 3.6.8, pytest-5.4.2, py-1.9.0, pluggy-0.13.1
rootdir: C:\projects\beancount
collected 1617 items / 10 errors / 1607 selected
=================================== ERRORS ====================================
_______________ ERROR collecting beancount/ingest/cache_test.py _______________
beancount\ingest\cache_test.py:9: in <module>
    from beancount.ingest import cache
beancount\ingest\cache.py:14: in <module>
    from beancount.utils import file_type
beancount\utils\file_type.py:22: in <module>
    import magic
C:\Python36\lib\site-packages\magic.py:179: in <module>
    libmagic = ctypes.CDLL(dll)
C:\Python36\lib\ctypes\__init__.py:348: in __init__
    self._handle = _dlopen(self._name, mode)
E   OSError: [WinError 193] %1 is not a valid Win32 application
```

I tries getting libmagic from here https://github.com/nscaife/file-windows, but I get the same error...

@yagebu How do you run fava tests on AppVeyor?

yagebu:
> How do you run fava tests on AppVeyor?

https://github.com/beancount/fava/blob/c3e1b67befd10f35a9279350bfaa5aeaab0314dd/appveyor.yml - due to a bug in python-magic, I've fixed the version of python-magic in `tox.ini`.

dnicolodi:
It turned out that if I use `PYTHON: "C:\\Python38"` in appveyor.yml I get the above error, but not if I use `PYTHON: "C:\\Python38-x64"`... However, now I get a lot of test failures. I am investigating.

yagebu:
I'd still consider this a bug that we should work around. A python lib with missing dependencies shouldn't crash Beancount if it's non-essential like python-magic. I'll open a PR with a work-around.

dnicolodi:
I thought the error was caused by something wrong I did. Because of the different behavior in Python38 vs Python38-x64 I thought it had to do with trying to `dlopen()` (or whatever is the equivalent on Windows) a DLL compiled for a different architecture than the main process. However, you are suggesting this is not the case.

PR #524, opened 2020-07-31
add github action to test on windows

yagebu:


yagebu:
Ref #523 

The Github actions aren't triggered here, but are triggered in the PR on my repo: https://github.com/yagebu/beancount/pull/1

beancount/scripts/deps.py
```diff
@@ -125,7 +125,7 @@ def check_python_magic():
             # 'filemagic' is installed; install python-magic.
             raise ImportError
         return ('python-magic', 'OK', True)
-    except ImportError:
+    except (ImportError, OSError):
         return ('python-magic', None, False)
```
dnicolodi: Why does an import result in an OSError? Can you please submit this patch as a standalone PR?

dnicolodi:
If I understand correctly, you dropped Python 3.5 because it is not supported by Github actions. Is this the case? Can you please squash the two remaining commits?

yagebu:
This should be considered WIP, sorry

dnicolodi:
One nice thing of AppVeyor is that it can be used to produce wheels and these can be easily downloaded and uploaded on PyPI at least it seems, I haven't tried. This could be automated and done at each release. Can Github actions be used easily to achieve the same?

beancount/scripts/deps.py
yagebu: Exactly the error in #523

dnicolodi: See my comments in #523

yagebu:
> Can Github actions be used easily to achieve the same?

I believe all the CI solutions are mostly equivalent in their features, so that should also be possible. With Fava, I've also been a mostly happy user of AppVeyor.

Somehow, the builds aren't triggered any more (on my fork), so I'll close this PR. Maybe I'll have another go at this at some point.

dnicolodi:
I am trying to get the tests going on AppVeyoir but so far they get stock somewhere in setup_test.py and I get a lot of tests failures.

PR #525, opened 2020-07-31
parser: Work around MSVC bug

### LINKED ISSUE #522: Latest release does not compile on Windows (on AppVeyor at least).
The CI build of Fava is currently broken due to the latest release of Beancount apparently not compiling on Windows (https://ci.appveyor.com/project/yagebu/fava/builds/34396347):

    beancount/parser/parser.c(336): error C2198: 'PyUnicode_FromString': too few arguments for call

Installing from git (branch v2) seems to fix it (https://ci.appveyor.com/project/yagebu/fava/builds/34397338). The line that throws the error is gated by an `#ifdef VC_CHANGESET` and maybe the passing of that compile option behaves differently on Windows.

dnicolodi:
Fixes #522.

### DIFF
--- a/beancount/parser/macros.h
+++ b/beancount/parser/macros.h
@@ -1,6 +1,12 @@
 #ifndef BEANCOUNT_PARSER_MACROS_H
 #define BEANCOUNT_PARSER_MACROS_H
 
+#define STRINGIFY(x) XSTRINGIFY(x)
+/* Prepend the empty string to work around MSVC bug for which an empty
+ * define results in no string rather than in the empty string. */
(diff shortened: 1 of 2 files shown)


dnicolodi:
I admit I am very tempted to simply get rid of `__vc_changeset__` and `__vc_timestamp__` as they are already gone from the Bazel build. @blais, what do you think? I don't think these constants are that useful.

blais:
IIRC this is similar to how it used to be.
Anyway, I don't care about v2, I want to minimize changes there. 
Putting my attention on v3, which won't need those.

PR #529, opened 2020-08-01
catch OSError on import of python-magic

### LINKED ISSUE #523: python-magic needs libmagic on Windows

yagebu:
Close #523 

Some time ago, python-magic started throwing `OSError`s when trying to import the libmagic or something dynamic library in the Fava CI builds on Windows (there's also an open issue on python-magic tracker: https://github.com/ahupp/python-magic/issues/216). I worked around this by fixing the version of python-magic, but catching this error seems a cleaner work-around.

### DIFF
--- a/beancount/scripts/deps.py
+++ b/beancount/scripts/deps.py
@@ -125,7 +125,7 @@ def check_python_magic():
             # 'filemagic' is installed; install python-magic.
             raise ImportError
         return ('python-magic', 'OK', True)
-    except ImportError:
+    except (ImportError, OSError):
         return ('python-magic', None, False)
 
 
--- a/beancount/utils/file_type.py
+++ b/beancount/utils/file_type.py
@@ -26,7 +26,7 @@ try:
         warnings.warn("You have installed 'filemagic' instead of 'python-magic'; "
                       "disabling.")
         magic = None # pylint: disable=invalid-name
-except ImportError:
+except (ImportError, OSError):
     magic = None
 
 


dnicolodi:
I think this is not the best solution: python-magic is pulled in by `setup.py` thus it should be present and an user would expect it to work, however this hides the failure in initializing the library and the user is left wondering why python-magic is not used. Is it true that some DLL needs to be installed to make python-magic to work? In this case, I would prefer to modify `setup.py` to install python-magic only on non-Windows systems and let the interested users go through the trouble of installing the DLL and python-magic if they really want it. Installing a non-functional python-magic is not very useful. What do you think?

yagebu:
Well, we're already catching the `ImportError`, so `python-magic` is considered somewhat optional by Beancount. 

> Is it true that some DLL needs to be installed to make python-magic to work?

Yes, the README points to binaries provided by some random people that one can download. So this is not really a library that I'd want to encourage users to install...

> In this case, I would prefer to modify setup.py to install python-magic only on non-Windows systems and let the interested users go through the trouble of installing the DLL and python-magic if they really want it. Installing a non-functional python-magic is not very useful. What do you think?

Also sounds like a fine solution :)

PR #533, opened 2020-08-06
Amount constructor to support string for number arg

### LINKED ISSUE #532: Amount does not take string as specified in docstring
Amount constructor states number can be string or Decimal instance, but returns error on providing string.

See: beancount.core.amount.Amount.__new__ (cls, number, currency)

**Currently**:
```
def __new__(cls, number, currency):
    """Constructor from a number and currency.

    Args:
      number: A string or Decimal instance. Will get converted automatically.
      currency: A string, the currency symbol to use.
    """
    assert isinstance(number, Amount.valid_types_number), repr(number)
    assert isinstance(currency, Amount.valid_types_currency), repr(currency)
    return _Amount.__new__(cls, number, currency)
```

**Proposed simple change:**
```
def __new__(cls, number, currency):
    """Constructor from a number and currency.

    Args:
      number: A string or Decimal instance. Will get converted automatically.
      currency: A string, the currency symbol to use.
    """
    if isinstance(number, str):
        number = Decimal(number)
    assert isinstance(number, Amount.valid_types_number), repr(number)
    assert isinstance(currency, Amount.valid_types_currency), repr(currency)
    return _Amount.__new__(cls, number, currency)
```

Filing as issue due to not knowing if auto-conversion of strings is desired or accidentally left in docstring. And I'd recommend accepting floats and ints, if accepting strings, since underlying Decimal constructor takes all easily.

antcer1213:
As specified in docstring located here: beancount.core.amount.Amount.__new__

```
def __new__(cls, number, currency):
    """Constructor from a number and currency.

    Args:
      number: A string or Decimal instance. Will get converted automatically.
      currency: A string, the currency symbol to use.
    """
```
closes #532

### DIFF
--- a/beancount/core/amount.py
+++ b/beancount/core/amount.py
@@ -43,7 +43,7 @@ class Amount(_Amount):
         """Constructor from a number and currency.
 
         Args:
-          number: A string or Decimal instance. Will get converted automatically.
+          number: A Decimal instance.
           currency: A string, the currency symbol to use.
         """
         assert isinstance(number, Amount.valid_types_number), repr(number)


antcer1213:
#532 #

dnicolodi:
I think this goes in the wrong directions and it would be much better to simply fix the docstring. If we allow strings to represent numbers, then there will be requests for handling commas as thousands separator, rapidly followed by requests to handle number representations in arbitrary locales. I don't think having to parse numbers before calling the Amount constructor causes any inconvenience and, in this case, there is no advantage to a polymorphic API. What use case are you trying to simplify?

dnicolodi:
Also, the patch to the tests introduces a syntax error.

antcer1213:
> Also, the patch to the tests introduces a syntax error.

Where is the syntax error in the two lines of additional code introducted? May result in a ValueError currently if the patch to Amount is not implemented first.

antcer1213:
> I think this goes in the wrong directions and it would be much better to simply fix the docstring. If we allow strings to represent numbers, then there will be requests for handling commas as thousands separator, rapidly followed by requests to handle number representations in arbitrary locales. I don't think having to parse numbers before calling the Amount constructor causes any inconvenience and, in this case, there is no advantage to a polymorphic API. What use case are you trying to simplify?

Since Amount already uses and expects a Decimal from the decimal library, it already supports commas in string-represented numbers. See patch.

No use-case, outside of docstring currently stating a string is an accepted 'number'. I do find it a bit odd that it requires explicitly converting a string to Decimal prior to creating an Amount, since beancount is a plaintext-based ledger accounting system and strings should be considered first-class members, accepted as much as possible.

antcer1213:
> Also, the patch to the tests introduces a syntax error.

Corrected the syntax error. Apologies. Did not run an execution check.

antcer1213:
Open to just editing the docstring to state it only accepts Decimals, for proper documentation. However, do not see how accepting strings for Amount (as originally intended) would be going against any envisioned direction.

dnicolodi:
> Since Amount already uses and expects a Decimal from the decimal library, it already supports commas in string-represented numbers. See patch.

Are you sure?
```
>>> from decimal import Decimal
>>> Decimal('1,234.00')
  File "<stdin>", line 1, in <module>
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
```
Running the tests before submitting a PR is good practice. See also results of the CI tests.

dnicolodi:
> I do find it a bit odd that it requires explicitly converting a string to Decimal prior to creating an Amount, since beancount is a plaintext-based ledger accounting system and strings should be considered first-class members, accepted anywhere and everywhere.

beancount is a plaintext-based ledger accounting system, in the sense that the input files are plain text.

antcer1213:
> > Since Amount already uses and expects a Decimal from the decimal library, it already supports commas in string-represented numbers. See patch.
> 
> Are you sure?
> 
> ```
> >>> from decimal import Decimal
> >>> Decimal('1,234.00')
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
> ```
> 
> Running the tests before submitting a PR is good practice. See also results of the CI tests.

Yikes, you're right. Will make sure to do so. I called the Decimal library directly (apparently, without support of commas), versus function D from beancount.core.number that already exists and has the ability to parse strings with commas.

dnicolodi:
Please, just fix the docstring. In any case, there is no need for three patches for a three lines change: you should squash the patches and force push your branch.

dnicolodi:
I think we may have a communication problem. Which part of "Please, just fix the docstring" was not clear? Thank you.

antcer1213:
> I think we may have a communication problem. Which part of "Please, just fix the docstring" was not clear? Thank you.

Sorry about that, I could have sworn I did last night. Apologies on delay; was originally using svn. Had to clone with git to squash and force-push after undoing.

blais:
Thank you!

PR #534, opened 2020-08-08
Improve the CSV exporter's chances of surviving an emoji-filled input file

jpluscplusm:
The British bank "Monzo" embeds emoji in its CSV exports.

Files containing these emoji (or at the very least, some of them) cause the underlying `chardet` library to explode.

The only user-level control that exists is the Importer's `encoding` param, but this setting wasn't being respected during attempts to sniff the CSV to see if it contains a header row.

This PR adds a harmless-if-absent parameter to the cache's head() methods, by which the CSV importer can communicate the encoding the user has set explicitly.

[Apologies for the lack of additional tests. I'd naively assumed that, as a python project, this would be trivially testable with a `pip install -r req.txt` and a `make test`. The fact that it's not (see below) was a big enough bump in the road that I thought it best to just PR this change without tests. The cache tests pass, but they're the only ones I can run without getting into why I can't build the parser ...

Tests produced oodles of these:

```
ImportError while importing test module '/home/user/code/beancount/beancount/ingest/scripts_utils_test.py'.
Hint: make sure your test modules/packages have valid Python names.                                                                                                                                                         
Traceback:                                                                                                                                                                                                                  
beancount/ingest/scripts_utils_test.py:11: in <module>                                                                                                
    from beancount.ingest import scripts_utils                            
beancount/ingest/scripts_utils.py:19: in <module>                                                                                                                                                                           
    from beancount.parser import version                                                                                                            
beancount/parser/version.py:11: in <module>                                                                                                              
    from beancount.parser import _parser                                                                                                                                                                                   
E   ImportError: cannot import name '_parser' from 'beancount.parser' (/home/user/code/beancount/beancount/parser/__init__.py)
```

I understand they're because the parser isn't written in python, but I lack the immediate expertise to get it building successfully.]

### DIFF
--- a/beancount/ingest/cache.py
+++ b/beancount/ingest/cache.py
@@ -58,9 +58,9 @@ class _FileMemo:
         """Computes the MIME type of the file."""
         return self.convert(mimetype)
 
-    def head(self, num_bytes=8192):
+    def head(self, num_bytes=8192, encoding=None):
         """An alias for reading just the first bytes of a file."""
-        return self.convert(head(num_bytes))
+        return self.convert(head(num_bytes, encoding=encoding))
 
     def contents(self):
         """An alias for reading the entire contents of the file."""
@@ -76,7 +76,7 @@ def mimetype(filename):
     return file_type.guess_file_type(filename)
 
 
-def head(num_bytes=8192):
+def head(num_bytes=8192, encoding=None):
     """A converter that just reads the first bytes of a file.
 
     Args:
@@ -87,9 +87,8 @@ def head(num_bytes=8192):
     def head_reader(filename):
         with open(filename, 'rb') as file:
             rawdata = file.read(num_bytes)
-            detected = chardet.detect(rawdata)
-            encoding = detected['encoding']
-            return rawdata.decode(encoding)
+            file_encoding = encoding or chardet.detect(rawdata)['encoding']
+            return rawdata.decode(file_encoding)
     return head_reader
 
 
--- a/beancount/ingest/cache_test.py
+++ b/beancount/ingest/cache_test.py
@@ -47,6 +47,14 @@ class TestFileMemo(unittest.TestCase):
             mimetype = wrap.convert(cache.mimetype)
             self.assertRegex(mimetype, r'text/(x-(python|c\+\+)|plain)')
 
+    def test_cache_head_obeys_explict_utf8_encoding_avoids_chardet_exception(self):
+        emoji_header = 'asciiHeader1,🍏Header1,asciiHeader2'.encode('utf-8')
+        with mock.patch('builtins.open',
+                mock.mock_open(read_data=emoji_header)):
+            try:
+                function_return = cache._FileMemo('anyFile').head(encoding='utf-8')
+            except UnicodeDecodeError:
+                self.fail("Failed to decode emoji")
 
 if __name__ == '__main__':
     unittest.main()
--- a/beancount/ingest/importers/csv.py
+++ b/beancount/ingest/importers/csv.py
@@ -184,7 +184,11 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
     def file_date(self, file):
         "Get the maximum date from the file."
         iconfig, has_header = normalize_config(
-            self.config, file.head(), self.csv_dialect, self.skip_lines)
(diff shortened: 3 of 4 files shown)


dnicolodi:
I don't understand the complaint about the tests. beancount provides one python extension that needs to be compiled in order for it to run, this is not in any way specific to beancount. If you went far enough to discover `make test` (which is definitely not standard for a Python project) you could as easily found the `make build` target. This also rebuilds the lexer and grammar from their definitions, thus you need Flex and Bison installed. To just build the extension, using the generated source file in the git repository, simply  run `python setup.py build_ext -i`. I am sure this is documented (not for beancount specifically) in dozen of places.

How does the chardet library explode? If your files are correctly encoded (in UTF-8, I guess) and chardet cannot handle them, this is a bug in chardet which is worth reporting and fixing. Do you have a traceback or something?

However, I agree that, if the user has specified an encoding, it should be used consistently, without reverting to guessing. The patches look reasonable to me. Can you please, with the instructions above, get the tests running and add a couple of tests specific to your changes?

jpluscplusm:
> How does the chardet library explode? If your files are correctly encoded (in UTF-8, I guess) and chardet cannot handle them, this is a bug in chardet which is worth reporting and fixing. Do you have a traceback or something?

Here's what I see, and all the files required to reproduce (with .txt suffixes so GitHub will accept them as uploads).

```shell
(venv) $ pip3 install beancount==2.3.1 >/dev/null
(venv) $ python3 --version
Python 3.7.3
(venv) $ bean-extract --version
Beancount 2.3.1
(venv) $ bean-extract personal.import.py monzo.csv
ERROR:root:Importer importers.monzo_current_account_csv.Importer: "Assets:Monzo".extract() raised an unexpected error: 'charmap' codec can't decode byte 0x8d in position 220: character maps to <undefined>
  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/cache.py", line 50, in convert
    result = self._cache[converter_func]
KeyError: <function head.<locals>.head_reader at 0x7eae9e843ea0>

During handling of the above exception, another exception occurred:

  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/extract.py", line 188, in extract
    allow_none_for_tags_and_links=allow_none_for_tags_and_links)
  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/extract.py", line 67, in extract_from_file
    new_entries = importer.extract(file, **kwargs)
  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/importers/csv.py", line 213, in extract
    self.config, file.head(), self.csv_dialect, self.skip_lines)
  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/cache.py", line 63, in head
    return self.convert(head(num_bytes))
  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/cache.py", line 54, in convert
    result = self._cache[converter_func] = converter_func(self.name)
  File "/home/jonathan/tmp/venv/lib/python3.7/site-packages/beancount/ingest/cache.py", line 92, in head_reader
    return rawdata.decode(encoding)
  File "/usr/lib/python3.7/encodings/cp1254.py", line 15, in decode
    return codecs.charmap_decode(input,errors,decoding_table)
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 220: character maps to <undefined>
```

[monzo.csv.txt](https://github.com/beancount/beancount/files/5046202/monzo.csv.txt)
[personal.import.py.txt](https://github.com/beancount/beancount/files/5046203/personal.import.py.txt)
[monzo_current_account_csv.py.txt](https://github.com/beancount/beancount/files/5046204/monzo_current_account_csv.py.txt)

jpluscplusm:
`make build`, which I'd of course found and run, turns out to be unnecessary. I'd stopped after it failed early, as I'm on debian stable which doesn't have a recent enough bison to run `make build`. However, the `build_ext` command you mentioned completes, sufficient to allow `make test` to then pass.

I'll add some tests.

jpluscplusm:
Tests added.

jpluscplusm:
Awesome! Thanks folks :-)

PR #535, opened 2020-08-10
Optimization of the construction of Decimal objects and assorted cleanups

dnicolodi:
This follows the conversion of the building of lexer token semantic values to compiled code. Some commits are trivial, thus I believe this is a reasonable number of patched to be reviewed at once.

### DIFF
--- a/beancount/parser/decimal.h
+++ b/beancount/parser/decimal.h
@@ -1,0 +1,28 @@
+#ifndef BEANCOUNT_DECIMAL_H
+#define BEANCOUNT_DECIMAL_H
+
+/**
+ * Python does not expose a C-API to create Decimal objects. This code
+ * implements a system similar to what Python provides for @date and
+ * @datetime objects from the datetime module. The exposed API is
+ * similar to the one implemented (but not exposed) by the _decimal
+ * extension module. Before using any of these functions, the macro
+ * PyDateTime_IMPORT must be invoked.
+ */
+
+extern PyObject* decimal_type;
+
+/**
+ * Initialization.
+ */
+#define PyDecimal_IMPORT do {                                           \
+        PyObject* decimal = PyImport_ImportModule("decimal");           \
+        decimal_type = PyObject_GetAttrString(decimal, "Decimal");      \
+    } while (0)
+
+/**
+ * Return a Decimal object from a C-string representation.
+ */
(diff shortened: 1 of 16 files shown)


beancount/parser/decimal.h
```diff
@@ -0,0 +1,25 @@
+#ifndef BEANCOUNT_DECIMAL_H
+#define BEANCOUNT_DECIMAL_H
+
+/*
+ * Python does not expose a C-API to create Decimal. This code
+ * implements a system similar to what Python provides for @date and
+ * @datetime objects from the datetime module. The exposed API is
+ * similart to the one implemented (but not exposed) by the decimal
+ * module. Before using any of these functions, the the macro
```
blais: the the

beancount/parser/decimal.h
```diff
@@ -0,0 +1,25 @@
+#ifndef BEANCOUNT_DECIMAL_H
+#define BEANCOUNT_DECIMAL_H
+
+/*
+ * Python does not expose a C-API to create Decimal. This code
+ * implements a system similar to what Python provides for @date and
+ * @datetime objects from the datetime module. The exposed API is
+ * similart to the one implemented (but not exposed) by the decimal
```
blais: similar

blais (approved):

dnicolodi:
Fixed

beancount/parser/decimal.h
```diff
@@ -0,0 +1,25 @@
...
+/*
+ * Python does not expose a C-API to create Decimal. This code
+ * implements a system similar to what Python provides for @date and
+ * @datetime objects from the datetime module. The exposed API is
+ * similar to the one implemented (but not exposed) by the decimal
+ * module. Before using any of these functions, the macro
+ * PyDateTime_IMPORT must be invoked.
+ */
+
+static PyObject* decimal_type;
```
blais: Hmmm, don't declare variables in a header file; this basically requires that header file to be included in a single compilation unit. You could make it a macro for the declaration, but it would be better to just put that in the .c file, or have a decimal.c file just for its definition.

dnicolodi: I know this is not optimal. I was trying to keep things simple and at the same time self contained. I'll add a decimal.c for this variable.

dnicolodi:
Done. What happened to the Travis CI jobs? I edited the Bazel build definition and I was hoping the CI would tell me if I got it right...

PR #538, opened 2020-08-14
Fix Windows build.

skobes-chromium:
MSVC has quirky handling of `__VA_ARGS__`, and does not support statement expressions.

### DIFF
--- a/beancount/parser/tokens.h
+++ b/beancount/parser/tokens.h
@@ -12,11 +12,14 @@
  * are handled and reported as lexing errors.
  */
 #define token(name, ...)                                        \
-    yylval->pyobj = build_##name(__VA_ARGS__);                  \
+    yylval->pyobj = invoke_builder(name, (__VA_ARGS__));        \
     if (yylval->pyobj == NULL) {                                \
         build_lexer_error_from_exception(yylloc, builder);      \
         return YYerror;                                         \
     }
+
+/* MSVC requires this indirection to expand __VA_ARGS__ correctly. */
+#define invoke_builder(name, arg_list) build_##name arg_list
 
 #define build_STR(_ptr, _len) PyUnicode_FromStringAndSize(_ptr, _len)
 #define build_KEY build_STR
@@ -24,7 +27,7 @@
 #define build_LINK build_STR
 #define build_CURRENCY build_STR
 #define build_BOOL(_value) PyBool_FromLong(_value)
-#define build_NONE() ({ Py_INCREF(Py_None); Py_None; })
+#define build_NONE() ( Py_INCREF(Py_None), Py_None )
 #define build_NUMBER(_str) pydecimal_from_cstring(_str)
 #define build_DATE(_str) pydate_from_cstring(_str)
 #define build_STRING(_str, _len, _enc) pyunicode_from_cquotedstring(_str, _len, _enc)
--- a/beancount/parser/tokens.h
+++ b/beancount/parser/tokens.h
@@ -18,7 +18,13 @@
         return YYerror;                                         \
     }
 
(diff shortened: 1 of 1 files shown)


skobes-chromium:
FYI - "setup.py install" now completes successfully, and bean-check seems to work.  However, I am not sure how to run the tests.

dnicolodi:
To run the tests:
```
python -m pytest beancount/
```
from the source directory.

However, the tests do not run on Windows. I have a branch that fixes some of the problems (Windows is really not a nice platform) but there are many tests that assume `/` being the filesystem path separator and adjusting all those is tedious work I have very little incentive to complete.

dnicolodi:
> MSVC has quirky handling of `__VA_ARGS__`, and does not support statement expressions.

What a nice compiler!

beancount/parser/tokens.h
```diff
@@ -12,19 +12,22 @@
...
+#define invoke_builder(name, arg_list) build_##name arg_list
+
 #define build_STR(_ptr, _len) PyUnicode_FromStringAndSize(_ptr, _len)
 #define build_KEY build_STR
 #define build_TAG build_STR
 #define build_LINK build_STR
 #define build_CURRENCY build_STR
 #define build_BOOL(_value) PyBool_FromLong(_value)
-#define build_NONE() ({ Py_INCREF(Py_None); Py_None; })
+#define build_NONE() ( Py_INCREF(Py_None), Py_None )
```
dnicolodi, line 36: I don't think this is correct.

skobes-chromium: Why?

trim21: @dnicolodi It's corret, it builds fine after this patch is apply on master.

skobes-chromium:
ping

dnicolodi:
@skobes Can you please add a comment to MSCV documentation about the handling of `__VA_ARGS__` and that possibly explain why the indirection fixes the problem? This code is likely to evolve in the future (actually I have the patches ready for submission already) and I would like to have the reference ready. The need for the indirection does not make any sense to me.

beancount/parser/tokens.h
dnicolodi: It works, indeed. I don't know what I was thinking when I read the code the first time.

skobes-chromium:
Not sure if you meant the inline comment, or documentation in a different place.  Anyway I expanded the inline comment to more fully explain the fix.

dnicolodi:
@skobes, Now that I understand what the problem with `__VA_ARGS__` is, I like more the solution i propose in #543 as it is easier to understand and it does not rely on the funny looking syntax of this solution, namely the parenthesis around `(__VA_ARGS__)` in the `invoke_builder` second argument. Can you please test that other approach and let me know if it works as intended?

dnicolodi:
@skobes, it would also be great if you could verify that the macro magic in https://github.com/beancount/beancount/blob/master/beancount/parser/macros.h works as intended when compiled with MSVC. The comment at the end of the file should allow to easily construct a test.

blais:
Fixed via #543.

skobes-chromium:
@dnicolodi, thanks for submitting your alternate fix.  I have verified that #543 makes MSVC happy.

Regarding `macros.h`, I think that MSVC does not currently handle this correctly. The example `_CC_FUNC(putc, "a", "b", "c")` expands to `putc("a", "b", "c")`, which seems different from the intended expansion, based on the comment.

dnicolodi:
Thank you for checking. I am surprised that the code using `macros.h` compiles at all. I'll investigate and see if there is something that need fixing.

dnicolodi:
Indeed. Fix pending in #545.

PR #544, opened 2020-08-31
add a windows deps to enable python-magic

trim21:
This pr enable `python-magic` as wheel is released on pypi, and user don't need to compile it from source.
And add `python-magic-bin` package as `python-magic` suggested 

> Windows
>You'll need DLLs for libmagic. @julian-r maintains a pypi package with the DLLs, you can fetch it with:
>
>$ pip install python-magic-bin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,62 +76,6 @@ for get_changeset in get_git_changeset, get_hg_changeset:
 else:
     vc_changeset, vc_timestamp = '', 0
 
-
-install_requires = [
-    # Testing support now uses the pytest module.
-    'pytest',
-
-    # This is required to parse dates from command-line options in a
-    # loose, accepting format. Note that we use dateutil for timezone
-    # database definitions as well, although it is inferior to pytz, but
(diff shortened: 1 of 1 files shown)


dnicolodi:
I don't think that requiring binaries from random sources is a great idea. Also, `python-magic-bin` seems to be an (unmaintained?) fork of `python-magic` and not only a distribution of the required DLLs. Thus the changes in `setup.py` do not look correct to me as two version of the `magic` module are installed on Windows. Fianally, I tried `python-magic-bin` and it worked only for some combinations 32bit vs 64bit architecture / 32bit vs 64bit python. We discussed other possible solutions when the `python-magic` dependency was dropped on Windows and we didn't find anything satisfying. `python-magic` is anyhow an optional dependency: if you need the functionality enabled by it, just install the package and the required DDLs and support files in the way you prefer.

trim21:
@dnicolodi you are right...

(In fact there is another problem. if `python-magic` is somehow installed, running bean-extract will get undebuggable `segmentation fault`.

PR #545, opened 2020-09-02
parser: More MSVC compiler fixes

dnicolodi:
More work-arounds to the broken MSVC handling of __VA_ARGS__. The
source was compiling but the generated code was most probably not
doing the right thing.


blais:
I have a feeling we won't need this macro tricks in the future.

PR #549, opened 2020-09-05
Lexer cleanup

dnicolodi:
Some trivial cleanups in the parser code. This is a first step toward having the lexer generated by RE/flex.


blais:
Loving it, but a bit scared it might break existing skipped lines. Will have to test before merging, but otherwise SGTM.

blais:
Seems to work fine. Merging.

dnicolodi:
> Loving it, but a bit scared it might break existing skipped lines. Will have to test before merging, but otherwise SGTM.

I think the unit tests cover this aspect well. I have patches to simplify the handling of skipped lines and flag characters, but they break backward compatibility in some cases. I would like to merge them, but maybe it is material for v3 once I upstream my port of the lexer to RE/flex. I'll soon open another PR with another batch of patches toward using RE/flex.

PR #558, opened 2020-09-19
parser: Move functions defined in tokens.h to own compilation unit

dnicolodi:
This requires hacking around the limitations of the Python C-API to
the datetime module to avoid having to initialize it in every
compilation unit using it.


blais:
Hmmm, I had done just that but somehow my tests were still dumping core.
I'll diff.
Thank yo,

PR #561, opened 2020-09-19
parser: Fix make ctest

dnicolodi:
Adjust compilation options and CI job definition.

### DIFF
--- a/.github/workflows/linux-ci.yaml
+++ b/.github/workflows/linux-ci.yaml
@@ -18,6 +18,8 @@ jobs:
       - run: pip install -r requirements_dev.txt
       - run: python setup.py build_ext -i
       - run: make ctest
+        env:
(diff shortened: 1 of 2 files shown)


blais:
BTW it may be that the unused variable warning is probably because assert() must compile to a noop in -c opt.
All of this doesn't matter much to me because in v3 it's all googletest.

dnicolodi:
The unused variable warnings come from the redirection of the `PyDateTimeAPI` symbol via the macro in `datetime.h`. `PyDateTimeAPI` is defined in the Python's `datetime.h` and used only via macros, thus it becomes unused. `PyDateTimeAPI` is declared static, thus I don't know a trick to make the compile do not complain.

blais:
I found another way (avoid redefining the macro, move the function in
parser.c).

PR #562, opened 2020-09-24
refactor: use py3.6 typing syntax to define directive

trim21:
consider now beancount require python>=3.6, it is possible to defina directives in beancount.core.data with new python typing syntax, and it's possible to define default value for all optional attribute(I didn't do it in this pr, not show if you want it).

This will make editor be able to suggest better completions for `__init__` args, and all `NamedTuple` method.

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -9,7 +9,7 @@ import enum
 import sys
 
 from decimal import Decimal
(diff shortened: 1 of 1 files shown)


blais:
I'm planning to do something like this in v3, but not for v2. I prefer not to mess with v2 too too much.

PR #569, opened 2020-10-25
core.data: create data classes directly with NamedTuple

yagebu:
This should fix the issue of pylint not recognising all these classes as
callable, apparently the function creating and returning a NamedTuple
confused it too much.

This also fixes the typing of these central data structures, which was
quite incorrect. `new_directive` claimed to return an instance (of
NamedTuple) without any further typings but indeed returned a class. At
least with mypy, the typings further down the file all produced errors.

For these central data structures, I think the more explicit (and in my
eyes simpler) definition trumps the succinctness of not having to repeat
the shared first two fields meta and date.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -119,7 +119,15 @@ disable=locally-disabled,
   assignment-from-no-return,
   c-extension-no-member,
   cyclic-import,
-  isinstance-second-argument-not-valid-type
+  isinstance-second-argument-not-valid-type,
+  bad-continuation,
+  inherit-non-class
+
+# Notes:
+#   bad-continuation: Is buggy, see https://github.com/PyCQA/pylint/issues/3512
+#   inherit-non-class: On Python 3.9, NamedTuple is a function that can be
+#                      inherited from, so this throws false positives, see
+#                      https://github.com/PyCQA/pylint/issues/3876
 
 
 [VARIABLES]
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -9,7 +9,7 @@ import enum
 import sys
 
 from decimal import Decimal
-from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, Union
+from typing import Any, Dict, List, NamedTuple, Optional, Set, Union
 
 from beancount.core.amount import Amount
 from beancount.core.number import D
@@ -53,22 +53,6 @@ class Booking(enum.Enum):
     LIFO = 'LIFO'
 
 
-def new_directive(clsname, fields: List[Tuple]) -> NamedTuple:
-    """Create a directive class. Do not include default fields.
-    This should probably be carried out through inheritance.
-
-    Args:
-      name: A string, the capitalized name of the directive.
-      fields: A string or the list of strings, names for the fields
-        to add to the base tuple.
-    Returns:
-      A type object for the new directive type.
-    """
-    return NamedTuple(
-        clsname,
-        [('meta', Meta), ('date', datetime.date)] + fields)
-
-
 # All possible types of entries. These are the main data structures in use
 # within the program. They are all treated as immutable.
 #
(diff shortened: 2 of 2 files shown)


blais:
Yes, I've been thinking about doing something similar:
https://github.com/beancount/beancount/blob/master/TODO#adding-types

Now, I wonder how the constructors change.
Also, when I merge this, I'll want to weave in the comments.

yagebu:
> Yes, I've been thinking about doing something similar:

:+1:  for Fava, I've created type stubs for most of the parts of Beancount that are used and I find those quite helpful. `mypy` works really well for me, it can infer quite a lot and works quite well for a gradual adoption of types :)

> Now, I wonder how the constructors change.

The metaclass of NamedTuple calls and returns a `collections.namedtuple` underneath (https://github.com/python/cpython/blob/3.8/Lib/typing.py#L1591), so the constructors should be the same.

> Also, when I merge this, I'll want to weave in the comments.

I went ahead and moved the comments to the class docstrings.

blais:
Thanks Jakob, I'm familiar with it, we use it ubiquitously at work (pytype checks are integrated in the build rules, and hopefully soon in Bazel). I'm still pondering whether this should be 3.x material (I'm not going to have much time to handle v2 issues going forward).

.pylintrc
```diff
@@ -120,10 +120,14 @@ disable=locally-disabled,
   c-extension-no-member,
   cyclic-import,
   isinstance-second-argument-not-valid-type,
-  bad-continuation
+  bad-continuation,
+  inherit-non-class
```
dnicolodi, line 124: Is there a way to disable this only for the relevant code? IIRC pylint checkers can be disabled per file.

beancount/core/data.py
```diff
@@ -225,9 +225,9 @@ class Transaction(NamedTuple):
 # Attributes:
 #   txn: The parent Transaction instance.
 #   posting: The Posting instance.
-TxnPosting = NamedTuple('TxnPosting', [
-    ('txn', Transaction),
-    ('posting', Posting)])
+class TxnPosting( NamedTuple ):
```
dnicolodi: Does not matter much, but there are extra spaces within the parentheses.

dnicolodi (commented):
I like this. One more thing it seems it does is to attache the documentation to the types, which is a Good Thing. It could also be the basis for introducing specialized types for more beancount elements that are currently represented simply as strings. @yagebu can you comment on any possible breakage that this may introduce? I don't see any possible adverse consequence and I would be in favor of merging it.

beancount/core/data.py
yagebu: Thanks

.pylintrc
yagebu: Yes, that would also be possible - but that would make it harder to adopt NamedTuple in other parts of Beancount. I think its better to have this globally disabled until this rule is adjusted for Python 3.9.

yagebu:
> can you comment on any possible breakage that this may introduce? 

https://docs.python.org/3/library/typing.html#typing.NamedTuple states that the two variants are equivalent. And the source code confirms that (apart from maybe some minor differences on some of the dunder attributes). So I don't really see any possible breakage due to this.

blais:
I hear you. But I really want to spend zero cycles on v2 at this point. I've been checking in harmless changes in v2 so far. If this breaks stuff, I don't want to spend time handling it. master is okay. I could merge this to master.

.pylintrc
blais: I think some of these are obsolete anyway, bad-continuation isn't supported yet (they advise using Black instead).

yagebu:
> I could merge this to master.

:+1:

blais:
+ 1
I'll move it, don't worry about it, I think it's switchable.

PR #576, opened 2020-11-14
parser: Print balance tolerances

aclindsa:
I noticed that balance tolerances were not printed with `bean-query example.beancount print`

I'm not sure my fix is the cleanest, but it wasn't immediately obvious to me how to better break up the traditional way the amounts are printed (which have the number and currency paired) in order to stick the tolerance between them. Suggestions welcome!

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -267,9 +267,16 @@ class EntryPrinter:
 
     def Balance(self, entry, oss):
         comment = '   ; Diff: {}'.format(entry.diff_amount) if entry.diff_amount else ''
-        oss.write(('{e.date} balance {e.account:47} {amount}'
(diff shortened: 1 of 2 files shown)


aclindsa:
I see that the `bazel-ci / compile` test failed, but the error message is a little opaque to me. Is this a known failure (not because of my PR)?

dnicolodi:
Printing the tolerance is of course desirable, however, your approach seems a bit contrived to me. I see two better strategies: don't use `Amount.to_string()` to format `balance` directives but open code the relevant parts, or extend `Amount.to_string()` to take an optional `tolerance` parameter and do the right thing. Either approaches are fine.

Don't worry about the failed Bazel build, the CI does not like it.

blais:
I'll adjust the code a bit and use the guts of Amount.,

aclindsa:
@blais Thanks for fixing it up - you were done before I even got a chance to respond!

PR #580, opened 2020-11-21
Parser: more cleanup

dnicolodi:
Some minor stuff that annoyed me during the port of the lexer from Flex to RE/flex.

### DIFF
--- a/beancount/parser/lexer.l
+++ b/beancount/parser/lexer.l
@@ -70,9 +70,6 @@ struct _yyextra_t {
     const char* encoding;
 };
 
-#define yy_filename yyget_extra(yyscanner)->filename
-#define yy_encoding yyget_extra(yyscanner)->encoding
-
 /* Build and accumulate an error on the builder object. */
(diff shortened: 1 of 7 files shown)


dnicolodi:
I am pretty sure the pylint error is not introduced by these patches.

beancount/parser/lexer.l
```diff
@@ -240,34 +235,31 @@ NULL		{
 ^[\*\:\#]/.	{ BEGIN(IGNORE); }
 ^{FLAGS}/.	{ BEGIN(IGNORE); }
 
- /* Default rule. {bf253a29a820} */
```
blais, line 243: Please leave the reference in (grep to find the other side).

blais:
I'll add it back in myself.
In the future, please LMK AOT before updating the versions of the generators.
Probably best to do those on their own.

dnicolodi:
> In the future, please LMK AOT before updating the versions of the generators.

Sure. It is not really needed in this case as the new version of Bison does not bring anything relevant to the use in beancount. I just upgraded my system and I was too lazy to put together an environment with the old version.

PR #581, opened 2020-11-21
Parser: port lexer from Flex to RE/flex

dnicolodi:
@blais this is the port of the lexer to RE/flex. I updated the classical build (ie not the Bazel one) to compile the code and I haven't tried to integrate it into the Bazel build. Therefore the CI will fail miserably.

I tested the code against v2 and it works just the same as the Flex one. Let me know if you have any question. You need RE/flex version 2.1.5 or later.

I have a couple more patches on top that rework the input handling. I was hoping to get support for encodings other than UTF-8, but that is implemented in RE/flex only for `FILE*` type input sources and not for `std::istream` ones (see Genivia/RE-flex#94).

This contains also the commits in #580: #580 should be merged into v2 and this into master.


blais:
Hmmm, I started to make changes to configure the auto-generation of the flex/bison files, this will probably class somewhat with this. I'll have to merge by hand (or you'll have to update). I'll keep  you posted.

dnicolodi:
Let me know when you commit that work and I'll rebase this in top.

blais:
I need to figure out how to remove the dependency of the scanner and grammar on parser.h before doing anything.

dnicolodi:
I don't know why this is a requirement. However, `parser.h` has:

- `typedef void* yyscan_t;` which is a workaround to the fact that Flex is stupid. It should not be necessary anympre with RE/flex.

- `#define YY_DECL ...` also goes away with RE/flex.

- `extern PyObject* missing_obj;` this can be moved everywhere you like. We can also find a way to do not need it.

- `PyObject* pydate_from_cstring(const char* string);` this is here because of the problems with the initialization of the `datetime` API from different compilation units. You didn't like my solution to do not have to put it here :-)

blais:
I've got it fixed. Spaghetti dependencies for more but slightly better.
I'll submit the changes to v2 tomorrow, merge to master, and so with that I
should be able to setup the build rules to avoid pregenerated code.
Then I'll try to follow the template from github.com/blais/oblique.

blais:
Done

dnicolodi:
The changes do not seem to impact the RE/flex port too much. Let me try to rebase.

blais:
This PR raises some interesting questions:

- Are we going to maintain the Python-based parser in v3 (with C++, as you do here)?

I was thinking that we would dispatch between 
(a) the C parser using flex (not requiring C++) in v3, calling the Python builder and 
(b) the newer C++ based RE/flex parser, calling a purely C++/protos builder, in Bazel build,
finding a way to share the same rules. Just for a while.

This adds a different beast: a C++ RE/flex parser using a locally installed RE/flex.
Introduces deps on C++, RE/flex library in the local installation.

I thought the idea is that developers could work in v3 (to test the latest changes in practice) while still using the current C-level parser, until the new parser is feature complete ready. The question is whether regular users will do this, and if so, is adding new dependencies on their local box reasonable, and whether we want to support them.

dnicolodi:
I was working under the assumption that v3 will have a C++ parser based on RE/flex and Bison only and a build system based on Bazel. The problem with maintaining a Flex C lexer and a RE/flex based one is that RE/flex can only emit C++ code and there is not easy way to have the same lexer definition work with C and C++ (unless we take a somewhat macro heavy approach).

One thing I haven't tried is to use the same definition for Flex in C++ mode and RE/flex. In theory it should work without (too many) hacks. At that point we could switch also the grammar to C++ and have Python object or protobufs emitted by two different sub-classes of the same grammar. However, this also mean that in v3 we cannot use any RE/flex-only feature, at least till we support Flex too.

At least on macOS and Linux, installing RE/flex is trivial, thus I would not worry too much about adding a dependency on it.

dnicolodi:
Rebased. I hope I didn't break anything. Tests pass locally.

blais:
> I was working under the assumption that v3 will have a C++ parser based on RE/flex and Bison only and a build system based on Bazel.

Yes.

> The problem with maintaining a Flex C lexer and a RE/flex based one is that RE/flex can only emit C++ code and there is not easy way to have the same lexer definition work with C and C++ (unless we take a somewhat macro heavy approach).

Indeed, RE/flex + C not possible.


> One thing I haven't tried is to use the same definition for Flex in C++ mode and RE/flex.

Oh... let's not waste time on that. We'll never use Flex C++.

> In theory it should work without (too many) hacks. At that point we could switch also the grammar to C++ and have Python object or protobufs emitted by two different sub-classes of the same grammar. However, this also mean that in v3 we cannot use any RE/flex-only feature, at least till we support Flex too.

I think this is what I'm suggesting in our email thread.
Basically:
- I'll copy my working C++ build for RE/flex and Bison C++ from oblique.
- I'll insert the Beancount lexer and grammar rules, pulling in the Unicode categories from your port.
- I'll change the grammar rules to issue to a C++ driver object.
- Then I'll derived a version that calls our to the Python grammar.
- Finally, I'll make the parser.c module definition call onto this parser.
This makes master back into a working state using the Python objects.

Step 2: 
- Derive another driver class that issues protobufs/C++ darta structures instead of calling to Python builder.

I'd also like to reformat the lexer and grammar rules a bit to remove the variable indent. Contemplating doing that in v2 so it's as close as possible to v3 (even if the guts is a bit different).




> 
> At least on macOS and Linux, installing RE/flex is trivial, thus I would not worry too much about adding a dependency on it.

blais:
I was working under the assumption that v3 will have a C++ parser based on
RE/flex and Bison only and a build system based on Bazel.

Yes.

The problem with maintaining a Flex C lexer and a RE/flex based one is that
RE/flex can only emit C++ code and there is not easy way to have the same
lexer definition work with C and C++ (unless we take a somewhat macro heavy
approach).

Indeed, RE/flex + C not possible.

One thing I haven't tried is to use the same definition for Flex in C++
mode and RE/flex.

Oh... let's not waste time on that. We'll never use Flex C++.

In theory it should work without (too many) hacks. At that point we could
switch also the grammar to C++ and have Python object or protobufs emitted
by two different sub-classes of the same grammar. However, this also mean
that in v3 we cannot use any RE/flex-only feature, at least till we support
Flex too.

I think this is what I'm suggesting in our email thread.
Basically:

   - I'll copy my working C++ build for RE/flex and Bison C++ from oblique.
   - I'll insert the Beancount lexer and grammar rules, pulling in the
   Unicode categories from your port.
   - I'll change the grammar rules to issue to a C++ driver object.
   - Then I'll derived a version that calls our to the Python grammar.
   - Finally, I'll make the parser.c module definition call onto this
   parser.
   This makes master back into a working state using the Python objects.

Step 2:

   - Derive another driver class that issues protobufs/C++ darta structures
   instead of calling to Python builder.

I'd also like to reformat the lexer and grammar rules a bit to remove the
variable indent. Contemplating doing that in v2 so it's as close as
possible to v3 (even if the guts is a bit different).

At least on macOS and Linux, installing RE/flex is trivial, thus I would
not worry too much about adding a dependency on it.

dnicolodi:
Sounds like a good plan to me. I think this PR already implements a C++ lexer flee of most cruft. Please have a look.

If you plan to reformat the lexer and grammar rules I think it would be nice to do it on v2 too, so merging patches to the two versions does not get too difficult. I tried to clean up the formatting of the lexer rules a bit, but I haven't found a style I really like.

blais:
On Sun, Nov 22, 2020 at 2:17 PM Daniele Nicolodi <notifications@github.com>
wrote:

> Sounds like a good plan to me. I think this PR already implements a C++
> lexer flee of most cruft. Please have a look.
>
Yep, thank you. Well I'll have a quick shot at creating the fork and making
it work now, and either you or I can cherrypick the relevant bits from your
CL to apply to that new code parser. I'll put the code in beancount/cparser
(similar to beancount/ccore), and eventually all the Python code in there
will get ported to C++ in beancount/cparser.

(This will mark a departure of v2 of sorts; changes to the grammar in v2
would require changes to two parsers and three builder interfaces, so I'm
going to go largely reluctant to making grammar changes in v2 once that's
working.)


If you plan to reformat the lexer and grammar rules I think it would be
> nice to do it on v2 too, so merging patches to the two versions does not
> get too difficult. I tried to clean up the formatting of the lexer rules a
> bit, but I haven't found a style I really like.
>
I'll do that right now. It's all just whitespace changes, basically
removing unnecessary indentations (which have been getting on my nerves for
a while). Will ping you on it.

blais:
https://github.com/beancount/beancount/commit/66ada442b113f9c9f97f22de9137b3c959720269

blais:
Given the v3 cparser work, is there anything to keep from this PR?
Let me know.

dnicolodi:
I don't know if there anything of value in here anymore. Given your work in `cparser` in v3 I don't think so. Feel free to close this, I'll keep my branch, just in case.

blais:
Closing in light of changes in v3.

PR #582, opened 2020-11-22
Stricter parser

dnicolodi:
This is mostly a proposal to understand if something like this may be desirable. The idea would be to have a way to disable strict parsing in v2 and make it the only mode in v3. I decided to make it default and have an environment variable to disable it. If we agree that this is a good thing, this may be material for a 2.4 (or 2.90 or something) release.


blais:
Why don't we make such a change in v3?

dnicolodi:
The idea was to easy the path between v2 and v3 via the possibility to relax the parser to the old behavior in v2. Alternatively, we can make the stricter parser opt in in v2. If does do not seem good ideas, implementing it in v3 only is of course fine.

blais:
Yeah... I'm not sure we care to modify v2 at this point. v3 is the new new thing. 
If you'd like to make the v3 parser stricter let me know, let's VC.
I'm almost done with it, want to make a few important changes, once done with the parser, if you'd like to resume work on the scanner / parser, there's a few important things TBD, like adding parsing of comments for instance.

dnicolodi:
I think implementing the equivalent of the first commit in this series in v3 is an important cleanup of the syntax. I am happy to rebase these changes on v3 and to work on other aspects of parsing. Let's coordinate. I'm back in Europe, finding a convenient time for VC may be a bit more tricky now.

PR #589, opened 2020-12-14
Improve matching of lots with computed cost basis.

bayesianmind:
When basis is computed, either via the {{ }} total cost syntax, an
expression, or when using average cook booking one can easily get 
large fractional unit basis. If a user attempts to match that lot, they
need to know the exact number of digits beancount uses
for quantizing digits to match the lot. By using tolerance, users can
use specify a fractional cost in a human-readable way:
3.33 USD for 0.01 tolerance will match 10 / 3.

For example:
```
        2016-01-01 * #ante
          Assets:Account              3 HOOL {{ 10 USD }}
; can today only be reduced via:
          Assets:Account              -1 HOOL { 3.333333333333333333333333333 USD }

; but with this PR, the user can now write
         option "inferred_tolerance_default" "USD:0.01" 
         Assets:Account              -1 HOOL { 3.33 USD }
```

A downside of this change: If a user specifies an absurdly large
tolerance, it will make lot matching by cost overly broad.

Note, reusing tolerances on balance seems OK to me but has some
downsides. Specifically, we're using the configured tolerance for a currency,
which is currently used to resolve ambiguous precision when that context
is unavailble locally, per http://furius.ca/beancount/doc/tolerances and
`infer_tolerances()`. I think in practice these values are similar
enough to reuse this config, but am open to feedback.

### DIFF
--- a/beancount/parser/booking_full.py
+++ b/beancount/parser/booking_full.py
@@ -176,7 +176,7 @@ def _book(entries, options_map, methods):
                 # CostSpec instances below.
                 (booked_postings,
                  booking_errors) = book_reductions(entry, group_postings, balances,
-                                                   methods)
+                                                   methods, tolerances)
 
                 # If there were any errors, skip this group of postings.
                 if booking_errors:
@@ -502,7 +502,7 @@ def has_self_reduction(postings, methods):
 
 
 def book_reductions(entry, group_postings, balances,
-                    methods):
+                    methods, tolerances):
     """Book inventory reductions against the ante-balances.
 
     This function accepts a dict of (account, Inventory balance) and for each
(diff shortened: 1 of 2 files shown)


bayesianmind:
This PR shouldn't be merged until I add code to handle multiple lots matching the looser spec. We should take the one that's closer, or fail and ask the user for an exact match.

bayesianmind:
There's a better proposal in v3 for how to handle large decimal remainders in v3.
See https://beancount.github.io/docs/beancount_v3.html#make-rewriting-the-input-first-class:

```
What's more, in order for this to work accurately, we'd have to delay processing of the arithmetic operations post-parsing, so that we can render them back out. This offers another advantage: if we process the calculations after parsing, we can afford to provide an option to let the user specify the precision configuration to use for mpdecimal. I really like that idea, because it avoids hard-coding calculation precision and better defines the outcome of these options, potentially opening the door to a more rational way to remove extra digits that often get rendered out.
```

This is a more elegant solution than what I've proposed here.

PR #591, opened 2020-12-19
Implement AVERAGE booking and lot merging via {*}

### LINKED ISSUE #213: Implement average cost booking
**[Original report](https://bitbucket.org/blais/beancount/issue/213) by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

Complete the work on AVERAGE booking method.
See booking_methods.py, at the bottom.
(This is a tracking ticket.)

bayesianmind:
This PR adds the AVERAGE booking method. Since the current booking methods are only capable of disambiguating lots, some minor tweaks were needed to the booking algorithm. I've attempted to carry on Martin's excellent example and make those as maintainable and clear as possible.

Usage:
- If you add a * in a CostSpec, we will merge the matching inventory as part of applying that transaction. That works for both augmentations and reductions. This works no matter what booking method is set on the account.
- If you append "AVERAGE" to the open account directive, all postings to that account held at cost will be merged at average cost as you post changes to the account.
- When attempting to reduce an position held at average cost, it's recommended to let Beancount select the lot:
`Assets:Account -1 HOOL {}`  instead of `Assets:Account -1 HOOL {9.3333333333333333333 USD}`. The reason is that it's very easy to get long fractional digits and currently Beancount will require you to use the full precision to match the lot.  #589 uses tolerances to make matching more natural so one could say `Assets:Account -1 HOOL {9.33 USD}` instead.

For background on the implementation, see https://docs.google.com/document/d/1rclTmxJzq-4Na1UPY0XfS9Yzuv_SuHpdfHtabNsUH0E

Some important notes:
- There were 2 testcases such as `test_ambiguous__AVERAGE__mixed_currencies__unambi_date` that I disagree with the behavior. If I understand their intent, it is to provide the ability to match specific lots even when we are otherwise holding the account at AVERAGE. Implementing this would've been much more complex, and I am skeptical that such complex functionality is in line with the goals of AVERAGE cost booking.
- Likewise, I see some evidence that Martin intends to use a user-provided basis on reductions to modify the existing inventory. Consider the following transaction:
```
2020-11-16 * "Buy HOOL"
  Assets:Broker 1 HOOL {1 USD}
  Assets:Broker 1 HOOL {1.2 USD} ; price varies on the same day
 
2020-11-17 * "Sell HOOL"
  Assets:Broker -1 HOOL {1.5 USD} @ 2 USD ; not at all average basis
  Income:StockPnL:Taxable:HOOL  ; If indeed the 1.5 USD basis is right, we can compute this. BUT our ex-inventory will then be wrong! If 1.5 USD was a typo, the user will now have a wrong ex-inventory that might make a future correctly written transaction wrong. That'd be trickier to debug for the user.
  Assets:Checking 2 USD
 ; the ex-inventory is now 1 HOOL {0.7 USD} since the total basis was 2.2 USD and we reduced 1.5 USD of that. This is almost surely not what the institution has for the ex-inventory. If beancount's computed average is off it's likely due to missing history, so we should throw an error so the user can fix this.
```

I have not come up with a useful reason to support this. NOT supporting it lets us provide useful validation - the user can put the basis they expected they were selling the stock at and if it differs from our computations, it will fail to match and the error will show the average price we computed for that commodity.

Fixes #213


bayesianmind:
(closing and reopening to trigger CI)

bayesianmind:
Martin: It looks like you usually merge rather than rebase + squash. My commit messages were made with a squash workflow in mind - I'll squash on my branch shortly.

blais:
> Martin: It looks like you usually merge rather than rebase + squash. My commit messages were made with a squash workflow in mind - I'll squash on my branch shortly.

Sure, doesn't matter that much to me. I'm not super careful about history. If this was a team of 5+ people backing a commercial endeavor it might be different, but here I'm just forging ahead. And frankly it's unclear to me which is best anyway, not squashing preserves the detail of each commit, which can be useful. I don't mind either way.

blais:
I will review this before the new year; in the meantime, just curious, had you seen my old snippets for tests from a previous attempt creating this?
https://github.com/beancount/beancount/blob/a55431db6a58ec0faf7f568f056aea53e72974cd/experiments/average/average_test.py

bayesianmind:
Thanks Martin. I hadn't seen this doc but I believe all the scenarios there
are covered, other than what I think are orthogonal notes about the
currency inference process.

Let me know if you have any questions about the feature. I'm happy to VC
with you as well, and to take any questions that arise from folks using
this.

blais:
Sorry for delais; haven't forgotten you, just been busy closing the year and preparing for putting on new year positions.

beancount/parser/booking_full.py
```diff
@@ -195,24 +195,48 @@ def _book(entries, options_map, methods):
 
                 if interpolation_errors:
                     errors.extend(interpolation_errors)
+
```
blais, line 203: This file and function... is the heart of the beast and honestly this _book() function is the code I'm most concerned with changing in the entire codebase myself. I've had a lot of issues changing this code in the past, and the order in which things have to happen can have really quite subtle effects. This is a change that has potential breaks that are difficult to foresee.

Are you okay if I merge this in v3 only?
v3 is in working state (the C++ code hasn't been hooked up yet).

FWIW, I ran a before/after comparison between v2 and your branch (updated), and the output is the same.

beancount/parser/booking_full.py
```diff
@@ -195,24 +195,48 @@ def _book(entries, options_map, methods):
...
+                # Update the running balances for each account using the
+                # booked and interpolated values. Note that we could optimize away
+                # some of this in book_reductions() but we choose not to do so, as a
+                # sanity check that the direct aggregation of the final booked lots
+                # will compute the same result as that during the book_reductions()
+                # process.
+                for posting in inter_postings:
```
blais, line 205: In particular, this is changing the order of update of the balance inventory. 
Here's a question: can partial updates between currency groups influence each other? 
Could this potentially change the outcome of matching because the inventory state isn't the same as in the previous version?

beancount/parser/booking_full.py
```diff
@@ -570,6 +594,16 @@ def book_reductions(entry, group_postings, balances,
                 balance.is_reduced_by(units)):
                 # This posting is a reduction.
 
+                # If {*} if specified for this posting, re-book inventory at
+                # average cost before matching lots, so the match is more
+                # likely unambiguous.
```
blais, line 641: Please insert note on the side-effect on 'balance' here, it's totally not obvious.
I'd much rather have the caller apply it TBH

beancount/parser/booking_method.py
```diff
@@ -170,77 +168,82 @@ def booking_method_NONE(entry, posting, matches):
...
+            filtered_inv.add_position(inv_position)
+
+    if len(filtered_inv) <= 1:
+        return []
+
+    booked_postings = []
+    avg_position = filtered_inv.average().get_only_position()
+    # Negate existing inventory.
+    for inv_position in filtered_inv:
+        full_inventory.add_amount(-inv_position.units, inv_position.cost)
```
blais: Avoid this and the next update, and let the callers update their inventories.
(Avoid the subtle mutation.)

beancount/parser/booking_full.py
```diff
@@ -195,24 +195,48 @@ def _book(entries, options_map, methods):
...
+                            booking_method.rebook_inventory_at_average_cost(
+                                balances[inter_posting.account],
+                                inter_posting.account,
+                                inter_posting.units.currency,
+                                inter_posting.cost.currency,
+                            )
+                        )
+
                 repl_postings.extend(inter_postings)
+                repl_postings.extend(cost_merging_postings)
```
blais: The first line could move to before the block; the second one would be within the special loop and conditionally extend.
You could avoid 'cost_merging_postings' entirely.

beancount/parser/booking_full.py
```diff
@@ -195,24 +195,48 @@ def _book(entries, options_map, methods):
...
+
+                # If lot merging was requested, resolve it after interpolation.
+                # This allows interpolation to resolve an unspecified cost before
+                # it gets merged together into a merged inventory.
+                cost_merging_postings = []
+                for i, inter_posting in enumerate(inter_postings):
+                    # Interpolated postings are always Cost instead of CostSpec,
+                    # so we need to look back at the pre-interpolated postings
+                    # to check the merge flag.
+                    posting_cost = booked_postings[i].cost
```
blais: This assumes len(inter_postings) == len(booked_postings).
Is this always true?

beancount/parser/booking_full.py
```diff
@@ -195,24 +195,48 @@ def _book(entries, options_map, methods):
```
blais: Did you measure the performance diff?
I'm seeing little tiny speedup - strangely, which is concerning. 
Could be statistical noise.

beancount/parser/booking_full_test.py
```diff
@@ -1457,6 +1465,55 @@ def test_augment__from_empty__incomplete_cost__with_currency(self, entries, __):
...
+          Assets:Account          1 HOOL {3 USD}
+
+        2015-10-01 * #apply
+          ; This inventory is merged with all existing matching inventory into a single
+          ; aggregate lot.
+          Assets:Account          1 HOOL {5 USD, *}
+
+        2015-10-01 * #booked
+          Assets:Account          1 HOOL {5 USD, 2015-10-01}
+          Assets:Account          -1 HOOL {1 USD, 2015-10-01}
```
blais: I would insert some metadata, or at least a special flag, on all the inserted merging postings.

beancount/parser/booking_full_test.py
```diff
@@ -1726,7 +1783,42 @@ def test_reduce__missing_units_number(self, _, __):
...
+        """
+        2016-01-01 * #ante
+          Assets:Account          10 HOOL {10.00 USD, 2016-01-01}
+          Assets:Account          10 HOOL {20.00 USD, 2016-04-15}
+
+        2016-05-02 * #apply
+          Assets:Account          -5 HOOL {99 USD, *}
+
+        2016-05-02 * #booked
+          error: "No position matches"
```
blais, line 1820: Wellll that's an interesting case. We could allow an arbitrary amount to be deducted from the average cost when reducing. Not absolutely necessary,  but that could make sense.

Do you have a test for a pure cost basis adjustment? How would that look like?

beancount/parser/booking_full_test.py
```diff
@@ -2381,18 +2473,61 @@ def test_ambiguous__LIFO__test_matching_more_than_is_available(self, _, __):
         """
 
 
-@unittest.skip('Booking.AVERAGE is disabled.')
 class _TestBookAmbiguousAVERAGE(_BookingTestBase):
```
blais: Remove the leading _.

beancount/parser/booking_full_test.py
```diff
@@ -2424,17 +2594,14 @@ def test_ambiguous__AVERAGE__simple_merge2_match1(self, _, __):
           Assets:Account        -49 HOOL {}
 
         2015-06-01 * #booked #reduced
-          M Assets:Account      -50 HOOL {100.00 USD, 2015-10-01}
```
blais, line 2427: Explain this to make sure we're on the same wavelength... because the original two lots already got merged on application of the first transaction?
(I think I know why I used to have those: in the original test development, I assumed that I'd only be merging on reductions.)

blais (commented):
Great work! Spent a solid hour reviewing. Many comments.

beancount/parser/booking_full.py
bayesianmind: Yes absolutely, that makes sense to fully focus our efforts on v3. I changed the base branch to master.

beancount/parser/booking_full.py
bayesianmind: I didn't, are you using pytest instrumentation to measure the diff or a different way? Great idea to check that.

beancount/parser/booking_full_test.py
bayesianmind: I thought carefully about allowing an arbitrary amount to be deducted from the cost basis when reducing (thus repricing the remaining ante-inventory), especially since I saw several test cases indicating that's the behavior you had in mind when you put the test cases together. Unless I'm missing a compelling usecase for it, it seems much better to check the basis is close to our average basis instead.
I rewrote the original section from my google doc discussing this to be much more clear. Would you take a look at just that section and let me know if you still want it to work this way? https://docs.google.com/document/d/1rclTmxJzq-4Na1UPY0XfS9Yzuv_SuHpdfHtabNsUH0E/edit#bookmark=id.qd80c26c1n8u
Sections: "Using reducing basis as correctness verification", and "Cost basis adjustment - leads to confusing ante-inventory"
Thanks!

bayesianmind: I'm not sure what you mean by "pure cost basis adjustment", When would someone do that? Is it just someone adding a transaction that manually removes all the holdings at a particular price and re-adds them with a different basis?

beancount/parser/booking_full.py
bayesianmind: Good idea, done.

beancount/parser/booking_full.py
bayesianmind: It appears to be, but it does bother me that it could change to not be. Also it can differ when there are interpolation errors.

So I rewrote this to determine merge requirements once per group, which obviates the need to check the merge flag again here.

beancount/parser/booking_full_test.py
bayesianmind: Yes exactly, when the booking test sets up the ante-inventory, that is an augmentation that triggered rebooking at average. The #apply transaction then is just reducing that rebooked lot.
This especially makes the test_AVERAGE__mixed_currencies__unambi_cost_ccy__merging family of tests less specific. If the booking test runner had the ability to apply multiple postings as a unit I'd have used that rather than #ante to get more specific asserts.

beancount/parser/booking_full.py
bayesianmind: I don't think so, but it's sufficiently complex that I didn't want to take the chance. I've rewritten this logic to rely fully on a separately maintained local_balance that gets updated as we process the transactions. Then I added an assert that verifies that this local_balance matches the balance we get by applying the fully resolved transaction stream.

beancount/parser/booking_full.py
bayesianmind: Good point. I removed that side-effect and have the caller apply it instead.

beancount/parser/booking_full_test.py
bayesianmind: Done

beancount/parser/booking_method.py
bayesianmind: Done

bayesianmind (commented):
Thanks for your feedback Martin! This is ready for another look.

blais:
Hi Ben,
I'm sorry for the delay, your PR has been on my mind but really too busy to handle.
Tonight I had a solid second look at merging it and I came close to it;
I was going to, but upon review I think it needs a few more things, and I have another urgent task to handle, so I'll comment:
- I'd like to see some unit tests around that rebook_inventory_at_average_cost() method from booking_method. It worries me a bit, it's isolatable, should be easy.
- I'd also much prefer if we found a way to not mutate its inventory argument.
- In particular, its first conditional  doesn't check whether the posting has a non-null or not. I think in some cases this might fail. 
- The calling method (reboot_inventory_at_average() tripped me up at first because of the similarity of the name. I'd rename it.
- Nit: avoid using \ in Python (google style); I use parens instead (I can change this after merging, no big deal).

With the added test I would have merged this just now

bayesianmind:
No worries, thanks for your thorough review Martin! This should be ready to go now.

> * I'd like to see some unit tests around that rebook_inventory_at_average_cost() method from booking_method. It worries me a bit, it's isolatable, should be easy.

Done.

> * I'd also much prefer if we found a way to not mutate its inventory argument.

I believe this is already the case? `generate_rebook_at_average_postings` doesn't mutate the full_inventory argument anymore. I added a `local_inventory` in `booking_full` that tracks ongoing inventory as booking is proceeding, and then we assert it matches the booked inventory once booking+interpolation is complete. I need that for correctness.

> * In particular, its first conditional  doesn't check whether the posting has a non-null or not. I think in some cases this might fail.

Yes good call, added that.

> * The calling method (reboot_inventory_at_average() tripped me up at first because of the similarity of the name. I'd rename it.
> * Nit: avoid using \ in Python (google style); I use parens instead (I can change this after merging, no big deal).

Good call, renamed the method in booking_method to `generate_rebook_at_average_postings` and replaced the `\`.

beancount/parser/booking_full.py
```diff
@@ -145,6 +145,10 @@ def _book(entries, options_map, methods):
 
             # Resolve reductions to a particular lot in their inventory balance.
             repl_postings = []
+            # This local balance tracks changes while processing this transaction.
```
blais, line 148: Is this new? I suspect this is going to make booking much slower.
See comment below.

blais: So I timed it and I'm surprised it doesn't.
Let's merge it.
I think the next version we'll have to review and define inputs and outputs better so we don't have to compute partial balances like this.

blais:
```
bergamot [git|v2]:~/p/beancount$ hyperfine -i -m 25  -- "bean-check -C $L"
Benchmark #1: bean-check -C /home/blais/r/q/office/accounting/blais.beancount
  Time (mean ± σ):      5.617 s ±  0.971 s    [User: 5.485 s, System: 0.085 s]
  Range (min … max):    4.609 s …  7.947 s    25 runs

  Warning: Ignoring non-zero exit code.
bergamot [git|master]:~/p/beancount$ cat /tmp/after 
bergamot [git|master]:~/p/beancount$ hyperfine -i -m 25  -- "bean-check -C $L"
Benchmark #1: bean-check -C /home/blais/r/q/office/accounting/blais.beancount
  Time (mean ± σ):      5.647 s ±  0.455 s    [User: 5.501 s, System: 0.090 s]
  Range (min … max):    4.709 s …  6.406 s    25 runs
```

bayesianmind: It's new the the revision I did when making the average cost calculator avoid mutating the inventories.
Thanks for checking the latency. I agree on doing some work here in the future. I was thinking about how to do lot transfer for the crypto case and making booking more configurable in general. An example that would be nice is being able to write a tax optimizing booker that auto selects losses before gains. Both of those need clarity

blais:
Okay, this is not working. I must have made an error in benchmarking, because my booking which is normally 1.5 seconds now clocks in at 25s (!). I can't use the software. I'm going to have to roll this back.

blais:
I rolled this back and kept the merged head to branch average_booking_rollback.

PR #592, opened 2020-12-19
Add google_auth_oauthlib to requirements to fix CI

bayesianmind:
Fix CI by adding a missing dependency required by `beancount/tools/sheets_upload.py`

This could also be just added to requirements_dev.txt if preferred. I put it in requirements.txt based on google-api-python-client already being present.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ bottle>=0.12
 ply>=3.4
 lxml>=3.0
 google-api-python-client>=1.8.2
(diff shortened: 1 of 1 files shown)


blais:
Thanks!

jamielynn333143-code:
Hacked

jamielynn333143-code:
Hacked

jamielynn333143-code:
Hacked

PR #595, opened 2020-12-23
Provide SSL context on request.urlopen()

gpoul:
This fixes an error returned with newer OpenSSL default settings
with some beanprice sources where a URLError is now raised with message
<urlopen error [SSL: DH_KEY_TOO_SMALL] dh key too small (_ssl.c:1123)>

### DIFF
--- a/beancount/utils/net_utils.py
+++ b/beancount/utils/net_utils.py
@@ -6,7 +6,7 @@ __license__ = "GNU GPLv2"
 import logging
 from urllib import request
 from urllib import error
-
+import ssl
 
 def retrying_urlopen(url, timeout=5, max_retry=5):
     """Open and download the given URL, retrying if it times out.
@@ -22,10 +22,13 @@ def retrying_urlopen(url, timeout=5, max_retry=5):
     for _ in range(max_retry):
         logging.debug("Reading %s", url)
         try:
-            response = request.urlopen(url, timeout=timeout)
+            ctx = ssl.create_default_context()
+            ctx.set_ciphers('DEFAULT@SECLEVEL=1')
+            response = request.urlopen(url, timeout=timeout, context=ctx)
             if response:
                 break
-        except error.URLError:
+        except error.URLError as err:
+            logging.error("Error when opening URL: {0}".format(err))
             return None
     if response and response.getcode() != 200:
         return None


blais:
Hmmm this is odd. I can't quite believe the new default way of doing things requires all client code to do this.
Can you provide some supporting context?

dnicolodi:
I belive this comes from the new host wide OpenSSL policy implemented by RedHat. I don't think the fix belongs in all the clients that use OpenSSL but rather in the policy itself.

blais:
Hmm, I don't know what to think.
Well if you can motivate it... 
in the face of a 1000 other more urgent things, well

gpoul:
I saw this when running bean-price on Fedora 33; I researched why it is not working and found that you're ignoring the exception that is raised, so I logged it and looked at the error message and found this also to be reported by a few Ubuntu users for other systems, so it might not be totally RedHat-specific.

One of the resources I found was for example this: https://stackoverflow.com/questions/38015537/python-requests-exceptions-sslerror-dh-key-too-small

There seems to be a way of providing a custom openssl config, but the fact is that bean-price is not able to retrieve oanda prices on Fedora 33 in the default config, at least on my end. I think the only "reliable" way of fixing this is to either change this in the distribution or on the application side.

I only wanted to raise this; it's definitely your decision how you'd like to handle this. We can also keep this here and see whether others report the same thing in the future.

dnicolodi:
OpenSSL refuses the remote certificate because of host wide policies installed of the by the administrator or by the OS vendor. I don' think it is desirable nor appropriate for single applications to override this policy if there are no specific reasons (ie compliance with a specific protocol). If you do not agree with the choices of the system administrator or OS vendor, raise the issue with them (or change your system settings).

blais:
What is we just issued a warning pointing users at what they need to do on
their systems too fix it, or at least the reason why it's failing?

dnicolodi:
I haven't looked at it in detail, but I suspect it is difficult to detect whether OpenSSL refuses connection because some too strict local settings or because of a genuine problem with the connection or the certificate trust chain (the exception raised seems to be a generic `SSLError`). I guess that soon either the affected systems will be upgraded to a less restrictive policy, or the servers will be upgraded to "better" certificates. I would not spend much effort on this.

blais:
After reading about the context, agree with dnicolodi.

PR #610, opened 2021-01-15
cparser: std::filesystem is not experimental anymore in c++17

dnicolodi:
Other c++17 features like std::string_view are already used in the codebase. There is no reason to do not use this one too.

PR #612, opened 2021-01-15
Meson

dnicolodi:
Technology preview of how the Meson build looks like.

I gave up on integrating the Bison build, it is a build time only dependence and I compile it on the build machine. It takes about 1 minute to build and install, notably with a sizeable fraction of it spent in the configure script. This time can be reduced with some clever use of the cache or producing a deb package for Bison 3.7.4 for Ubuntu 20.04.

It also uses an oldish version of protobuf, 3.12.2. Updating to a newer release should be easy. However, tests pass with the old version too.

This does not build the Python protobuf modules or a Python package, mostly because I haven't yet understood how to test it.

It would be nice to see if this does something sensible on Windows, but I don't know if there is an easy way to get Bison on Windows.

Oh, this includes the commit form #610 to make things a tiny bit easier.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,172 @@
+project('beancount', 'cpp',
+        version : '2.90.0',
+        license : 'GPL-v2-only',
+        default_options: [
+          'cpp_std=c++17',
+          'optimization=2',
+          'default_library=static',
+        ])
+
+absl_container_dep = dependency('absl_container')
+absl_hash_dep = dependency('absl_hash')
+absl_status_dep = dependency('absl_status')
+absl_strings_dep = dependency('absl_strings')
+absl_time_dep = dependency('absl_time')
+
+mpdecimal_dep = dependency('mpdecpp')
+
+re2_dep = dependency('re2')
+
+pybind11_dep = dependency('pybind11')
+
+gtest = subproject('gtest')
+gtest_main_dep = gtest.get_variable('gtest_main_dep')
+gmock_dep = gtest.get_variable('gmock_dep')
+
+# FIXME: Fix protobuf wrap to do not require this. Also update to latest protobuf release.
+protobuf_dep = dependency('protobuf', required : false)
+if protobuf_dep.found()
+  protoc = find_program('protoc')
+else
+  protobuf = subproject('protobuf')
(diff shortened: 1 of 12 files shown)


blais:
Would it make sense to put each of the .wrap files under //third_party/..../ next to their Bazel counterparts?
Hmm, Maybe not.

How about //third_party/meson instead of subprojects?

dnicolodi:
The `subprojects` directory is configurable, but it is very discouraged to do so. The option can actually go away in future releases of Meson. I think that, adding a README in the directory that explains what it is about, a separate `subprojects` directory may be less confusing.

dnicolodi:
I dropped the `std::filesystem` change and fixed the build for old compilers

meson.build
```diff
@@ -0,0 +1,172 @@
...
+      deps,
+      gmock_dep,
+      gtest_main_dep,
+    ])
+
+  test(test.split('.')[0], t, protocol : 'gtest')
+endforeach
+
+python.extension_module(
+  'beancount.ccore._core',
```
eli-schwartz, line 145: This seems wrong, you should be installing a `_core` extension_module into the `beancount.ccore` install_dir.

Python doesn't expect `site-packages/beancount.ccore._core.{SOEXT}`, but rather, expects `site-packages/beancount/ccore/_core.{SOEXT}`

dnicolodi: Indeed. Thank you @eli-schwartz for having a look. When I was working on this I used a local hack that uses Meson as a Setuptools backend for compilation to install or produce wheels and it handled placing the extension in the right place. I cannot remember now if naming the extension like that was required to make the hack work. I definitely should revisit that and publish it somewhere.

By the way, do you have an interest on Beancount or are you reviewing all PRs on Github that mention Meson? 😃

eli-schwartz: I'm (trying to) review all PRs on GitHub that mention Meson. 😀

blais:
This should be moved to the cpp branch. 
Happy to accept it in the branch (though C++ is likely gone).

blais:
We're using Meson for the Python build now. 
Is there any value is keeping this PR or can it easily be redone by AI for a more recent version of the code in the future?
(

PR #613, opened 2021-01-15
validation: allow reopening closed accounts

### LINKED ISSUE #611: Allow reopening an account ?
For quick debts with friends, I'd like to just dynamically open and close debt accounts in my ledger, like so:

```
2021-01-15 open Liabilities:Dette:Firstname-Lastname

2021-01-15 * "Firstname bought me lunch"
  Liabilities:Dette:Firstname-Lastname    -8.00 EUR
  Expenses:Food

2021-01-15 * "Firstname Lastname" "Reimbursement for lunch"
  Assets:Checking                         -8.00 EUR
  Liabilities:Dette:Firstname-Lastname

2021-01-16 balance Liabilities:Dette:Firstname-Lastname 0.00 EUR

2021-01-16 close Liabilities:Dette:Firstname-Lastname
```

Unfortunately that only works the first time, if I have a second debt with the same person then I can't reopen that account, I get a `Duplicate open directive for...` error.

[Relevant code here](https://github.com/beancount/beancount/blob/41f0d5209f92962fe8857ff3813b74fb413615e6/beancount/ops/validation.py#L70-L78)

Is this really a behavior you want ? I can definitely think of situations where reopening a previously closed account makes sense (the simplest one being, you change banks, then come back to your previous bank, for whatever reason).

It definitely makes the implementation of the validation a bit more complex though, but from what I can see the code linked above already assumes that the list of entries is sorted chronologically (otherwise validating `close` directives would need to process all `open` directives first).

alarsyo:
Fixes #611

First draft to allow re-opening a previously closed account. This only prevents the error from appearing, but allowing this may have effects elsewhere, I'll play around with my ledger to see if anything breaks!

Feedback welcome, feel free to point me to places where additional tests should be written for this use case.

### DIFF
--- a/beancount/ops/validation.py
+++ b/beancount/ops/validation.py
@@ -69,11 +69,17 @@ def validate_open_close(entries, unused_options_map):
 
         if isinstance(entry, Open):
             if entry.account in open_map:
-                errors.append(
-                    ValidationError(
-                        entry.meta,
-                        "Duplicate open directive for {}".format(entry.account),
-                        entry))
+                close_entry = close_map.get(entry.account, None)
+                if close_entry is not None and close_entry.date < entry.date:
+                    # reopening closed account
+                    del close_map[entry.account]
+                    open_map[entry.account] = entry
+                else:
+                    errors.append(
+                        ValidationError(
+                            entry.meta,
+                            "Duplicate open directive for {}".format(entry.account),
+                            entry))
             else:
                 open_map[entry.account] = entry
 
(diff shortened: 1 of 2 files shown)


alarsyo:
Nothing obvious seems broken, at least with the use case I described in #611!

blais:
This change would be a can of worms down the road, sorry. Create an second account name, or keep the account open, it's the right thing to do anyway.

Issue #598, opened 2020-12-28
Too many missing numbers on valid FIFO transaction should resolve.


blais:
```
option "booking_method" "FIFO"
plugin "beancount.plugins.book_conversions" "Assets:Bitcoin,Income:Bitcoin"
plugin "beancount.plugins.auto_accounts"

option "insert_pythonpath" "True"
plugin "plugins.transfer_lots" "transfer"

2020-12-01 * ""	"Buy BTC"
          Assets:Cash:Coinbase-Pro         -1500.00 EUR
          Assets:Crypto:BTC:Coinbase-Pro       0.15 BTC	{10,000 EUR} @ 10,000 EUR

2020-12-17 * ""	"Buy BTC"
          Assets:Cash:Coinbase-Pro        -6,000.00 EUR
          Assets:Crypto:BTC:Coinbase-Pro        0.3 BTC	{20,000 EUR} @ 20,000 EUR

2020-12-18 * ""	"Transfer BTC" #transfer
          Assets:Crypto:BTC:Coinbase-Pro     -0.175 BTC {}
          Assets:Crypto:BTC:Binance         

2020-12-20 * "" "Swap BTC => USDT (1/2)"
; assumed BTC sell
          Assets:Crypto:BTC:Binance           -0.16 BTC {} @ 16,000 EUR
          Assets:Cash:Binance                 2,560 EUR
          Income:Trading
```

javiertury:
I was hitting a similar issue.

#615 improves inference in the case that both costspec.number_per and costspec.number_total are missing. Note however that it doesn't solve every problem that you stated here.

EDIT: I think that I misunderstood how cost is represented: number_per represents variable cost and number_total represents fixed cost (not total cost).

Philogy:
Also ran into this issue, extremely frustrating, took my like an hour to understand what was happening and had to fudge dates on entries as result

PR #615, opened 2021-01-24
Infer number_per and number_total simultaneously in costspec #598

javiertury:
This is a small inference improvement for #598. It treats costspec quantities (number_per and number_total) as a single datum and makes beancount inference work in case both of these quantities are missing instead of raising a "Too many missing numbers" error.

### DIFF
--- a/beancount/parser/booking_full.py
+++ b/beancount/parser/booking_full.py
@@ -712,9 +712,8 @@ def convert_costspec_to_cost(posting):
 class MissingType(enum.Enum):
     """The type of missing number."""
     UNITS = 1
-    COST_PER = 2
-    COST_TOTAL = 3
-    PRICE = 4
+    COST = 2
+    PRICE = 3
 
 
 # An error raised if we are not able to interpolate.
@@ -757,10 +756,11 @@ def interpolate_group(postings, balances, currency, tolerances):
             incomplete.append((MissingType.UNITS, index))
 
         if isinstance(cost, CostSpec):
-            if cost and cost.number_per is MISSING:
-                incomplete.append((MissingType.COST_PER, index))
-            if cost and cost.number_total is MISSING:
-                incomplete.append((MissingType.COST_TOTAL, index))
+            if cost and (
+                cost.number_per is MISSING
+                or cost.number_total is MISSING
+            ):
+                incomplete.append((MissingType.COST, index))
         else:
             # Check that a resolved instance of Cost never needs interpolation.
             #
@@ -862,30 +862,27 @@ def interpolate_group(postings, balances, currency, tolerances):
             else:
                 new_posting = None
 
-        elif missing == MissingType.COST_PER:
+        elif missing == MissingType.COST:
             units = incomplete_posting.units
             cost = incomplete_posting.cost
-            assert cost.currency == weight_currency, (
-                "Internal error; residual currency different than missing currency.")
-            if units.number != ZERO:
-                number_per = (weight - (cost.number_total or ZERO)) / units.number
-                new_cost = cost._replace(number_per=number_per)
-                new_pos = Position(units, new_cost)
-                new_posting = incomplete_posting._replace(units=new_pos.units,
-                                                          cost=new_pos.cost)
-            else:
-                new_posting = None
 
-        elif missing == MissingType.COST_TOTAL:
-            units = incomplete_posting.units
-            cost = incomplete_posting.cost
-            assert cost.currency == weight_currency, (
-                "Internal error; residual currency different than missing currency.")
-            number_total = (weight - cost.number_per * units.number)
-            new_cost = cost._replace(number_total=number_total)
-            new_pos = Position(units, new_cost)
-            new_posting = incomplete_posting._replace(units=new_pos.units,
-                                                      cost=new_pos.cost)
+            if units.number == ZERO:
+                new_posting = None
+            else:
+                assert cost.currency == weight_currency, (
+                    "Internal error; residual currency different than missing currency.")
+
+                cost_per_unit = weight / units.number
+
+                new_cost = Cost(
+                    number=cost_per_unit, currency=cost.currency,
+                    date=cost.date, label=cost.label
+                )
+                new_pos = Position(units, new_cost)
+                new_posting = incomplete_posting._replace(
+                    units=new_pos.units,
+                    cost=new_pos.cost
+                )
 
         elif missing == MissingType.PRICE:
             units = incomplete_posting.units


javiertury:
Just one warning. I assume that when both "number_per" and "number_total" are specified, then quantities are sumed like this:
```
weight = number_total + units*number_per
```

That is, beancount changes the meaning of the variables. "number_per" is still the variable cost but "number_total" behaves like a fixed cost instead of being the total cost.

EDIT: After thinking again about it, perhaps I got it wrong. If "costspec.number_per" and "costspec.number_total" **always** represent variable and fixed costs, then interpolation when both are missing is not possible since infinite representations are possible (1*n+0, 2*n-10, 3*n-20, ...). Then this PR should be closed.

blais:
Why close?

javiertury:
Let's start by sharing how I view the whole interpolation process. In abstract, I think this is the equation to interpolate:

```
residual_weights = units * weight_distribution_per_unit
```

The `weight_distribution_per_unit` is a category of **interchangeable** bundles of variables. Each of these bundles of variables corresponds to a different class representation: Price, Cost and CostSpec.

```
residual_weights = units * price_per_unit
residual_weights = units * cost.number
residual_weights = units * (costspec.number_per + costspec.number_total / units)
```

All the equations above should be equivalent. Moreover I think that CostSpec has only one purpose: to create later a Cost instance. If that's true, we don't care how weights are distributed between `costspec.number_per` and `costspec.number_total`, we only care about being able to map them to a `cost.number`. The equations of interest are then reduced to these:

```
residual_weights = units * price_per_unit
residual_weights = units * cost.number
```

Assuming that `residual_weights` is always known, the following interpolations are possible.

| Known variable | Interpolable variables          |
| -------------------- | ------------------------------------- |
| units                 | cost.number; price_per_unit |
| price_per_unit  | units; cost.number                |
| cost.number     | units; price_per_unit             |

Remember that we had made a "CostSpec -> Cost" substitution. We have to reverse it on the "known variables" side.

| Known bundle of variables                                  | Interpolable variables           |
| -------------------------------------------------------------- | -------------------------------------- |
| units                                                                    | cost.number; price_per_unit |
| price_per_unit                                                     | units; cost.number                |
| (costspec.number_per, costspec.number_total) | units; price_per_unit             |

If any bundle of variables is known, the rest can be trivially interpolated. If no bundle of variables is known, the equations become indeterminate and interpolation is not feasible.

In practice, these steps should help `interpolate_group` achieve maximum interpolation efficiency:
  1. Bundle `costspec.number_total` and `costspec.number_per` together for interpolation. This is what I submitted to do in this PR.
  2. Relax the assertions so that "units" can interpolate "price_per_unit" and "cost.number" simultaneously; "price_per_unit" can interpolate "units" and "cost.number" simultaneously; and "(costspec.number_per, costspec.number_total)" can interpolate "units" and "price_per_unit" simultaneously.

After an analysis similar to the one exposed so far, I was fairly confident and decided to apply those changes. I was able to implement step 1 without much problem. In retrospect, the code would look cleaner if I had built a Cost class directly, instead of creating another intermediary CostSpec class.

Then I tried to implement step 2 by making cost.number and price_per_unit simultaneously interpolable. I found an assertion in the existing code that contradicted the analysis I had made earlier. I'm referring to `Cannot infer price for postings with units held at cost` when "units" is known but "cost.number" and "price_per_unit" are unknown.

https://github.com/beancount/beancount/blob/7a8decf369a97b3fea6fa9e58bb4d894e4e45b23/beancount/parser/booking_full.py#L890-L904

At these point I was increasingly suspicious, thinking that I overlooked some unknown factors that could invalidate this analysis. I became unsure about implementing step 2, and decided to submit only a patch to bundle CostSpec variables (step 1). Later I started worrying about bundling CostSpec variables too. I realized that `costspec.number_per` and `costspec.number_total` could be used simultaneously and in that case their interpretation was a little bit different than I believed. Also I had assumed that the only purpose of CostSpec was to create a Cost instance. If the purpose of CostSpec was not limited to create a Cost instance, then `costspec.number_per` (variable costs) and `costpec.number_total` (fixed costs) would have a meaning on their own, and information about cost distribution would be lost when transformed to a simple Cost instance. This would mean that this interpolation method is incomplete and it's arbitrarily distributing costs between `costspec.number_per` and `costspec.number_total`.

I decided to close this PR because I became increasingly uncertain about its usefulness and correctness. However my disposition to collaborate is still firm.

blais:
On Sun, Jan 31, 2021 at 8:57 AM javiertury <notifications@github.com> wrote:

> Let's start by sharing how I view the whole interpolation process. In
> abstract, I think this is the equation to interpolate:
>
> residual_weights = units * weight_distribution_per_unit
>
> I don't know what that is.


> The weight_distribution_per_unit is a category of *interchangeable*
> bundles of variables. Each of these bundles of variables corresponds to a
> different class representation: Price, Cost and CostSpec.
>
> residual_weights = units * price_per_unit
> residual_weights = units * cost.number
> residual_weights = units * (costspec.number_per + costspec.number_total / units)
>
> The first one I call is "the value."
The second one is "the cost."
For a posting, what I've called "the weight" is either (1) or (2),
depending on whether there's a cost basis or noit.

The third is a correct formula for the number the cost-spec is designed to
express.


> All the equations above should be equivalent.
>
Price and cost are (in general) distinct.

Moreover I think that CostSpec has only one purpose: to create later a Cost
> instance.
>
Correct. The parser constructs a cost spec, which the booking code has to
match against lots of the inventory for that account just before the
transaction.


> If that's true, we don't care how weights are distributed between
> costspec.number_per and costspec.number_total, we only care about being
> able to map them to a cost.number.
>
I don't understand this sentence. There's no distribution.
The only purpose of having a CostSpec is that the parser reads that - it
allows the user to specify either a per-unit or total amount to be
distribution - and it converts that to a Cost instance.
As a user of the API, you should pretty much never haver to deal with a
CostSpec, it's an internal, temporary, implementation detail.


> The equations of interest are then reduced to these:
>
> residual_weights = units * price_per_unit
> residual_weights = units * cost.number
>
> Assuming that residual_weights is always known, the following
> interpolations are possible.
> Known variable Interpolable variables
> units cost.number; price_per_unit
> price_per_unit units; cost.number
> cost.number units; price_per_unit
>
> Remember that we had made a "CostSpec -> Cost" substitution. We have to
> reverse it on the "known variables" side.
> Known bundle of variables Interpolable variables
> units cost.number; price_per_unit
> price_per_unit units; cost.number
> (costspec.number_per, costspec.number_total) units; price_per_unit
>
> If any bundle of variables is known, the rest can be trivially
> interpolated. If no bundle of variables is known, the equations become
> indeterminate and interpolation is not feasible.
>
> In practice, these steps should help interpolate_group achieve maximum
> interpolation efficiency:
>
>    1. Bundle costspec.number_total and costspec.number_per together for
>    interpolation. This is what I submitted to do in this PR.
>
> Oh, now I understand.


>
>    1. Relax the assertions so that "units" can interpolate
>    "price_per_unit" and "cost.number" simultaneously; "price_per_unit" can
>    interpolate "units" and "cost.number" simultaneously; and
>    "(costspec.number_per, costspec.number_total)" can interpolate "units" and
>    "price_per_unit" simultaneously.
>
> Yes, that's a good idea, but price is only ever used either
a. To make a conversion when there's no cost basis (the number can be
interpolated)
b. Purely as an annotation for when there's no cost basis.

For (b), that's not the entire picture, as the
https://github.com/beancount/beancount/blob/master/beancount/plugins/sellgains.py
plugin actually books against a subset of the postings, so in theory, if
this were to integrated as a core rule, it could be used to interpolate the
price.

That's an interesting idea actually. Hadn't occurred to me.


>
>
> After an analysis similar to the one exposed so far, I was fairly
> confident and decided to apply those changes. I was able to implement step
> 1 without much problem. In retrospect, the code would look cleaner if I had
> built a Cost class directly, instead of creating another intermediary
> CostSpec class.
>
> Then I tried to implement step 2 by making cost.number and price_per_unit
> simultaneously interpolable. I found an assertion in the existing code that
> contradicted the analysis I had made earlier. I'm referring to Cannot
> infer price for postings with units held at cost when "units" is known
> but "cost.number" and "price_per_unit" are unknown.
>
>
> https://github.com/beancount/beancount/blob/7a8decf369a97b3fea6fa9e58bb4d894e4e45b23/beancount/parser/booking_full.py#L890-L904
>
> At these point I was increasingly suspicious, thinking that I overlooked
> some unknown factors that could invalidate this analysis. I became unsure
> about implementing step 2, and decided to submit only a patch to bundle
> CostSpec variables (step 1). Later I started worrying about bundling
> CostSpec variables too. I realized that costspec.number_per and
> costspec.number_total could be used simultaneously and in that case their
> interpretation was a little bit different than I believed. Also I had
> assumed that the only purpose of CostSpec was to create a Cost instance. If
> the purpose of CostSpec was not limited to create a Cost instance, then
> costspec.number_per (variable costs) and costpec.number_total (fixed
> costs) would have a meaning on their own, and information about cost
> distribution would be lost when transformed to a simple Cost instance. This
> would mean that this interpolation method is incomplete and it's
> arbitrarily distributing costs between costspec.number_per and
> costspec.number_total.
>
> I decided to close this PR because I became increasingly uncertain about
> its usefulness and correctness. However my disposition to collaborate is
> still firm.
>
Amazing.

This is all being rewritten in v3, and will be rewritten in C++ so I avoid
changes to v2 at this moment. It'll take a while until I get to the booking
code though, as I'm really busy and only limited time to code.

I'll keep this here, this is a good discussion with useful ideas, and I
think what you're suggesting provides some clarity for the next
implementation.
Thank you Javier,

javiertury:
>> If that's true, we don't care how weights are distributed between costspec.number_per and costspec.number_total, we only care about being able to map them to a cost.number.

> I don't understand this sentence

One of the difference between CostSpec and Cost is how they represent information about costs. CostSpec has information about variable cost-per-unit ("number_per") and fixed cost shared by all units ("number_total"). On the other hand Cost just stores a summarized cost-per-unit ("cost.number"). CostSpec provides more detailed information because it's able explain how costs are "distributed" between variable costs ("number_per") and fixed costs ("number_total"). For instance, buying 10 shares for 20 USD and a fee of 2 USD can be represented as a `CostSpec(number_per=2, number_total=2, currency='USD')` or as `Cost(number=2.2, currency='USD')`. Any complete CostSpec can be converted to a Cost instance without losing information, but the reverse is not true.

Let's say we know there are 10 units and that `Cost(number=2.2, currency='USD')`. We want to know how much we paid for each share (variable cost) and how much we paid for fees (fixed cost), that is, how costs are "distributed". There is no way to find out. But since you confirmed that the only purpose of CostSpec is to create a Cost instance, then we don't have to worry about that problem. If beancount internally uses only Cost, it's because it cares only about "cost.number" and not about "number_per" or "number_total".

> Price and cost are (in general) distinct.

> price is only ever used either
> a. To make a conversion when there's no cost basis (the number can be interpolated)
> b. Purely as an annotation for when there's no cost basis.

> For (b), that's not the entire picture, as the https://github.com/beancount/beancount/blob/master/beancount/plugins/sellgains.py plugin actually books against a subset of the postings

Thank you so much for this explanation, now I think I understand some things.

Only cost equations are precise and allow interpolation without noise.

```
residual_weights = units * cost.number
residual_weights = units * (costspec.number_per + costspec.number_total / units)
```

Price interpolation is a completely different beast. It's just an approximation. That's why price is used only as a last resort in interpolations.

```
price ≈ abs( residual_weights / units )                  # booking_full.py
price ≈ non_income_residual_weights / units      # sellgains.py
```

There are two different approximations and that's fine. They just target different situations. The `sellgains.py` approximation seems more accurate for closing long or short positions with gains. The approximation in `booking_full.py` is accurate for opening long or short positions of positively-priced securities, since there are no capital gain or losses at opening.

This begs the question, can the price estimator be generalized? These are the remaining cases.

- Closing long positions with losses
- Closing short positions with gains
- Closing short positions with losses
- Opening long or short positions of negatively-priced securities

In general they all follow the pattern `price ≈ proceeds / units`, where proceeds are the residual weights excluding gains and losses. If we wanted the price estimate to include fees, only capital gains and losses should be excluded from residual weights. Imagine closing a long position with losses, `Expenses:Fees` have to be differentiated somehow from `Expenses:CapitalLoses`. On the other hand, if we want the raw market price, all income and expenses accounts should be excluded. This last alternative is feasible to compute, since income and expenses accounts are known but the subaccounts used for capital gains and losses are unknown.

```
price ≈ non_income_or_expenses_residual_weights / units
```

This is just a general approximation that should not be very difficult to compute. But each user probably has different expectations about what should be included in the price and is ultimately responsible for providing prices that are adequate to their specific context. I imagine there are more contexts than beancount could hardcode.

Coming back to the debate about which varaibles are interpolable, I think the situation is quite similar to that of my previous comment. Now we have the following equations and approximations for interpolations

```
residual_weights = units * cost.number
residual_weights = units * (costspec.number_per + costspec.number_total / units)
price ≈ non_income_or_expenses_residual_weights / units
```

Which reduce to

```
residual_weights = units * cost.number
price ≈ non_income_or_expenses_residual_weights / units
```

For interpolation, we try to use first the "cost" equation, otherwise we fallback to the price approximation.

```python
if units is MISSING and cost_number is MISSING and price is MISSING:
  raise Exception('Not enought information to interpolate')

if units is MISSING:
  units = (
    residual_weights / cost_number if cost_number is not MISSING
    else non_income_or_expenses_residual_weights / price
  )

if cost_number is MISSING:
  cost_number = residual_weights / units
  
if price is MISSING:
  price = non_income_or_expenses_residual_weights / units
  
# All variables "units", "cost_number" and "price" are interpolated
```

javiertury:
Now that I'm sure CostSpec is only used to create a Cost instance, I've reopened the PR. It bundles costspec.number_total and costspec.number_per.

PR #616, opened 2021-01-25
Use Click library for command line parsing

dnicolodi:
This converts the ingest mechanism and `bean-doctor` to use Click for command line parsing. Other scripts can be converted later. I did some copy editing of the help strings for the commands of `bean-doctor` but some more could be done.

I haven't decided yet if it is better to name the file argument `FILENAME` or `LEDGER`. The first may be a tiny bit clearer, but the second allows for nicer help text: `Read transaction from ledger FILENAME` vs `Read transactions from LEDGER`.

I hope I haven't introduced any bug rebasing the PR.

### DIFF
--- a/beancount/scripts/doctor.py
+++ b/beancount/scripts/doctor.py
@@ -7,12 +7,12 @@ __copyright__ = "Copyright (C) 2014-2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
 import collections
+import logging
 import os
 import re
 import sys
-import argparse
-import logging
-from os import path
+
+import click
 
 from beancount import loader
 from beancount.core import account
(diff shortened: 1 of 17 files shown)


dnicolodi:
Turns out that the commit reworking the ingest scripts needs some more work. I don't know if it is worth spending time on it given that this part of the code is going to be simplified substantially soon. @blais Are you still planning to spin off the ingest code to its own repository?

blais:
Definitely going to move ingest out of there, and the subcommand part will change as well (I'm going to remove the binaries).
Don't worry about converting those.

dnicolodi:
I dropped the commit dealing with the ingest framework.

dnicolodi:
I fixed the tests and  couple of unrelated things.

dnicolodi:
Will need to be rebased once #619 is merged.

dnicolodi:
Rebased. This should leave only the ingest framework using argparse. I plan to submit a separate PR that removes the compatibility layer for `bean-extract` and friends.

beancount/scripts/doctor.py
```diff
@@ -7,67 +7,119 @@
...
 from beancount.core import convert
 from beancount.core import data
-from beancount.core import display_context
+from beancount.core.display_context import Align
 from beancount.core import getters
 from beancount.core import inventory
 from beancount.core import prices
 from beancount.core import realization
-from beancount.parser import context
+from beancount.parser.context import render_file_context
```
blais, line 28: It's an artifact of using Google style for so many years that I tend to favor always importing packages over functions directly.

dnicolodi: I would not have changed it if there would not have been a collision with a function with the same name. I though that the easier solution was to reduce the import scope.

blais:
Oh it's really not that big a deal
I'm not super consistent either :-)
Just mentioning it
The reason they like to do that at G is so using libraries can patch code
for testing. It's part of the style guide. I don't apply it consistently in
my own code. Basically if it's something low level like D or a ZERO
constant I break the rules and allow myself the syntactic convenience. And
event a wee bit more. But compared to open source code you will find my
code has a heavy bias towards importing packages and using dotted syntax. I
always feel like a thief when I do it now :-)

PR #620, opened 2021-01-26
bean-extract, bean-file, bean-identify: Remove deprecated scripts

dnicolodi:
Remove the dedicated binaries entry points to the ingest framework. This allows substantial simplification and better extensibility of the ingestion framework.

Adjust the tests that relied on the removed interfaces.

This is a first pass at removing some cruft, there is probably more.


beancount/ingest/extract_test.py
```diff
@@ -18,10 +18,7 @@
 from beancount.ingest import extract
 from beancount.ingest import importer
 from beancount.ingest import scripts_utils
-
-
-extract_main = functools.partial(scripts_utils.trampoline_to_ingest, extract)
-
+from beancount.ingest.test_utils import TestScriptsBase, TestExamplesBase
```
blais, line 21: Favor importing package, not symbols within, most of the time,.

blais (approved):

beancount/ingest/extract_test.py
dnicolodi: Noted. I don't recall why I wrote this import like this.

PR #623, opened 2021-02-03
Assorted cleanups

dnicolodi:
This has been obsoleted with the switch to Click

### DIFF
--- a/beancount/projects/export.py
+++ b/beancount/projects/export.py
@@ -1,32 +1,39 @@
 #!/usr/bin/env python3
-"""List assets and join with attributes of account and commodities to a CSV file.
+"""List assets and attributes of account and commodities.
 
-The purpose of this script is
+This script:
 
-- Produce a table of postings for the assets and liabilities
-- Produce a table of per-account attributes
-- Produce a table of per-commodity attributes
-- Join these tables
-- Output them to a CSV file.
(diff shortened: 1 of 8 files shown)


dnicolodi:
Uhm... I shluld have forgotten a commit somewhere: It seems that beancount/projects/export_tests.py still uses some of this.

dnicolodi:
Indeed, I forgot some commits in another branch. This grew a bit longer.

beancount/projects/export.py
```diff
@@ -1,32 +1,14 @@
 #!/usr/bin/env python3
-"""List assets and join with attributes of account and commodities to a CSV file.
```
blais, line 2: Oh that's a real bummer. Is there no way to provide the docstring at the top of the file?
I'd really like to keep it there. It's the first thing you see when you open the file.

blais (approved):
SGTM
Just the comment at the top, is it possible to do that?

beancount/projects/export.py
dnicolodi: Sure, it is possible. Fixing this I noticed that the help text gets re-wrapped wrong when output by Click.

dnicolodi:
I moved the docstring back to the top of the source file.

PR #625, opened 2021-02-05
setup: fix gcc error

simonla:
beancount/parser/tokens.c:174:5:error: 'for' loop initial declaration used outside C99 mode

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ def get_cflags():
         # See https://bitbucket.org/blais/beancount/issues/173/
         return ["-DYY_NO_UNISTD_H"]
     else:
-        return None
+        return ["-std=c99"]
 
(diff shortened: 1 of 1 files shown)


blais:
We could change the code instead to put the variable declaration above

dnicolodi:
@blais Do you know which compiler chokes on this? I guess it was working on your Ubuntu 16.04, thus it must be something older, and anything older than that does not come with Python 3.6, which we require.

blais:
No idea, but I don't think moving the declarations up in the function for v2 would hurt anything if it helps someon

dnicolodi:
I just would like to have a way to check if this is the only fix needed.

blais:
Good point let's have users send a patch

PR #643, opened 2021-04-02
use env marker in `install_require`

trim21:
This avoids some installer dependence resolve bug that can't detect `python-magic` isn't required on windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -120,17 +120,13 @@ install_requires = [
     # programmatically and to upload lists of holdings to a Google
     # Spreadsheet for live intra-day monitoring.
     'google-api-python-client',
+  
+    # This library is needed to identify the type of a file for
+    # import. It uses ctypes to wrap the libmagic library which is
+    # not generally available on Windows nor is easily installed,
+    # thus the conditional dependency.
+    'python-magic>=0.4.12; sys_platform != "win32"'
 ]
-
-if sys.platform != 'win32':
-    install_requires += [
-        # This library is needed to identify the type of a file for
-        # import. It uses ctypes to wrap the libmagic library which is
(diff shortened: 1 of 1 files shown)


trim21:
Any chance we get a pypi release for this PR in these days?

dnicolodi:
Uh? The code that was there before, while not the most elegant, was doing exactly the same. Which resolver is confused by a dependency that is not in `install_requires` at all?

trim21:
@dnicolodi poetry. I guess it take all if condition as true when parsing setup.py.

dnicolodi:
This is a huge bug in poetry then. Have you reported it?

trim21:
There are many similar bug reports in poetry issues. poetry author suggests using env marker instead of dynamic execution.

dnicolodi:
I haven't played with poetry much but I was really expecting something more from it. Unfortunately it seems that the python packaging ecosystem is evolving toward a disjoint set of tools that each cover 80% of the use cases. Apparently a flashy website is more important that correctness :(

trim21:
poetry is a very good packaging tool and dependency manager. It works fine in most cases.

I think most users choose it not because of its website but the function it provided.

blais:
Python packaging has been a confusing shitshow for decades now

blais:
Slim
Away from the console for s few days

dnicolodi:
That is exactly what I meant: it works in most cases, and, from what you wrote, the author does not consider making it work for the remaining cases worth the trouble. A dependency manager that requires changes to the packages that it should manage (and that are perfectly correctly under the specifications of the pre-existing tools) does not seem like a very good tool to me.

blais:
It's all a mess if you ask me.
Whatever works
Would rather invest time in building features

PR #652, opened 2021-05-15
Aggregate the state of DisplayContext across parsed files

### LINKED ISSUE #131: Aggregate the state of DisplayContext across all parsed files.
**[Original report](https://bitbucket.org/blais/beancount/issue/131) by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

See thread "[fava] Display budgets (#165)" in email.
DisplayContext is uninitialized if the top-level file is just a list of includes...
I think it's time to make at least some of the state aggregated across all parsed files.
DisplayContext should be it for sure.

upsuper:
This fixes #131.

### DIFF
--- a/TODO
+++ b/TODO
@@ -12,8 +12,6 @@ booking_full.py: # FIXME: Refactor compute_cost_number() and convert_costspec_to
 - Pull request for Unicode support: https://bitbucket.org/blais/beancount/pull-requests/13/allow-unicode-in-account-names/diff#comment-20233638
 - Fix #128
 - Fix #130
-- Fix #131
-- Make included files accumulate the state of their DisplayContext.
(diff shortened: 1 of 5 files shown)


blais:
Thanks @upsuper. I'd pretty much frozen v2 but this is too good to pass on. Will merge if tests pass. Thank you.

upsuper:
This causes an unfortunate new issue that it makes `dcontext` vulnerable to the price data that `bean-price` fetches... I suspect that `price` directives shouldn't be taken into `dcontext`'s stats, but it may not be easy to change given that the stats is currently collected when creating `Amount` instance.

PR #655, opened 2021-05-23
Reduce dependencies

dnicolodi:
Now that `upload-to-sheets` has been moved we can reduce the list of dependencies significantly. There is still an optional but somehow large and somehow obscure dependency on `lxml` that may need some more thought.


blais:
lxml is for scraping, and that was used with bean-bake.
Now that bean-web is gone, I should probably remove that as well.

Also google-api-python-client is still only required for the update_options.py script, which is something that could potentially be done by hand.

Issue #653, opened 2021-05-20
Different behavior when using "include"


hellige:
I'm exploring beancount, hoping to migrate a few different projects from ledger, some personal books as well as the bookkeeping system that I handle for a non-profit. There are a bunch of things I find very attractive about beancount, but there are a few sticking points and trade-offs, especially in the non-profit scenario. So, most likely I'll have other questions while I try to figure out whether I can make this transition successfully.

But right away I ran into something very strange. I use the include directive heavily, organizing things into sub-directories by year, transaction type, etc. But right away, I see several behavioral differences when I  move transactions into include files. The number of decimal places changes, as well as the output alignment of bean-report.

In bar.beancount:

```
    2021-01-01 open Assets:Cash
    2021-01-01 open Expenses:Blah

    2021-01-01 * "Test"
      Assets:Cash         -100.00 USD
      Expenses:Blah
```

And in foo.beancount:

`    include "bar.beancount"`

Then:

```
    $ bean-report bar.beancount bal
    Assets:Cash     -100.00 USD
    Equity
    Expenses:Blah    100.00 USD
    Income
    Liabilities
```

But:

```
    $ bean-report foo.beancount bal
    Assets:Cash   -100.00000000 USD
    Equity
    Expenses:Blah 100.00000000 USD
    Income
    Liabilities

```
This seems totally bizarre... Googling, I found a few discussions of this type of issue, but no firm conclusions about whether or not it's a bug. I found that I can fix the decimal inference by adding a bogus transaction to the root file (but this is not really a long-term solution for a collection of books, where the root file is mostly just global options and accounts can be opened/closed as needed, by date, in individual years). And even so, I still can't fix the alignment issue.

More generally, this is concerning to me, because it seems to completely violate the stream-processing date-ordered philosophy of beancount, which is one of the things that really appealed to me in the first place. In fact, the beancount docs themselves explictly say: "All directives are parsed and then basically stably sorted before any calculation or validation occurs. This also makes it trivial to implement inclusions of multiple files (you can just concatenate the files if you want)." But this is certainly not true!

So, is this the expected behavior, or is this a bug? Is there any way to add explicit options to the top-level file to guarantee that the file-location of transactions really makes absolutely no difference? It doesn't seem to happen in `bean-query`, so maybe I should just presume that that's the future...

dnicolodi:
> most likely I'll have other questions while I try to figure out whether I can make this transition successfully.

Please direct questions to the Beancount mailing list.

[...]

PR #657, opened 2021-05-24
parser: Implement an #include directive

dnicolodi:
The existing `include` directive has several limitations and its behavior is counterintuitive. Add an `#include` directive that works similarly as the C preprocessor homonym directive and hopefully less surprisingly. Other options could be `!include` or `.include` or simply rework the existing `include` directive to work this way. I vote for the latter option for v3. If we want to merge this into v2 maybe one of the other options is best for backward compatibility.

The cache invalidation part is not implemented yet, but this should give a flavor of the required parser changes.

Also the validation of the `#include` line needs to be improved somehow. Handling the include directly in the lexer makes it transparent to the other parts of the parser (which I think is a good thing) but makes the parsing of the include directive itself more cumbersome.

It currently does not support including encrypted files. Adding support for this is straightforward, but I have no idea of how much this feature is used and whether it makes sense to keep supporting it in v3.

See #653.


dnicolodi:
@blais WDYT?

mlell:
I was just stung by the absence of this, so I would have use for it. My use-case:

Main file: `current.beancount`

```
include accounts.beancount
include "2025.beancount"
include "204.beancount"
```

My `options` directives are in `assets.beancount`

I also have `full.beancount`, which includes `assets` and more years before 2024. I have `pad` directives at the start of each year's beancount file, so the balances are ok regardless of how many years I include. 

If you don't want/don't have time now to include this PR, please consider to amend the documentation: #990

blais:
@dnicolodi  do you still want this?

mlell:
That `include` in beancount does not work like `#include` in C does seem to confuse quite a few people (@dnicolodi has collected several issues [here](https://github.com/beancount/beancount/issues/653#issuecomment-845690548)). On the other hand, there could be value as well in file-local options, for example for `pushtag`. For example, consider a ledger for private activities and a ledger for a solo-business. They might be included in one big beancount ledger via `include` (for example, in Germany, there is one common tax number for private and mini-business and you can deduct private expenses like social security from your business revenue). It might be very convenient to be able to use `pushtag` only within one of the ledgers, for example for a business trip. Also, probably there will be different Documents folders. Then again, there are also use-cases for `pushtag` which transcend one file, e.g. if splitting by year, and using `pushtag` in end of December and `poptag` in beginning of January....

blais:
Why not use m4 to preprocess your includes to a temp file and run beancount on that?
Should be near instant and take less than 2-3 LOCs

PR #659, opened 2021-05-26
parser: Stop pretending we support encodings other than UTF-8

dnicolodi:
During the work on the include directive I had to think about which encoding to use for the included files. And I realized that our encoding support is mostly a joke. Ripping it out completely is easier and more honest.

The parser does not really understand any encoding other than UTF-8. Support for other encodings is limited to string literals, which makes sense only for a very limited subset of encodings.

The limited support for other encodings is not exposed to the command line tools and no one ever complained, so I assume that all Beancount users use systems with UTF-8 default encoding or arrange for their ledgers to be encoded in UTF-8 or in an encoding which is subset equivalent to UTF-8 like ascii or latin1.

I stopped at the innermost interface layer in removing the `encoding` argument to parser functions. It can be removed all the way and we can be happy supporting UTF-8 only (which is not a bad choice in 2021, IMHO) or we can support ether encodings via transparent recoding using, for example, `codecs.EncodedFile`.

### DIFF
--- a/beancount/loader_test.py
+++ b/beancount/loader_test.py
@@ -388,8 +388,8 @@ class TestLoadIncludesEncrypted(encryption_test.TestEncryptedBase):
 
         self.assertFalse(errors)
         self.assertEqual(2, len(entries))
-        self.assertRegex(entries[0].meta['filename'], 'apples.beancount')
-        self.assertRegex(entries[1].meta['filename'], 'oranges.+count.asc')
+        self.assertTrue(entries[0].meta['filename'].endswith('/apples.beancount'))
+        self.assertTrue(entries[1].meta['filename'].endswith('/oranges.beancount.asc'))
 
 
 class TestLoadCache(unittest.TestCase):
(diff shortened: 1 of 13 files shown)


blais:
I'm pretty sure it wasn't "a joke," but I don't see a problem only supporting UTF8.

dnicolodi:
Maybe the intention was to support other encodings but the current code tries to do it only for string literals, and even there I am not sure it works for most encodings. I think the lexer could be in principle be fixed, but as long as we use Flex it is very hard to do. For things like encodings I thing supporting only one and clearly state it is better than incomplete support for many.

RE/flex explicitly supports UTF-8 and other encodings via trans-coding to UTF-8. A similar mechanism could be implemented for the Flex lexer. However, the include mechanism support (either the old one or the new one in #660) would need a way to specify the encoding of the included files. I think the added complexity outweights the advantages. Given that the command line tools do not really allow for selecting and encoding and no one complained, I think requiring UTF-8 encoding is the easiest solution.

Unless someone sees any drawback in it, I'll remove the encoding parameter from the parser functions and I'll see if the documentation needs to be extended to clarify that the ledgers needs to be in UTF-8.

PR #662, opened 2021-06-02
Amount units cannot be None

lijunle:
Instead, we should use `NULL_AMOUNT` for it.

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -12,6 +12,7 @@ from decimal import Decimal
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Union
 
 from beancount.core.amount import Amount
+from beancount.core.amount import NULL_AMOUNT
 from beancount.core.number import D
 from beancount.core.position import Cost
 from beancount.core.position import CostSpec
@@ -483,7 +484,7 @@ def create_simple_posting(entry, account, number, currency):
     if isinstance(account, str):
         pass
     if number is None:
-        units = None
+        units = NULL_AMOUNT
     else:
         if not isinstance(number, Decimal):
             number = D(number)


beancount/core/data.py
```diff
@@ -483,7 +484,7 @@ def create_simple_posting(entry, account, number, currency):
     if isinstance(account, str):
         pass
     if number is None:
-        units = None
+        units = NULL_AMOUNT
```
dnicolodi, line 487: What is triggering this change? `NULL_AMOUNT` is not used anywhere in the codebase (I would not be surprised if it is a left over from some old incarnation of the code) and I am sure there are several places where the code explicitly checks for `None`.

lijunle: I cannot quite remember my case. Do a quick search in the GitHub code base like this, there are few places does not check `posting.units` is not `None` before access it.

> https://github.com/beancount/beancount/search?q=posting.units

For example, this one:

https://github.com/beancount/beancount/blob/e1716b492c7619682a6d7c33c4873aa41954af1e/beancount/parser/booking_method.py#L75

---

Btw, after I sent out this PR and had more coding with `Amount`. I found its inconvenience. As a literal type, it should has its own Amount.ZERO to be compatible with add/substrate/multiple/divide with any other Amount instances. However, due to the `.currency` check, the current `NULL_AMOUNT` is not much useful.

For example, I could like to add up all postings about `Expense:Restaurant`, here is *what I want*:

```py
restaurant_expense = Amount.ZERO
for p in entry.postings:
  if p.account == "Expense:Restaurant":
    restaurant_expense = Amount.add(restaurant_expense, p.units)
```

But without `Amount.ZERO`, I need to:

```py
restaurant_expense = None
for p in entry.postings:
  if p.account == "Expense:Restaurant":
    restaurant_expense = Amount.add(restaurant_expense, p.units) if restaurant_expense else p.units
```

It is kinds of annoying to do a simple add with `if-else`.

dnicolodi: I am not sure I follow. After booking there cannot be cases in which a postings with unspecified units. How do you end up with a ledger with transactions that have None in the units field?

lijunle: Good question! Here is the story, I have a couple of documents related to health transactions. I could like to put these documents into one single folder (e.g., under one account).

I was also using [link_document plug-in](https://github.com/beancount/fava/blob/main/src/fava/plugins/link_documents.py) to link the document with its corresponding transaction. It means, all these transactions need to have one ghost posting for the account of document. 

That is why I am writing code to append a zero amount posting to transactions (The method of this PR)


Later, I rewrite the login to link documents by myself, without the need about document’s account being a part of transaction. 

I don’t need the changes of this PR anymore although I could still suggest on the Amount.Zero situation.

dnicolodi: I still not understand what you are trying to do.

> Good question! Here is the story, I have a couple of documents related to health transactions. I could like to put these documents into one single folder (e.g., under one account).

This has nothing to do with postings or amounts, AFAICT.

> I am writing code to append a zero amount posting to transactions

This does not require you to look at the units in the existing posting either, AFAICT.

> I could still suggest on the Amount.Zero situation.

I don't understand what this `Amount.ZERO` thing is supposed to do.

The `amount.add` function (the `Amount` object does not have an `add` method you use in your example) allows you to add `Amount`s that have the same currency, thus in your code you must know in which currency the postings you are dealing with are, thus you can assign to the `restaurant_expense` variable in your example an explicit zero value in the right currency, for example `restaurant_expense = amount.Amount(decimal.Decimal('0.0'), 'EUR')`, or, shorter, `restaurant_expense = amount.A('0.0 EUR')`. What am I missing?

The idea of a special object that acts as a zero amount in all currencies is interesting and I see how it can make some operations easier, but it would require adapting all code paths that deal with `Amount`s to guard from this special object propagating in places where it cannot be dealt with. I don't think the advantages are such to justify this.

lijunle: Never mind. Let me focus on the Amount.Zero issue. What is the simple way to sum up *parts of* postings in one transaction? The postings could be any currency but same.

dnicolodi: Something like:
```python
amounts = [p.units for p in entry.postings]
total = functools.reduce(amount.add, amounts[1:], amounts[0])
```
or
```python
total = functools.reduce(amount.add, (p.units for p in entry.postings[1:]), entry.postings[0].units)
```
if you like one-liners.

blais:
Sorry guys I'm unable to participate in this right now, I have other urgent things to deal with. I'll get around to it eventually.

beancount/core/data.py
lijunle: That looks good! I am OK to close this PR if `NULL_AMOUNT` is legacy and not suggested to use.

dnicolodi: Also
```python
amounts = [p.units for p in entry.postings]
total = amount.Amount(sum(a.number for a in amounts), amounts[0].currency)
```
if you don't need to verify that all amounts are in the same currency.

blais: > What is triggering this change? `NULL_AMOUNT` is not used anywhere in the codebase (I would not be surprised if it is a left over from some old incarnation of the code) and I am sure there are several places where the code explicitly checks for `None`.

I agree with you Daniele, NULL_AMOUNT was probably a mistake. It's was really only used in the Position's sortkey a long time ago and its use has since been removed.
I'll remove it now (in v3).

PR #664, opened 2021-06-30
Fix meta handling of check_closing plugin

upsuper:
I believe the intention of the replace is to remove the `closing` from the posting, rather than overriding the meta of `entry` with the meta from `posting`. The current overriding just doesn't make any sense to me. It also causes issues like beancount/fava#1301.

### DIFF
--- a/beancount/plugins/check_closing.py
+++ b/beancount/plugins/check_closing.py
@@ -55,12 +55,13 @@ def check_closing(entries, options_map):
     new_entries = []
     for entry in entries:
         if isinstance(entry, data.Transaction):
-            for posting in entry.postings:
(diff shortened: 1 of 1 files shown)


blais:
SGTM, good find, but how come the test was working?
Can you change the test so it fails without your change, and then change the test so it passes?

upsuper:
I tried to change the test, yes, but the problem is that the test ignores the meta when comparing entries, so it's probably non-trivial to make the test check this.

The comparison happens here: https://github.com/beancount/beancount/blob/01fcd5d497f3f72b87d7e15c7975d5cbdd931bbb/beancount/core/compare.py#L133-L134

Issue #667, opened 2021-07-10
bean-price failed to fetch prices from yahoo


SEIAROTg:
Today my usual `bean-price commodities.bean -i` run failed with:
```
ERROR   : Error fetching GOOG: Expecting value: line 1 column 1 (char 0)
ERROR   : Could not fetch for job: DatedPrice(base='GOOG', quote='USD', date=None, sources=[PriceSource(module=<module 'beancount.prices.sources.yahoo' from '/usr/local/lib/python3.9/site-packages/beancount/prices/sources/yahoo.py'>, symbol='GOOG', invert=False)])
```

I did some print-debugging in `yahoo.py` and found it returns 403 with contents `Forbidden` which failed to be parsed as json. Some curl attempts show that yahoo no longer likes a user agent that contains `python-requests\S`:
```
$ curl -H "User-Agent: python-requests/2.25.1" -H "Accept-Encoding: gzip, deflate" -H "Accept: */*" -H "Connection: keep-alive" 'https://query1.finance.yahoo.com/v7/finance/quote?symbols=GOOG&fields=symbol%2CregularMarketPrice%2CregularMarketTime&exchange=NYSE&lang=en-US&corsDomain=finance.yahoo.com&.tsrc=finance'
Forbidden

$ curl -H "User-Agent: python-requests" -H "Accept-Encoding: gzip, deflate" -H "Accept: */*" -H "Connection: keep-alive" 'https://query1.finance.yahoo.com/v7/finance/quote?symbols=GOOG&fields=symbol%2CregularMarketPrice%2CregularMarketTime&exchange=NYSE&lang=en-US&corsDomain=finance.yahoo.com&.tsrc=finance'
{"quoteResponse":{"result":[{"language":"en-US","region":"US","quoteType":"EQUITY","quoteSourceName":"Delayed Quote","triggerable":true,"sourceInterval":15,"exchangeDataDelayedBy":0,"tradeable":false,"marketState":"POSTPOST","firstTradeDateMilliseconds":1092922200000,"priceHint":2,"regularMarketTime":1625860802,"regularMarketPrice":2591.49,"fullExchangeName":"NasdaqGS","exchange":"NMS","exchangeTimezoneName":"America/New_York","exchangeTimezoneShortName":"EDT","gmtOffSetMilliseconds":-14400000,"market":"us_market","esgPopulated":false,"symbol":"GOOG"}],"error":null}}
```

It's also reproduced with https://reqbin.com/curl so it's unlikely just me.

[...]

PR #668, opened 2021-07-10
Remove User-Agent sent to yahoo finance api.

SEIAROTg:
The default "python-requests/$VERSION" is causing 403.

https://github.com/beancount/beancount/issues/667


tonoriesco:
Thank you SEIAROTg, works perfectly for me.

PR #670, opened 2021-07-15
Test files are deleted before they can be opened.

michelgb:
In Windows test files are deleted before they can be used by the test code.

See https://stackoverflow.com/questions/6416782/what-is-namedtemporaryfile-useful-for-on-windows

### DIFF
--- a/beancount/utils/test_utils.py
+++ b/beancount/utils/test_utils.py
@@ -186,7 +186,7 @@ def docfile(function, **kwargs):
         allowed = ('buffering', 'encoding', 'newline', 'dir', 'prefix', 'suffix')
         if any(key not in allowed for key in kwargs):
             raise ValueError("Invalid kwarg to docfile_extra")
-        with tempfile.NamedTemporaryFile('w', **kwargs) as file:
+        with tempfile.NamedTemporaryFile('w', delete=False, **kwargs) as file:
             text = contents or function.__doc__
             file.write(textwrap.dedent(text))
             file.flush()


beancount/utils/test_utils.py
```diff
@@ -186,7 +186,7 @@ def new_function(self):
         allowed = ('buffering', 'encoding', 'newline', 'dir', 'prefix', 'suffix')
         if any(key not in allowed for key in kwargs):
             raise ValueError("Invalid kwarg to docfile_extra")
-        with tempfile.NamedTemporaryFile('w', **kwargs) as file:
+        with tempfile.NamedTemporaryFile('w', delete=False, **kwargs) as file:
```
dnicolodi, line 189: Who deletes the temporary file then?

dnicolodi (changes requested):
Making the test suite correctly run on Windows requires many more and more invasive changes. I don't think making a few more test pass at the cost of leaving temporary files around is particularly useful.

beancount/utils/test_utils.py
blais: How is that possible that the files get deleted first?
Is it triggered by my call to flush()?

I agree, not deleting the files is horrible.

michelgb: @dnicolodi - I was curious and ran all the tests in Windows and surprisingly I got only 33 failures out of 1300+ tests. Perhaps with a little effort we could support it. However I think there are other features that I need more (like AVERAGE booking) so probably it is just easier to install a Linux VM.

```
Ran 1355 tests in 36.826s

FAILED (failures=33, errors=29, skipped=22, expected failures=2)
Test failed: <unittest.runner.TextTestResult run=1355 errors=29 failures=33>
error: Test failed: <unittest.runner.TextTestResult run=1355 errors=29 failures=33>
```

@blais - I honestly don't know why Windows immediately deletes the file after the `with` block. It might be a garbage collection nuance since you pass the name of the file and not the file handler itself. It is dumb because this behavior defeats the very purpose of `NamedTemporaryFile`.

That being said we could drop this PR. 

Thanks for beancount :-)

dnicolodi: I looked into this a while ago: IIRC, the issue is not that the file is deleted, but that on Windows the file cannot be opened again while it is already opened for writing. The `delete=False` simply triggers a code path where the last file descriptor to the temporary file is dropped before the subsequent open.

I count 62 problematic tests (33 failures and 29 errors that do not happen when running the tests on a Unix platform). I think having the test suite work on Windows is a worthwhile goal, however, the proposed change is an ugly work-around. The issue needs to be fixed properly, ie without leaving temporary files around.

blais: Can't we just insert a random suffix to avoid collisions?

dnicolodi: Unfortunately, the issue is not of name collisions but of stupid Windows behavior regarding file handlers.

michelgb:
I don't really have any other solution at the moment. Even with this patch there are 62 tests we need to investigate. So I think I will close this and use the time on other things. Thank you.

PR #674, opened 2021-08-02
Fix docs for ABS() in bean-query

### LINKED ISSUE #673: Copy/paste error in documentation for ABS bean-query function
```
$ bean-query /dev/null
Input file: "Beancount"
Ready with 0 directives (0 postings in 0 transactions).
beancount> help targets

Simple functions:                                                                                                                                                                                           
  'ABS(Decimal)': Compute the length of the argument. This works on sequences.                                                                                                                              
  'ABS(Inventory)': Compute the length of the argument. This works on sequences.                                                                                                                            
  'ABS(Position)': Compute the length of the argument. This works on sequences.                                                                                                                             
```

Looks like this has been copied inadvertently from

```
  'LENGTH(list|set|str)': Compute the length of the argument. This works on                                                                                                                                 
    sequences.                                                                                                                                                                                              
```

I'm guessing these are absolute value rather than Length?

mhansen:
Should fix #673

### DIFF
--- a/beancount/query/query_env.py
+++ b/beancount/query/query_env.py
@@ -57,7 +57,7 @@ class NegInventory(_Neg):
 
 
 class AbsDecimal(query_compile.EvalFunction):
-    "Compute the length of the argument. This works on sequences."
+    "Absolute value."
     __intypes__ = [Decimal]
 
     def __init__(self, operands):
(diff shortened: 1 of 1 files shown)


blais:
Ideally to be replicated on https://github.com/beancount/beanquery

PR #681, opened 2021-10-17
tests: Fix random test failures in core/number_test.py

dnicolodi:
Some of the quantization tests assume that all numbers in the test
corpus have a non zero integer part. Fix the generation of the random
numbers composing the test corpus to always be larger than one.


blais:
Good catch!

PR #685, opened 2021-11-10
Fix transaction alignment in parser (ValueError: '=' alignment not allowed in string format specifier)

ArthurFDLR:
I encounter the following error using the ```context``` command of bean-doctor:
```
[...]
  File "[...]\venv\lib\site-packages\beancount\parser\printer.py", line 210, in Transaction
    oss.write(fmt(account,
ValueError: '=' alignment not allowed in string format specifier
```
I did not manage to locate the precise source of the error but adding [left justifier specifiers](https://docs.python.org/3/library/string.html#formatspec) ```<``` to the formatter string seems to solve the issue.

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -197,7 +197,7 @@ class EntryPrinter:
                                if self.render_weight and width_weight > 0
                                else False)
         if non_trivial_balance:
-            fmt = "{0}{{:{1}}}  {{:{2}}}  ; {{:{3}}}\n".format(
+            fmt = "{0}{{:<{1}}}  {{:<{2}}}  ; {{:<{3}}}\n".format(
                 self.prefix, width_account, width_position, width_weight).format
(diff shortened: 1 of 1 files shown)


blais:
Cargo cult... Can you find out the source of the error?
Print out the values of the arguments.
Won't accept the change without understanding what's making it break really well.

reedlaw:
In [this block](https://github.com/beancount/beancount/blob/c70b38eec8a5e74ad9f3a37fa5fdf626f4e41931/beancount/parser/printer.py#L199-L210) the `non_trivial_balance` check is redundant on [line 208](https://github.com/beancount/beancount/blob/c70b38eec8a5e74ad9f3a37fa5fdf626f4e41931/beancount/parser/printer.py#L208). Instead, the guard should be against a missing `weight_str`.

blais:
Seems right @reedlaw

blais:
BTW this code rewritten in v3

PR #687, opened 2021-11-25
Make parse_file retain original file order

### LINKED ISSUE #649: Is there a reason why parse.parse_file() sorts entries by date?
The sorting gets in the way of scripts that try to programmatically update the ledger. It is not a big deal because the original order can be obtained sorting again the entries by line number `entries.sort(key=lambda x: x.meta['lineno'])` but it is a bit surprising that the parsed entries are returned in an order different than the file order.

josephw:
The discussion in #649 suggests that there's no specific reason for sorting directly after parsing. Looking at the code, it appears to be a relatively surface-level decision to sort the results. This PR adds a small test for ordering, and removes the sorting from `get_entries`.

The original discussion [says](https://github.com/beancount/beancount/issues/649#issuecomment-825893407):
> A deep code inspection would be required through.

While this change doesn't cause any tests to fail, I'm unclear on whether there are other subtleties here that I've missed.

resolves #649

### DIFF
--- a/beancount/parser/grammar.py
+++ b/beancount/parser/grammar.py
@@ -36,7 +36,6 @@ from beancount.core.data import EMPTY_SET
 from beancount.parser import lexer
 from beancount.parser import options
 from beancount.core import account
-from beancount.core import data
 
 
 ParserError = collections.namedtuple('ParserError', 'source message entry')
@@ -176,9 +175,9 @@ class Builder(lexer.LexBuilder):
         """Return the accumulated entries.
 
         Returns:
-          A list of sorted directives.
+          A list of directives.
         """
-        return sorted(self.entries, key=data.entry_sortkey)
+        return list(self.entries)
 
     def get_options(self):
         """Return the final options map.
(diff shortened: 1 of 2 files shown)


dnicolodi:
This would be an API change. If it is desired to have the entries sorted as in the ledger file, is easy to sort them by the line number saved in the metadata. Thus I don't think it is worth the API breakage in v2. In v3 the parser is being anyhow rewritten in C++, thus the change to the Python code is not particularly meaningful. Also, almost all tests are on ledger snippets where the entries are sorted by date, thus the fact that the tests still pass is not a good indication that there aren't palces whete the entries are expected to be sorted and that would break if this change is introduced.

josephw:
Thanks for the context.

> thus the change to the Python code is not particularly meaningful.

You're right; and the `grammar.py` change here is not significant. My intention here was to turn that issue into a part of v3's unit test suite, with the assumption that the C++ rewrite would be tested against the existing unit tests.

If this is an acceptable API change for v3, updating the existing Python version seems like a reasonable way to record that. If it's _not_, #649 could probably be closed now as WONTFIX.

PR #688, opened 2021-12-15
Enable addition and subtraction on two Amounts.

IguanaBen:
I would like to be able to do addition and subtraction of Amounts in the bql language. This code lets python know how to add and subtract Amounts using add/sub functions that already exist but weren't linked to the class. EvalAdd and EvalSub no longer force the output to be Decimal as they can now be Amounts (or maybe Positions as well in the future).  To test, I added a EvalAmount to create an Amount similar to how EvalConstant is defined.

### DIFF
--- a/beancount/core/amount.py
+++ b/beancount/core/amount.py
@@ -42,6 +42,20 @@ class Amount(_Amount):
     valid_types_number = (Decimal, type, type(None))
     valid_types_currency = (str, type, type(None))
 
+    def __add__(self, other):
+        return add(self, other)
+
+    def __sub__(self, other):
+        return sub(self, other)
+
+    def __mul__(self, other):
+        return mul(self, other)
+
+    __rmul__ = __mul__
+
+    def __truediv__(self, other):
+        return div(self, other)
+
     def __new__(cls, number, currency):
         """Constructor from a number and currency.
 
@@ -163,8 +177,8 @@ def mul(amount, number):
     """
     assert isinstance(amount.number, Decimal), (
         "Amount's number is not a Decimal instance: {}".format(amount.number))
-    assert isinstance(number, Decimal), (
-        "Number is not a Decimal instance: {}".format(number))
+    assert isinstance(number,(int, Decimal)), (
+        "Number is not an int or Decimal instance: {}".format(number))
     return Amount(amount.number * number, amount.currency)
 
 def div(amount, number):
@@ -178,8 +192,8 @@ def div(amount, number):
     """
     assert isinstance(amount.number, Decimal), (
         "Amount's number is not a Decimal instance: {}".format(amount.number))
-    assert isinstance(number, Decimal), (
-        "Number is not a Decimal instance: {}".format(number))
+    assert isinstance(number, (int, Decimal)), (
+        "Number is not an int or Decimal instance: {}".format(number))
     return Amount(amount.number / number, amount.currency)
 
 def add(amount1, amount2):
--- a/beancount/query/query_compile.py
+++ b/beancount/query/query_compile.py
@@ -15,6 +15,7 @@ import operator
 from decimal import Decimal
 
 from beancount.core import inventory
+from beancount.core.amount import Amount
 from beancount.query import query_parser
 
 
@@ -192,26 +193,39 @@ class EvalContains(EvalBinaryOp):
 class EvalMul(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x * y)
-        super().__init__(f, left, right, Decimal)
+        if left.dtype == Amount and right.dtype in [int, Decimal]:
+            super().__init__(operator.mul, left, right, Amount)
+        elif left.dtype in [int, Decimal] and right.dtype == Amount:
+            super().__init__(operator.mul, left, right, Amount)
+        elif left.dtype == int and right.dtype == int:
+            super().__init__(operator.mul, left, right, int)
+        elif left.dtype in [int, Decimal] and right.dtype in [int, Decimal]:
+            super().__init__(operator.mul, left, right, Decimal)
+        else:
+            raise TypeError("Multiply of {} with {} not supported".format(
+                left.dtype,right.dtype))
 
 class EvalDiv(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x / y)
-        super().__init__(f, left, right, Decimal)
+        if left.dtype == Amount and right.dtype in [int, Decimal]:
+            super().__init__(operator.truediv, left, right, Amount)
+        elif left.dtype in [int, Decimal] and right.dtype in [int, Decimal]:
+            f = lambda x, y: Decimal(x) / Decimal(y)
+            super().__init__(f, left, right, Decimal)
+        else:
+            raise TypeError("Division of {} with {} not supported".format(
+                left.dtype,right.dtype))
 
 class EvalAdd(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x + y)
-        super().__init__(f, left, right, Decimal)
+        super().__init__(operator.add, left, right, right.dtype)
 
 class EvalSub(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x - y)
-        super().__init__(f, left, right, Decimal)
+        super().__init__(operator.sub, left, right, right.dtype)
 
 
 # Interpreter nodes.
--- a/beancount/query/query_compile_test.py
+++ b/beancount/query/query_compile_test.py
@@ -6,10 +6,12 @@ import unittest
 from decimal import Decimal
 
 from beancount.core.number import D
+from beancount.core.amount import A
+
 from beancount.query import query_parser as qp
 from beancount.query import query_compile as qc
 from beancount.query import query_env as qe
-
+from beancount.core import amount
 
 class TestCompileExpression(unittest.TestCase):
 
@@ -171,20 +173,39 @@ class TestCompileDataTypes(unittest.TestCase):
         self.assertEqual(bool, c_or.dtype)
 
     def test_compile_EvalMul(self):
-        c_plus = qc.EvalMul(qc.EvalConstant(17), qc.EvalConstant(18))
+        c_plus = qc.EvalMul(qc.EvalConstant(D(17)), qc.EvalConstant(D(18)))
         self.assertEqual(Decimal, c_plus.dtype)
 
     def test_compile_EvalDiv(self):
-        c_plus = qc.EvalDiv(qc.EvalConstant(17), qc.EvalConstant(18))
+        c_plus = qc.EvalDiv(qc.EvalConstant(D(17)), qc.EvalConstant(D(18)))
         self.assertEqual(Decimal, c_plus.dtype)
 
     def test_compile_EvalAdd(self):
-        c_plus = qc.EvalAdd(qc.EvalConstant(17), qc.EvalConstant(18))
+        c_plus = qc.EvalAdd(qc.EvalConstant(D(17)), qc.EvalConstant(D(18)))
         self.assertEqual(Decimal, c_plus.dtype)
 
     def test_compile_EvalSub(self):
-        c_plus = qc.EvalSub(qc.EvalConstant(17), qc.EvalConstant(18))
+        c_plus = qc.EvalSub(qc.EvalConstant(D(17)), qc.EvalConstant(D(18)))
         self.assertEqual(Decimal, c_plus.dtype)
+
+    def test_compile_EvalAddAmount(self):
+        c_plus = qc.EvalAdd(qc.EvalConstant(A('1.11 USD')), qc.EvalConstant(A('3.55 USD')))
+        self.assertEqual(amount.Amount, c_plus.dtype)
+
+    def test_compile_EvalSubAmount(self):
+        c_plus = qc.EvalSub(qc.EvalConstant(A('1.11 USD')), qc.EvalConstant(A('3.55 USD')))
+        self.assertEqual(amount.Amount, c_plus.dtype)
+
+    def test_compile_EvalMultAmount(self):
+        c_plus = qc.EvalMul(qc.EvalConstant(A('1.11 USD')), qc.EvalConstant(D('2.31')))
+        self.assertEqual(amount.Amount, c_plus.dtype)
+        c_plus = qc.EvalMul(qc.EvalConstant(D('2.31')), qc.EvalConstant(A('1.11 USD')))
+        self.assertEqual(amount.Amount, c_plus.dtype)
+
+    def test_compile_EvalDivAmount(self):
+        c_plus = qc.EvalDiv(qc.EvalConstant(A('1.11 USD')), qc.EvalConstant(D('2.31')))
+        self.assertEqual(amount.Amount, c_plus.dtype)
+
 
 
 class TestCompileMisc(unittest.TestCase):


beancount/query/query_compile_test.py
```diff
@@ -186,6 +186,14 @@ def test_compile_EvalSub(self):
         c_plus = qc.EvalSub(qc.EvalConstant(17), qc.EvalConstant(18))
         self.assertEqual(Decimal, c_plus.dtype)
 
+    def test_compile_EvalAddAmount(self):
+        c_plus = qc.EvalAdd(qc.EvalAmount(Decimal('1.11','USD')), qc.EvalAmount(Decimal('3.55','USD')))
```
dnicolodi: This cannot work, the second arguments to `Decimal` are wrong.

beancount/query/query_compile_test.py
```diff
@@ -186,6 +186,14 @@ def test_compile_EvalSub(self):
...
+        self.assertEqual(Decimal, c_plus.dtype)
+
+    def test_compile_EvalSubAmount(self):
+        c_plus = qc.EvalSub(qc.EvalAmount(Decimal('1.11','USD')), qc.EvalAmount(Decimal('3.55','USD')))
```
dnicolodi: Here too.

beancount/query/query_compile.py
```diff
@@ -204,13 +215,13 @@ def __init__(self, left, right):
 class EvalAdd(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x + y)
+        f = lambda x, y: (x + y)
```
dnicolodi: Redundant parentheses.

beancount/query/query_compile.py
```diff
@@ -204,13 +215,13 @@ def __init__(self, left, right):
...
         super().__init__(f, left, right, Decimal)
 
 class EvalSub(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x - y)
+        f = lambda x, y: (x - y)
```
dnicolodi: Here too.

beancount/query/query_compile.py
```diff
@@ -90,6 +91,16 @@ def __init__(self, value):
...
+class EvalAmount(EvalNode):
+    __slots__ = ('decimal','currency')
+
+    def __init__(self, decimal,currency):
+        super().__init__(type(amount.Amount))
+        self.decimal = decimal
+        self.currency = currency
+
+    def __call__(self, _):
+        return amount.Amount(self.decimal,self.currency)
```
dnicolodi: Missing space after `,`. Also, wouldn't it be simpler to have a single `value` class member, construct the `Amount` instance in the `__init__` method and assign it to `self.value` and just return it here?

beancount/query/query_compile.py
```diff
@@ -90,6 +91,16 @@ def __init__(self, value):
     def __call__(self, _):
         return self.value
 
+class EvalAmount(EvalNode):
```
dnicolodi: Is this class used only in the unit tests? If so, why can't you simply use `EvalConstant` and pass an `amount.Amount` instance to its constructor?

beancount/core/amount.py
```diff
@@ -42,6 +42,13 @@ class Amount(_Amount):
     valid_types_number = (Decimal, type, type(None))
     valid_types_currency = (str, type, type(None))
 
+    def __add__(self,other):
+        return add(self,other)
```
dnicolodi: Missing space after the commas.

beancount/core/amount.py
```diff
@@ -42,6 +42,13 @@ class Amount(_Amount):
...
+
+    def __sub__(self,other):
+        return sub(self,other)
+
+
```
dnicolodi, line 58: Extra newline.

beancount/query/query_compile_test.py
```diff
@@ -186,6 +186,14 @@ def test_compile_EvalSub(self):
         c_plus = qc.EvalSub(qc.EvalConstant(17), qc.EvalConstant(18))
         self.assertEqual(Decimal, c_plus.dtype)
 
+    def test_compile_EvalAddAmount(self):
+        c_plus = qc.EvalAdd(qc.EvalAmount(Decimal('1.11','USD')), qc.EvalAmount(Decimal('3.55','USD')))
+        self.assertEqual(Decimal, c_plus.dtype)
```
dnicolodi: Uh!? You testing that the result is of type `Deciamal`. Shouldn't it be of type `Amount`?

beancount/query/query_compile.py
```diff
@@ -90,6 +91,16 @@ def __init__(self, value):
...
+    __slots__ = ('decimal','currency')
+
+    def __init__(self, decimal,currency):
+        super().__init__(type(amount.Amount))
```
dnicolodi: `type(amount.Amount)` seems wrong here, it should simply be `amount.Amount`.

dnicolodi (commented):
The changes go in the right direction, but there are a few issues. Also, please make sure that the unit test pass.

dnicolodi:
@blais, we need to decide if we want `bean-query` continue to live here or in the dedicated repository.

beancount/query/query_compile.py
```diff
@@ -204,13 +215,13 @@ def __init__(self, left, right):
 class EvalAdd(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x + y)
+        f = lambda x, y: (x + y)
         super().__init__(f, left, right, Decimal)
```
dnicolodi: With the above change this could simply be `super().__init__(operator.add, left, right, Decimal)` which can also be slightly faster.

beancount/query/query_compile_test.py
IguanaBen: Sorry, I ran the unit test but didn't realize I can it on my installed version of beancounter (2,3.4) and not the source I was editing. My latest commit fixes all the errors I introduced in the unit test.

beancount/query/query_compile.py
IguanaBen: Personally, I find this easier to read, but happy to change if you like.

beancount/query/query_compile.py
IguanaBen: see above

beancount/query/query_compile.py
IguanaBen: code is removed.

beancount/query/query_compile.py
IguanaBen: I can do and now do. Thanks.

beancount/core/amount.py
IguanaBen: fixed

beancount/query/query_compile.py
IguanaBen: removed in latest.

beancount/query/query_compile.py
IguanaBen: Good point. I made that change in my latest.

beancount/query/query_compile.py
IguanaBen: Actually, moot now since I removed the lambda function and replaced with operator.add as you suggested.

IguanaBen:
With these newer commits, multiplication and division with numbers/decimals is also supported, provided the currency matches.
For example:

select 4*value(position)/2.12  + cost(position) + cost(position)/3

beancount/core/amount.py
```diff
@@ -48,6 +48,22 @@ def __add__(self,other):
     def __sub__(self,other):
         return sub(self,other)
 
+    def __mul__(self,other):
+        try:
+            other = Decimal(other)
+        except ValueError:
+            pass
```
dnicolodi: Why the attempted implicit conversion? The right way to implement this is:
```python
def __mul__(self, other):
    if not isinstance(other, Decimal):
        return NotImplemented
    return mul(self, other)
```
and similarly for `__truediv__`.

Probably `__add__` and `__sub__` should be in the form:
```python
def __add__(self, other):
    if not isinstance(other, Amount):
        return NotImplemented
    return add(self, other)
```

And at this point most of the asserts in the `add()` and `mul()` functions is redundant and maybe open coding them in the method implementations could be worth for performance.

beancount/core/amount.py
```diff
@@ -48,6 +48,22 @@ def __add__(self,other):
...
+        return mul(self,other)
+
+    def __rmul__(self,other):
+        return self.__mul__(other)
```
dnicolodi: Just `__rmul__ = __mul__` is enough.

beancount/query/query_compile.py
```diff
@@ -204,23 +193,21 @@ class EvalMul(EvalBinaryOp):
...
     def __init__(self, left, right):
         f = lambda x, y: (x * y)
+        resultType = right.dtype
         if left.dtype != right.dtype:
             if left.dtype == amount.Amount:
-               dtype = left.dtype
-            else:
-               dtype = right.dtype
-        super().__init__(f, left, right, dtype)
+               resultType = left.dtype
```
dnicolodi: Multiplying a `Decimal` by an `Amount` is permitted, but the computation of the return type here would get it wrong. Also, we use snake case and not camel case for variables.

beancount/query/query_compile_test.py
```diff
@@ -171,28 +171,28 @@ def test_compile_EvalOr(self):
...
 
     def test_compile_EvalSub(self):
-        c_plus = qc.EvalSub(qc.EvalConstant(17), qc.EvalConstant(18))
+        c_plus = qc.EvalSub(qc.EvalConstant(D(17)), qc.EvalConstant(D(18)))
         self.assertEqual(Decimal, c_plus.dtype)
 
     def test_compile_EvalAddAmount(self):
-        c_plus = qc.EvalAdd(qc.EvalAmount(Decimal('1.11','USD')), qc.EvalAmount(Decimal('3.55','USD')))
-        self.assertEqual(Decimal, c_plus.dtype)
+        c_plus = qc.EvalAdd(qc.EvalConstant(amount.Amount(Decimal('1.11'),'USD')), qc.EvalConstant(amount.Amount(Decimal('3.55'),'USD')))
```
dnicolodi: `D('1.11')` instead of `Decimal('1.11')` for consistency with the other tests. Or you could also add a `from beancount.core.amount import A` and simply write `A('1.11 USD')` for brevity.

beancount/query/query_compile_test.py
```diff
@@ -171,28 +171,28 @@ def test_compile_EvalOr(self):
...
+        self.assertEqual(amount.Amount, c_plus.dtype)
 
     def test_compile_EvalSubAmount(self):
-        c_plus = qc.EvalSub(qc.EvalAmount(Decimal('1.11','USD')), qc.EvalAmount(Decimal('3.55','USD')))
-        self.assertEqual(Decimal, c_plus.dtype)
+        c_plus = qc.EvalSub(qc.EvalConstant(amount.Amount(Decimal('1.11'),'USD')), qc.EvalConstant(amount.Amount(Decimal('3.55'),'USD')))
+        self.assertEqual(amount.Amount, c_plus.dtype)
```
dnicolodi, line 204: Tests for multiplication and division are missing.

dnicolodi (commented):
Getting closer. Please squash the comments.

beancount/core/amount.py
IguanaBen: Actually, without __rmul__ you can do "Amount * 4" but not "4 * Amount" .

beancount/core/amount.py
IguanaBen: The reason for the converting to Decimal is to support Numbers /  Ints (and I guess floats come from the ride). If I run:

select 4*value(position)

It will error without the conversion because 'Number is not a Decimal'. If you prefer, I could explicitly check for type Number or Decimal (and convert Numbers), but currently the behavior is to try to make the input a Decimal and if that fails it'll pass the original type which will then return a helpful error. If you try to multiply a Position by an Amount, you'll get:

 unsupported operand type(s) for /: 'Position' and 'Amount'

which seems like reasonable behavior. Another option would be to edit the add function itself to support more than just Decimal inputs, but since that function is (potentially) used elsewhere, I thought it best not to touch it. Let me know what you think is best.

IguanaBen:
To clarify what you said about squashing the comments (did you mean commits?). Should I squash all these commits into 1? If so, should I create a new pull request with that commit or force push that on this pull request?

Thanks,

beancount/core/amount.py
IguanaBen: done

beancount/query/query_compile.py
IguanaBen: I'm not seeing the issue. If Left=Decimal, and Right=Amount, then the default right.dtype (Amount) will be amount and this will work. If Left=Amount and Right=Decimal, then then left.dtype == amount.Amount will be true and resultType will be left.dtype which is Amount. Am I missing something.

I'll switch to snake_case on the next commit.

beancount/core/amount.py
dnicolodi: What I was trying to say is that you can replace the function definition with the assignment.

beancount/core/amount.py
dnicolodi: AFAIK, the BQL parser only produces `Decimal` instances. If this is not the case, the conversion should be done somewhere in bean query, not here.

dnicolodi:
I meant squash the commits. Force pushing to this PR is enough.

beancount/core/amount.py
dnicolodi: Correction, the BQL parser also emits `int`s. The check should be `isinstance(other, (Decimal, int))` and the implementation should support that: multiplying a `Decimal` by an `int` is supported and well defined.

beancount/query/query_compile_test.py
IguanaBen: Yeah, this was a mess with EvalAmount converting to an amount, but this is cleaned up now.

beancount/core/amount.py
IguanaBen: Sounds good. Just to clarify, do you think its better in the __add__ function to check the instance and if it is an int, convert to a Decimal? Or check just check and pass Decimal or Int to the add function and then edit the add function to work with type int in addition to Decimal?

dnicolodi: The fact that the parser emits both `Deciamal` and `int` objects has repercussions on the definition of the return types of the arithmetic operations below and on the definition of the operations. With your patches, `select 1/3` is a legal operation but the reported return type is `int` while it is `float`. It seems that correctly handling arithmetic functions requires some more thinking about data types in BQL.

beancount/core/amount.py
IguanaBen: Ok. Done.

beancount/query/query_compile_test.py
IguanaBen: Added

beancount/query/query_compile.py
dnicolodi: In this context `Decimal` and `Amount` are not the only two possible dtypes.

beancount/core/amount.py
IguanaBen: That's a good point. My latest (squashed) commit tries to address this. Instead of handling it in the __mul__ function, I put it in the EvalMul function to check the types and respond accordingly, which I think might be a little cleaner.

beancount/query/query_compile_test.py
IguanaBen: Done

beancount/query/query_compile.py
IguanaBen: I think this is addressed in my latest version.

beancount/query/query_compile.py
```diff
@@ -192,26 +192,41 @@ def __call__(self, context):
 class EvalMul(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x * y)
-        super().__init__(f, left, right, Decimal)
+        if (left.dtype == amount.Amount) and (right.dtype in [int,float, Decimal]):
```
dnicolodi: Where is `float` coming from?

beancount/query/query_compile.py
```diff
@@ -192,26 +192,41 @@ def __call__(self, context):
...
+            result_type = amount.Amount
+            super().__init__(operator.mul, left, right, result_type)
+        elif right.dtype == amount.Amount and (left.dtype in [int,float, Decimal]):
+            result_type = amount.Amount
+            super().__init__(operator.mul, left, right, result_type)
+        elif (left.dtype in [int,Decimal]) and (right.dtype in [int,Decimal]):
+            f = lambda x, y: Decimal(x * y)
```
dnicolodi: Why do you force the multiplication between `int` to a `Decimal`?

beancount/query/query_compile.py
```diff
@@ -192,26 +192,41 @@ def __call__(self, context):
...
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x / y)
-        super().__init__(f, left, right, Decimal)
+        if (left.dtype == amount.Amount) and (right.dtype in [int,float, Decimal]):
+            result_type = amount.Amount
+            super().__init__(operator.truediv, left, right, result_type)
+        elif right.dtype in [int,float,Decimal] and (left.dtype in [int,float, Decimal]):
+            result_type = amount.Decimal
+            f = lambda x, y: Decimal(x / y)
```
dnicolodi: This looses precision. I think the correct way to write it is `Decimal(x) / Decimal(y)`

beancount/core/amount.py
```diff
@@ -42,6 +42,28 @@ class Amount(_Amount):
...
+        return mul(self, other)
+
+    __rmul__ = __mul__
+
+    def __truediv__(self, other):
+        try:
+            other = Decimal(other)
+        except ValueError:
+            pass
+        return div(self, other)
```
dnicolodi, line 57: The definition of `__mul__` and `__truediv__` are still wrong: no implicit conversions, please. If we add these methods, they need to respect the Python expected semantics.

dnicolodi (commented):
Some remaining comments. Also, please make sure to follow the code formatting conventions: there are many missing spaces after commas and redundant parentheses. `pylint` should warn about those. Also the commit message need to be fixed, something like:
```
bean-query: Allow arithmetic operations on data.Amount dtype

This requires to extend the data.Amount class to support arithmetic dunder methods.
```
would do.

beancount/query/query_compile.py
IguanaBen: I was thinking the output of 1/3 would be a float just like in python. But since we explicitly converting int/int to a decimal, then this isn't necessary.

beancount/query/query_compile.py
IguanaBen: I can break out the int*int case.

beancount/query/query_compile.py
IguanaBen: I don't think you lose precision. In the case of Int/Int, the python return will be a float. If you doe 1/3, then if you take the resulting float and multiply it by 3, you get a float with value 1. And if you then convert it to a Decimal, you'll get a decimal of value 1. But if you convert to Decimal before the division, you get: 0.999999:

```python
>>>Decimal(3*(1/3))
Decimal('1')

>>>Decimal(3*(1/3))
Decimal('1')

```

Is there a use-case I'm not thinking of? In the case of int or decimal divided by decimal, I don't think the conversion makes a difference.

beancount/core/amount.py
IguanaBen: Ok, I removed the explicit conversions and edited the underlying add/ div functions to allow ints as data types.

IguanaBen:
Ok, more changes. I think I fixed the missing spaces for formatting, but I ran pylint on the files and it didn't report any issues with the spacing before, so its hard to know for sure. Is there a flag for pylint to get it to issue a warning on [a,b,c]?

beancount/query/query_compile.py
dnicolodi:
```python
>>> from decimal import Decimal as D
>>> D(1) / D(3)
Decimal('0.3333333333333333333333333333')
>>> D(1 / 3)
Decimal('0.333333333333333314829616256247390992939472198486328125')
```
The first is correct up to the precision of the decimal representation, the second is not.

beancount/query/query_compile.py
```diff
@@ -192,26 +193,45 @@ def __call__(self, context):
 class EvalMul(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x * y)
-        super().__init__(f, left, right, Decimal)
+        if (left.dtype == Amount) and (right.dtype in [int, Decimal]):
```
dnicolodi: You don't need the parentheses.

beancount/query/query_compile.py
```diff
@@ -192,26 +193,45 @@ def __call__(self, context):
...
+            result_type = Amount
+            super().__init__(operator.mul, left, right, result_type)
+        elif right.dtype == Amount and (left.dtype in [int, Decimal]):
```
dnicolodi: ...and indeed you don't have them on the left hand side of the `and` here. But you switch the order of the checks. It is more readable if the check for `left.dtype` always comes first.

beancount/query/query_compile.py
```diff
@@ -192,26 +193,45 @@ def __call__(self, context):
...
+            result_type = Amount
+            super().__init__(operator.mul, left, right, result_type)
+        elif (left.dtype == int) and (right.dtype == int):
+            result_type = int
+            super().__init__(operator.mul, left, right, result_type)
+        elif (left.dtype in [int,Decimal]) and (right.dtype in [int,Decimal]):
```
dnicolodi: Space after comma.

beancount/query/query_compile.py
```diff
@@ -192,26 +193,45 @@ def __call__(self, context):
...
+            result_type = Decimal
+            super().__init__(operator.mul, left, right, result_type)
+        else:
+            raise NotImplementedError("Multiply of {} with {} not supported".format(
```
dnicolodi: Other parts of bean-query use `TypeError` for this kind of errors, I would use the same here)

beancount/query/query_compile.py
```diff
@@ -192,26 +193,45 @@ def __call__(self, context):
 class EvalMul(EvalBinaryOp):
 
     def __init__(self, left, right):
-        f = lambda x, y: Decimal(x * y)
-        super().__init__(f, left, right, Decimal)
+        if (left.dtype == Amount) and (right.dtype in [int, Decimal]):
+            result_type = Amount
```
dnicolodi: I am not sure that having a variable assigned to use its value only at the very next line is an improvement in readability. If you feel that the code is hard to follow like this, you can transform it into a lookup table:
```python
dtypes = {
    (Amoun, int): Amount,
    (int, Amount): Amount,
    (Amont, Decimal): Amount,
    (Decimal, Amount): Amount,
    (int, int): int,
    (int, Decimal): Decimal,
    (Decimal, int): Decimal,
    (Decimal, Decimal): Decimal,
}
dtype = dtypes.get((left.dtype, right.dtype))
if dtype is None:
    raise TypeError(...)
super().__init__(operator.mul, left, right, dtype)
```
which has the benefit of being easily checked for correctness and to be easily extensible.

beancount/query/query_compile.py
IguanaBen: Interesting. Fixed.

beancount/core/amount.py
IguanaBen: Done.

beancount/core/amount.py
```diff
@@ -161,10 +175,10 @@ def mul(amount, number):
     Returns:
       An Amount, with the same currency, but with 'number' times units.
     """
-    assert isinstance(amount.number, Decimal), (
-        "Amount's number is not a Decimal instance: {}".format(amount.number))
-    assert isinstance(number, Decimal), (
-        "Number is not a Decimal instance: {}".format(number))
+    assert isinstance(amount.number, (int, Decimal)), (
+        "Amount's number is not an Int or Decimal instance: {}".format(amount.number))
```
dnicolodi: The type name is `int` not `Int` but anyhow, `amount.number` is supposed to be `Decimal`, not `int`.

beancount/core/amount.py
```diff
@@ -161,10 +175,10 @@ def mul(amount, number):
...
+    assert isinstance(number,(int, Decimal)), (
+        "Number is not an Int or Decimal instance: {}".format(number))
```
dnicolodi: `int` not `Int`

beancount/core/amount.py
```diff
@@ -176,10 +190,10 @@ def div(amount, number):
...
       An Amount, with the same currency, but with amount units divided by 'number'.
     """
-    assert isinstance(amount.number, Decimal), (
-        "Amount's number is not a Decimal instance: {}".format(amount.number))
-    assert isinstance(number, Decimal), (
-        "Number is not a Decimal instance: {}".format(number))
+    assert isinstance(amount.number, (int, Decimal)), (
+        "Amount's number is not an Int or Decimal instance: {}".format(amount.number))
+    assert isinstance(number, (int, Decimal)), (
+        "Number is not an Int or Decimal instance: {}".format(number))
```
dnicolodi: Same here.

beancount/query/query_compile.py
IguanaBen: removed.

beancount/query/query_compile.py
IguanaBen: consolidated into one line and reordered so left always first.

beancount/query/query_compile.py
IguanaBen: Changed

beancount/query/query_compile.py
IguanaBen: That has some benefits.  However it takes 4 conditionals and turns into an array of 8 options, so it also has some drawbacks. For now, I left it with the current structure (but cleaned to up). Let me know if you think this approach is better.

beancount/core/amount.py
IguanaBen: fixed.

beancount/core/amount.py
IguanaBen: fixed

dnicolodi:
If this parch has any future, it is in the beanquery project incarnation of the code. I have some reserves about adding addition and subtraction of Amounts, please see beancount/beanquery#85

blais:
Agreed. Moved to bean-query.  Closing.

PR #692, opened 2022-01-09
bean-query: decimal(str, [int]) - convert str to decimal, round(decimal, int)

mbafford:
**Summary**

Adds `decimal(<str>)`, `decimal(<str>, <Decimal>)` and `round(<Decimal>, <int>)` functions to bean-query.

Merges fine into HEAD and seems to run fine on v3, but I'm not as confident on how to test v3, or the roadmap for bean-query for v3, or if it makes sense to extend bean-query for v3 in this manner.

----
**Background:**

Portfolio balance check-in time - I set up my commodities as follows, and wanted to get a quick summary of US/non-US, stocks/bonds exposure.

```
2016-11-19 commodity INDX
    pcnt_stocks_us: 0.60
    pcnt_stocks_int: 0.40
```

This is *almost* usable out of the box, but `getitem(commodity_meta(<commodity>), <key>)` returns a `str`, so I can't multiply out the values. I worked around this by creating many rows with `group by` and then using Google Sheets to do the final calculations, but with these two functions a nice report portfolio allocation report can easily be done just with bean-query.

With these two functions, I can instead write a query like:

```
select
    sum(convert(position, 'USD')) as total_value,
    round(safediv(
        sum(getprice(commodity(units(position)), 'USD') * decimal(getitem(commodity_meta(commodity(units(position))), 'pcnt_stocks_us'), 0.0) * number(units(position))),
        sum(getprice(commodity(units(position)), 'USD') * number(units(position)))
    )*100.0, 2) as pcnt_value_us_stocks
    where account ~ 'Assets:Retirement' and commodity(units(position)) != 'USD'
```

----

**More Functions?**

As an aside, there are some convenience functions I can see adding, like `commodity_meta(<commodity>, <key>)` to simplify `getitem(commodity_meta(<commodity>), <key>)` , and I'd really love to be able to just do `commodity_meta(<position>, <key>)`.  If PRs for more Python-based functions for bean-query make sense for the v3 roadmap, what's the best way to handle proposing these convenience functions? PR with implementation, GitHub issue, Google Groups email?

### DIFF
--- a/beancount/query/query_env.py
+++ b/beancount/query/query_env.py
@@ -55,6 +55,17 @@ class NegPosition(_Neg):
 class NegInventory(_Neg):
     __intypes__ = [inventory.Inventory]
 
+class RoundDecimal(query_compile.EvalFunction):
+    "Rounds a decimal to the specified number of decimal digits"
+    __intypes__ = [Decimal, int]
+
+    def __init__(self, operands):
+        super().__init__(operands, Decimal)
+
+    def __call__(self, context):
+        args = self.eval_args(context)
+        return round(args[0], args[1])
+
 class ToDecimal(query_compile.EvalFunction):
     """Converts a string into a Decimal type.
     Defaults to no value for non-decimal values."""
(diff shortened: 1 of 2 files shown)


fabform:
Handy to locate google sheet it
[find google sheets id](https://fabform.io/a/how-to-find-google-sheet-id)

mbafford:
It seems like this belongs in beancount/beanquery instead - moved:
https://github.com/beancount/beanquery/pull/20

PR #700, opened 2022-02-06
core: Remove requirement for filename and lineno metadata entries

dnicolodi:
See also https://github.com/beancount/beangulp/pull/101

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -560,8 +560,6 @@ def sanity_check_types(entry, allow_none_for_tags_and_links=False):
     """
     assert isinstance(entry, ALL_DIRECTIVES), "Invalid directive type"
     assert isinstance(entry.meta, dict), "Invalid type for meta"
-    assert 'filename' in entry.meta, "Missing filename in metadata"
-    assert 'lineno' in entry.meta, "Missing line number in metadata"
     assert isinstance(entry.date, datetime.date), "Invalid date type"
     if isinstance(entry, Transaction):
         assert isinstance(entry.flag, (NoneType, str)), "Invalid flag type"
@@ -655,7 +653,7 @@ def entry_sortkey(entry):
       A tuple of (date, integer, integer), that forms the sort key for the
       entry.
     """
-    return (entry.date, SORT_ORDER.get(type(entry), 0), entry.meta["lineno"])
+    return (entry.date, SORT_ORDER.get(type(entry), 0), entry.meta.get("lineno", -1))
 
 
 def sorted(entries):


blais:
Why remove this? This is a nice invariant.
In v3 it's reified as an actual dedicated proto field:
https://github.com/beancount/beancount/blob/master/beancount/ccore/data.proto#L399

Why not instead make it easier to add defaults?

dnicolodi:
There are two issues: sorting entries on a key that contains `entry.meta["lineno"]` and requiring the `lineno` and `filename` metadata entries.

For entries coming from the parser, including `entry.meta["lineno"]` into the sortkey is not necessary because Python's sorting functions are guaranteed to be stable and the parser produces entries in the order found in the source file, thus, all other sortkey entries being equal, entries are always sorted by line number. However, because the line number is in the sort key but not the source file path, when dealing with entries coming from multiple files, the position of the entries in the sorted list depends on the relative location of the entries in the two source files. The sort order for entries that share the same date does not matter for Beancount working properly, however I find this surprising and it takes a while to understand why suddenly entries are ordered differently when lines are inserted in the source file in apparently unrelated locations.

Importers are forced to populate the returned entries with `lineno` and `filename` metadata entries. The `filename` entry is never used except to satisfy the the check removed in this patch, and the `lineno` is used only to sort entries with the default sortkey and to satisfy the same check. For entries coming from the importers, including `entry.meta["lineno"]` into the sortkey is not necessary for the same reasons as above, except when it is actually problematic. The simplest thing for importers is to assign progressive `lineno`s as they come across entries in the source document. When the source document has entries sorted by descending date, the importer could simply return the list of imported entries in reversed order. However, when Beancount later sorts the entries accordingly to the default sortkey, the entries appearing in the same day are sorted by `lineno` and thus the sort order chosen by the importer is reversed.

For the above reasons I would drop `entry.meta["lineno"]` from the sortkey. This patch choose the less invasive approach of making it optional, but I think that dropping it is preferable.

Once the use of `lineno` is removed from the sortkey, there is no reason for importers to include `lineno` and `filename` in the extracted entries, thus the check for their presence can be dropped. Not requiring them is much simpler than having to populate them with fake values. Of course `beangulp` can simply stop using the Beangulp provided integrity check and sortkey, but this does not solve the first issue above.

blais:
This is a great description.

dnicolodi:
`beangulp` now does not use this function anymore (for the reason highlighted above) therefore modifying it is not required.

blais:
I'd like to keep the thread open and eventually revisit the master branch in the light of your description.

PR #701, opened 2022-02-06
bean-sql: Remove

dnicolodi:
bean-sql is an attempt to adapt the Beancount data scheme to an SQL
database to provide a way to extract data from the ledger through SQL
queries, using SQLite as backing SQL engine. SQLite however does not
support a real decimal data type, making any serious use of the tool
challenging.

Users interested in a SQL-like way to query the ledger should look at
bean-query instad. To avoid false starts for new users, remove this
tool from the repository.


blais:
I sort-of disagree on your assessment. bean-query does provide a nice SQL-ish way to query the table of postings, but experimenting with ways to export to a SQL db is still relevant... Maybe someone can come up with a creative way to use that, and that belongs in this project IMO, since it's really just about exposing the list of postings in a different way. I think I can move this to experiments/ instead of scripts, that would make sense.

blais:
Done in https://github.com/beancount/beancount/commit/1f180e717680813931e868ad748fd51e5da876e0

PR #703, opened 2022-02-06
bean-check: Fix auto plugins feature

dnicolodi:
Modifying the module variable holding the list of the plugins to run interferes with running the code as part of the test suite. We probably need a better interface to enable plugins programmatically.

This had me scratching my head for a while when looking at the test results.


blais:
good catch

PR #705, opened 2022-02-16
Feat chinese account

RenLangMan:
The account type regex rule remains unchanged

Add CJK character support for account name subjects


dnicolodi:
I am sorry but these patches are not cannot be reviewed in the current form. At a quick look there are changes in earlier patches that are undone by later ones. Also the comments and the commit messages in Chinese should be translated in English.

blais:
Yeah, we need more context, and some testing.
(Daniele: the comments translate pretty clean in google translate in case you're curious.)

PR #706, opened 2022-02-17
Add CJK character set "Lu_and_CJK" to beancount/utils/regexp_utils.py.

RenLangMan:
Add **CJK character set** "Lu_and_CJK" to `beancount/utils/regexp_utils.py`

Modify `beancount/core/account.py` `ACC_COMP_NAME_RE` defined in the file, make the account name can contain **CJK** characters:

- The **account type** must begin with an **uppercase letter** A-Z;

- The **account name** can contain **CJK characters**, but must be in uppercase or CJK character or digit **beginning**.


dnicolodi:
I haven't looked at the proposed patch in detail, but I think that maintaining the expansion of the unicode ranges in beancount is not ideal: it is hard to get right, it is very hard to verify that the list is complete, and it is almost impossible to test thoroughly. I think either doing the matching "by hand" using string splitting and checking against the unicode codepoint properties via the `unicode` module, or using the `regexp` module https://pypi.org/project/regex/ which support matching against unicode codepoints are better alternatives.

beancount/utils/regexp_utils.py
```diff
@@ -821,6 +819,488 @@
            r'\u2ce2'
            r'\uff21-\uff3a'),
 
+    # Uppercase letters AND CJK character set:(\u4E00–\u9FFC) (\u3400–\u4DBF)
```
blais, line 822: How was this specific list generated?
A comment about how to regenerate it.

blais:
I'm not against trying this out but I'd like this feature behavior to be gated on a conditional so it can be reverted easily (by setting a global flag, e.g., `SUPPORT_LOGOGRAMS = False` if it breaks things.

dnicolodi:
I think it is much easier to depend on the `regexp` package and write proper regular expressions. If there is desire to not add the dependency, the support may be made conditional to the presence of the `regexp` module. But it seems extra unnecessary work to me.

blais:
Sounds like a project. I don't have the cycles to convert v2 to something new and deal with the aftermath.

I propose we either
(a) accept this in v2 with conditionals
(b) accept this in v3 without conditionals
(c) install regexp and do something in v3

dnicolodi:
`regexp` is a drop in replacement for `re`. I expect the patch to be 10 lines long. I propose to leave v2 alone and fix this properly in v3 requiring `regexp`. I can probably work on a patch. I already lost track of the fixes landed in v2 but not in v3 and the other way around. I would try to minimize the differences. 

I suspect that there are a number of things that will need to be forward-ported from v2 to v3. I'll see if I can work on that.

blais:
Thanks for doing the patch Daniele, it's indeed short. PR rebased and merged.

RenLangMan: Do you mind rebasing, removing the functional changes (it should work with Daniele's patch) and keeping the unit tests? I wouldn't want to lose those tests.
Let me know,

PR #707, opened 2022-02-20
ci: Fix workflow to upload wheels to PyPI

dnicolodi:
This should fix publishing wheels when a tag is pushed that does not point to HEAD and when the workflow is triggered manually. @blais you should be able to push wheels for the 2.3.5 release going to the project Actions tab, selecting "upload-wheel" then "Run workflow" and punching in "2.3.5" in the "Release tag" field.

### DIFF
--- a/.github/workflows/pypi-wheel.yaml
+++ b/.github/workflows/pypi-wheel.yaml
@@ -2,11 +2,12 @@ name: upload-wheel
 on:
   push:
     tags:
-      - "2.*"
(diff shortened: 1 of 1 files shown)


blais:
done.
Thank you!

dnicolodi:
I made a mistake somewhere: https://github.com/beancount/beancount/actions/runs/1873310762 I wonder how it worked in my tests. I'll send a fix for the fix.

PR #715, opened 2022-04-10
format: add support for parenthesized expressions

### LINKED ISSUE #714: `bean-format` does not support arithmetic expressions in balance assertions
The following input:

```beancount
2000-01-01 balance Assets:Example (10.00 - 0.01) EUR
2000-01-01 balance Assets:Example 9.99 EUR
```

Run through `bean-format -c 80`, leads to the following output;

```beancount
2000-01-01 balance Assets:Example (10.00 - 0.01) EUR
2000-01-01 balance Assets:Example                                         9.99 EUR
```

ambroisie:
Closes #714 with *very* limited support for arithmetic expressions: only parenthesized, binary operations are matched.

### DIFF
--- a/beancount/scripts/format_test.py
+++ b/beancount/scripts/format_test.py
@@ -244,6 +244,44 @@ class TestScriptFormat(test_utils.ClickTestCase):
 
         """), result.stdout)
 
+    @test_utils.docfile
+    def test_parenthesized_binary_expressions(self, filename):
+        """
+        2016-08-01 open Expenses:Test
+        2016-08-01 open Assets:Test
+
+        2016-08-02 * "" ""
+          Expenses:Test     10.0 USD
+          Expenses:Test (10.0/2) USD
+          Expenses:Test (10.0+2) USD
+          Expenses:Test (10.0-2) USD
+          Expenses:Test (10.0*2) USD
+          Expenses:Test (-10.0*+2) USD
+          Expenses:Test (-1,000.0*+2.0) USD
+          Assets:Test
+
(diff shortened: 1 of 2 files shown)


beancount/scripts/format_test.py
```diff
@@ -244,6 +244,44 @@ def test_arithmetic_expressions(self, filename):
...
+        2016-08-01 open Assets:Test
+
+        2016-08-02 * "" ""
+          Expenses:Test     10.0 USD
+          Expenses:Test (10.0/2) USD
+          Expenses:Test (10.0+2) USD
+          Expenses:Test (10.0-2) USD
+          Expenses:Test (10.0*2) USD
+          Expenses:Test (-10.0*+2) USD
+          Expenses:Test (-1,0.0*+2.0) USD
```
dnicolodi: This looks like a funny number. Updating the regexp to reject these may be overkill, but I would at least use a proper thousands separator in the example, ie `-1,000.00` instead than `-1,0.0`.

dnicolodi:
Thanks for the patch @ambroisie It looks reasonable to me, but I don't use `bean-format` and I am not sure I understand all implications of this change.

beancount/scripts/format_test.py
ambroisie: > Updating the regexp to reject these may be overkill

Especially since different cultures use different separators (e.g: [the Indian numbering system](https://en.wikipedia.org/wiki/Indian_numbering_system)) :-).

I can update the test however.

ambroisie:
Bump, is there anyone else who should review this?

PR #722, opened 2022-06-22
Add HIFO booking method

einarsi:
This adds a new booking method called HIFO (Highest In, First Out). With HIFO you sell the lots with the highest cost basis first. HIFO is commonly used for crypto, which in many countries has arbitrary booking requirements for tax reporting. HIFO is often a better choice than other booking methods since it leads to the lowest gains and largest losses, thus minimizing taxes for for the year you are reporting.


blais:
Elegant and simple. v2 is normally frozen but this is simple enough, harmless enough, has a test... approved and merging.

PR #726, opened 2022-08-09
printer: do not sort metadata key upon printing

### LINKED ISSUE #420: Make metadata into an OrderedDict
**[Original report](https://bitbucket.org/blais/beancount/issue/420) by Martin Michlmayr (Bitbucket: [tbm13](https://bitbucket.org/tbm13), GitHub: [tbm](https://github.com/tbm)).**

I’m wondering whether it would be possible to make metadata \(both on transactions and postings\) into an OrderedDict.

I have written an importer which basically does something like:

```
            meta = data.new_metadata(f.name, index)
            meta['donor_name'] = get_name(row)
            meta['donor_email'] = row['Email']
```

It makes sense to have name first, email second, but currently beancounts sorts the keys and email comes out first.

Can you think of any disadvantages of using an OrderedDict? If not, maybe I can try to see if I can contribute a patch.

‌

zacchiro:
This makes the metadata key ordering fall back to the "natural" dictionary
ordering that, sarting with Python 3.7, is insertion order (as it was the case
for OrderedDicts with Python < 3.7).

As part of this change we hence also bump up Python required version to 3.7, so
that Beancount importers can impose a desired metadata ordering.

Closes #420.


zacchiro:
About the CI failures, they seem all due to the fact "chardet" does not get installed in the CI environment. I'm not sure what's going on, but it seems unrelated to the proposed changes.
(**Update**: yep, previous commits in the v2 branch fail the CI in the same way. So this should not be a blocker.)

dnicolodi:
The CI uses the `requirements.txt` file in the repo to install dependencies. However, `requirements.txt` is incomplete as it at least misses the dependency on `chardet`. So far we got `chardet` installed in the CI environment because of a transitive dependency via `requests`. However, since release 2.26 https://requests.readthedocs.io/en/latest/community/updates/#id5 `requests` does not depend on `chardet` anymore, preferring `charset_normalizer` to it, unless `chardet` is already installed.

`requirements.txt` needs to be updated. Probably `setup.py` should be updated to require `requests` to use `chardet` otherwise it may be that `pip` ends up installing both `charset_normalizer` and `chardet`.

zacchiro:
> The CI uses the `requirements.txt` file in the repo to install dependencies. However, `requirements.txt` is incomplete as it at least misses the dependency on `chardet`.

All this seems unrelated to this PR, but just in case I've just pushed an additional separate commit that updates requirements.txt (and also setup.py, but I'm not sure how to enforce that requrests relies on chardet; for now I'm just hoping the older in the list in setup.py would be enough). It has fixed all of the chardet-related issues.

Another test (related to linting) fails now, but that too is unrelated to this PR.

dnicolodi:
> All this seems unrelated to this PR

Sure, I didn't want to imply that the fix had to be part of this PR. I'm sorry to have given you this impression. I just wanted to record somewhere my notes on what is going on as I didn't have time to prepare a PR to fix the issue.

blais:
Not sure this is going to cause problems but I don't have a clear view; let's find out.
Okay with requiring 3.7

blais (approved):

PR #728, opened 2022-08-15
ci(v2): add python 3.10 and remove 3.6

trim21:
some options are removed from pylint's new version, also remove them from pylintrc.


trim21:
😂 Oh sorry, didn't notice currently v2 is behind master and no need to be updated anymore.

PR #730, opened 2022-09-13
Don't set quotechar in csv register_dialect

kanru:
quotechar is not allowed to be empty after [Python 3.11](https://docs.python.org/3.11/library/csv.html#csv.Dialect.quotechar)

quoting is already QUOTE_NONE so quotechar is not relevant and can be safely ignored.


blais:
Somebody else changed the quotechar to '"'.

PR #745, opened 2023-01-19
Fix a build failure in SystemError on all but macOS

omasanori:


beancount/utils/errors.cc
```diff
@@ -22,13 +23,20 @@ absl::StatusCode kSystemError = absl::StatusCode(kUserErrors + 1);
...
+  strerror_r(status, buffer, 2048);
+  status = errno;
 #else
   int status = strerror_r(errno, buffer, 2048);
-  assert(status == 0);
 #endif
 
+  // TODO: Handle errors raised by strerror_r correctly.
+  assert(status == 0);
+
```
dnicolodi, line 27: The right way to fix this is a configure time check for the `strerror_r()` function. I'm sure Bazel should have an easy way to do it. However, what's the point to trade a compilation error with an assertion failure at runtime?

omasanori: Thank you for reviewing!

The assertion here, derived from the present code, checks if strerror_r itself fails run-time. For example, the function has no message string for given errno for some reasons.

The proposed code, however, is not ideal as no precise error message does not mean it is a fatal error. Returning an error indicating a failure to retrieve an appropriate error message may be better. Your opinion?

dnicolodi: Sorry, I misread the code.

omasanori: Added a proper error handling.

beancount/utils/errors.cc
```diff
@@ -22,13 +23,21 @@ absl::StatusCode kSystemError = absl::StatusCode(kUserErrors + 1);
 
 absl::Status SystemError(absl::string_view message) {
   char buffer[2048];
-#ifndef __APPLE__
-  description = strerror_r(errno, buffer, 2048);
+
+#ifdef __GLIBC__
+  // GNU libc provides non-XSI-compliant strerror_r.
```
dnicolodi: Accordingly to the man page https://linux.die.net/man/3/strerror_r GNU libc can be made to use the XSI-compliant version of `strerror_r()`. I would much prefer this to the `#ifdef`s here. 

Also, the only documented non-zero return values for `strerror_r()` I've found are `EINVAL` and `ERANGE` for an unrecognized error number and a too short buffer, respectively. However, in both cases the buffer should be filled with an error message that makes sense: an `Unknown error X` in the first case and a truncated message in the second. I think that ignoring the return value of `strerror_r()` is perfectly fine.

omasanori: Indeed. I will dig into Bazel documentation. Thank you for your comments!

omasanori: So, if we ignore the return value of `strerror_r`, we don't need to distinguish between XSI version and GNU version. It simplifies the code to just three lines.

blais:
Why remove the check? What's the motivation? I like assertions.

omasanori:
Thank you for your comment, @blais!

My conclusion is that `strerror_r` fills an appropriate error message even in the cases the function itself fails to fill the original error message, as @dnicolodi [pointed out](https://github.com/beancount/beancount/pull/745#discussion_r1082239635):
> Also, the only documented non-zero return values for `strerror_r()` I've found are `EINVAL` and `ERANGE` for an unrecognized error number and a too short buffer, respectively. However, in both cases the buffer should be filled with an error message that makes sense: an `Unknown error X` in the first case and a truncated message in the second. I think that ignoring the return value of `strerror_r()` is perfectly fine.

omasanori:
While I think the assertion here is not necessary, an alternative fix would be:
```diff
diff --git a/beancount/utils/errors.cc b/beancount/utils/errors.cc
index a89b2def..a85beebd 100644
--- a/beancount/utils/errors.cc
+++ b/beancount/utils/errors.cc
@@ -1,6 +1,10 @@
 // Copyright (C) 2020  Martin Blais
 // License: "GNU GPLv2"

+// Use XSI-compliant strerror_r with GNU libc
+#undef _GNU_SOURCE
+#define _XOPEN_SOURCE 700
+
 #include "beancount/utils/errors.h"

 #include "absl/strings/str_cat.h"
@@ -22,12 +26,9 @@ absl::StatusCode kSystemError = absl::StatusCode(kUserErrors + 1);

 absl::Status SystemError(absl::string_view message) {
   char buffer[2048];
-#ifndef __APPLE__
-  description = strerror_r(errno, buffer, 2048);
-#else
+
   int status = strerror_r(errno, buffer, 2048);
   assert(status == 0);
-#endif

   return absl::Status(kSystemError, absl::StrCat(message, ": ", buffer));
 }
```
Is this preferred, @blais? If so, I am happy to switch the approach.

PR #751, opened 2023-03-16
Add a new 'close_tree' plugin to close entire trees by closing the pa…

redstreet:
Add a new 'close_tree' plugin to close entire trees by closing the parent.

See https://groups.google.com/g/beancount/c/c0Fm4hrxpUQ

### DIFF
--- a/beancount/ops/validation.py
+++ b/beancount/ops/validation.py
@@ -87,7 +87,7 @@ def validate_open_close(entries, unused_options_map):
             else:
                 try:
                     open_entry = open_map[entry.account]
-                    if entry.date <= open_entry.date:
+                    if entry.date < open_entry.date:
                         errors.append(
                             ValidationError(
                                 entry.meta,
--- a/beancount/plugins/close_tree.py
+++ b/beancount/plugins/close_tree.py
@@ -1,0 +1,61 @@
+"""This plugin inserts close directives for all of an account's descendants when an account is
+closed. Unopened parent accounts can also be closed. Any explicitly specified close is left
+untouched.
+
+For example, given this::
+
+    2017-11-10 open Assets:Brokerage:AAPL
+    2017-11-10 open Assets:Brokerage:ORNG
+    2018-11-10 close Assets:Brokerage  ; this account does not necessarily need to be opened
+
+the plugin turns it into::
+
+    2017-11-10 open Assets:Brokerage:AAPL
+    2017-11-10 open Assets:Brokerage:ORNG
+    2018-11-10 close Assets:Brokerage:AAPL
+    2018-11-10 close Assets:Brokerage:ORNG
+
+Invoke this plugin _after_ any plugins that generate `open` directives for account trees that you
+want to auto close. An example is the `auto_accounts` plugin that ships with Beancount::
+
+    plugin "beancount.plugins.auto_accounts"
+    plugin "beancount.plugins.close_tree"
+
+"""
+
+from beancount.core import data
+from beancount.core.data import Open, Close
+
+__plugins__ = ('close_tree',)
+
+
+def close_tree(entries, unused_options_map):
+    """Insert close entries for all subaccounts of a closed account.
+
+    Args:
+      entries: A list of directives. We're interested only in the Open/Close instances.
+      unused_options_map: A parser options dict.
+    Returns:
+      A tuple of entries and errors.
+   """
+
+    new_entries = []
+    errors = []
+
+    opens = set(e.account for e in entries if isinstance(e, Open))
+    closes = set(e.account for e in entries if isinstance(e, Close))
+
+    for entry in entries:
+        if isinstance(entry,  Close):
+            subaccounts = [a for a in opens if a.startswith(entry.account + ':') and a not in closes]
+            for subacc in subaccounts:
+                meta = data.new_metadata('<beancount.plugins.close_tree>', 0)
+                close_entry = data.Close(meta, entry.date, subacc)
+                new_entries.append(close_entry)
+                closes.add(subacc)  # So we don't attempt to re-close a grandchild that a child closed
+            if entry.account in opens:
+                new_entries.append(entry)
+        else:
+            new_entries.append(entry)
+
+    return new_entries, errors
--- a/beancount/plugins/close_tree_test.py
+++ b/beancount/plugins/close_tree_test.py
@@ -1,0 +1,167 @@
+import unittest
+
+from beancount.plugins import close_tree
+from beancount.parser import options
+from beancount.parser import cmptest
+from beancount import loader
+
+
+class TestCloseTree(cmptest.TestCase):
+    def test_empty_entries(self):
+        new_entries, _ = close_tree.close_tree([], options.OPTIONS_DEFAULTS.copy())
+        self.assertEqual([], new_entries)
+
+    @loader.load_doc()
+    def test_basic(self, entries, _, options_map):
+        """
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
+            2015-01-01 close Assets:XBank
+        """
+        new_entries, _ = close_tree.close_tree(entries, options_map)
+
+        self.assertEqualEntries("""
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
+            2015-01-01 close Assets:XBank
+            2015-01-01 close Assets:XBank:AAPL
+        """, new_entries)
+
+    @loader.load_doc()
+    def test_leave_others_untouched(self, entries, _, options_map):
+        """
+            2014-01-01 open Assets:YBank
+            2014-01-01 open Assets:YBank:AAPL
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
+            2014-01-01 open Assets:XBank:AAPL:Fuji
+            2014-01-01 open Assets:XBank:AAPL:Gala
+            2014-01-01 open Assets:XBank:ORNG
+            2014-01-01 open Assets:XBank:BANANA
+            2015-01-01 close Assets:XBank
+        """
+
+        new_entries, _ = close_tree.close_tree(entries, options_map)
+        self.assertEqualEntries("""
+            2014-01-01 open Assets:YBank
+            2014-01-01 open Assets:YBank:AAPL
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
+            2014-01-01 open Assets:XBank:AAPL:Fuji
+            2014-01-01 open Assets:XBank:AAPL:Gala
+            2014-01-01 open Assets:XBank:ORNG
+            2014-01-01 open Assets:XBank:BANANA
+            2015-01-01 close Assets:XBank
+            2015-01-01 close Assets:XBank:AAPL
+            2015-01-01 close Assets:XBank:AAPL:Fuji
+            2015-01-01 close Assets:XBank:AAPL:Gala
+            2015-01-01 close Assets:XBank:ORNG
+            2015-01-01 close Assets:XBank:BANANA
+        """, new_entries)
+
+    @loader.load_doc()
+    def test_override(self, entries, _, options_map):
+        """
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
+            2014-01-01 open Assets:XBank:AAPL:Fuji
+            2015-01-01 close Assets:XBank:AAPL:Fuji
+            2016-01-01 close Assets:XBank
+        """
+        new_entries, _ = close_tree.close_tree(entries, options_map)
+
+        self.assertEqualEntries("""
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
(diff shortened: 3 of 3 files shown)


beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+    for entry in new_entries:
+        if isinstance(entry, Open) and (entry.account, entry.date) in closes:
+            continue
+        if isinstance(entry, Close) and (entry.account, entry.date) in opens:
+            continue
+        retval.append(entry)
+
+    if DEBUG:
+        elapsed_time = time.time() - start_time
+        print("Close account tree [{:.2f}s]: {} close entries added.".format(elapsed_time, close_count))
```
dnicolodi: From this I infer that you are worried about the execution time of this plugin. If this is the case, maybe storing the accounts tree in a flat list is not the best choice. You could use a tree data structure. Also iterating the list of entries six times is not optimal. If there are no performance issues, I would delete the debug code, as it is anyhow not discoverable or accessible by the users and there are better profiling tools.

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+    # 2021-06-17 close Expenses:Non-Retirement:Auto:Fit
+    # plugin "beancount.plugins.close_tree"
+    # 2019-01-01 * "Transaction"
+    #   Expenses:Non-Retirement:Auto:Fit:Insurance      10 USD
+    #   Expenses:Non-Retirement:Auto:Fit:Gas            20 USD
+    #   Assets:Transfer
+
+    # Above, auto_accounts inserts an Open for 'Expenses:Non-Retirement:Auto:Fit:Gas' on 2021-06-17
+    # since it appears in a close directive. But that's not what we want. The semantic of the close
+    # when u used with this plugin is: "close all children, and close the parent if it was opened
```
dnicolodi: Spurious `u`.

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+                new_entries.append(entry)
+        else:
+            new_entries.append(entry)
+
+    # beancount.plugins.auto_accounts, if used in conjunction with this plugin, inserts an Open
+    # entry for empty parent accounts which we are trying to close. For example:
+
+    # plugin "beancount.plugins.auto_accounts"
+    # 2021-06-17 close Expenses:Non-Retirement:Auto:Fit
+    # plugin "beancount.plugins.close_tree"
```
dnicolodi: I would put both `plugin` directives at the beginning.

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+    # 2019-01-01 * "Transaction"
+    #   Expenses:Non-Retirement:Auto:Fit:Insurance      10 USD
+    #   Expenses:Non-Retirement:Auto:Fit:Gas            20 USD
+    #   Assets:Transfer
+
+    # Above, auto_accounts inserts an Open for 'Expenses:Non-Retirement:Auto:Fit:Gas' on 2021-06-17
+    # since it appears in a close directive. But that's not what we want. The semantic of the close
+    # when u used with this plugin is: "close all children, and close the parent if it was opened
+    # earlier. Below, we filter out entries to correctly handle this scenario:
```
dnicolodi: I had a very hard time to follow what you are doing here. I had to unpack the different steps. Starting with this ledger:
```
2019-01-01 * "Transaction"
  Expenses:Non-Retirement:Auto:Fit:Insurance      10 USD
  Expenses:Non-Retirement:Auto:Fit:Gas            20 USD
  Assets:Transfer
2021-06-17 close Expenses:Non-Retirement:Auto:Fit
```
the `bencount.plugins.auto_accounts` adds the `open` directives:
```
2019-01-01 open Assets:Transfer
2019-01-01 open Expenses:Non-Retirement:Auto:Fit:Insurance
2019-01-01 open Expenses:Non-Retirement:Auto:Fit:Gas
2019-01-01 * "Transaction"
  Expenses:Non-Retirement:Auto:Fit:Insurance      10 USD
  Expenses:Non-Retirement:Auto:Fit:Gas            20 USD
  Assets:Transfer
2021-06-17 open Expenses:Non-Retirement:Auto:Fit
2021-06-17 close Expenses:Non-Retirement:Auto:Fit
```
Note that this this generates an error, because you have the open and the close directive on the same day, and this is not currently allowed. Let's ignore this for the moment. The `beancount.plugins.close_tree` adds the missing `close` directives based on the `open` directives present in the ledger:
```
2019-01-01 open Assets:Transfer
2019-01-01 open Expenses:Non-Retirement:Auto:Fit:Insurance
2019-01-01 open Expenses:Non-Retirement:Auto:Fit:Gas
2019-01-01 * "Transaction"
  Expenses:Non-Retirement:Auto:Fit:Insurance      10 USD
  Expenses:Non-Retirement:Auto:Fit:Gas            20 USD
  Assets:Transfer
2021-06-17 open Expenses:Non-Retirement:Auto:Fit
2021-06-17 close Expenses:Non-Retirement:Auto:Fit
2021-06-17 close Expenses:Non-Retirement:Auto:Fit:Insurance
2021-06-17 close Expenses:Non-Retirement:Auto:Fit:Gas
```
What you want to do is to remove the pair of `open` and `close` directives that appear on the same day. But waht if there is also a transaction to the relative account on the same day? You would need to keep the `open` and `close` directives. I think that the error message for `open` and `close` on the same day is not correct: Beancount puts `open` directives at the beginning of day and `close` directives at the end of day, and transactions in between. Thus I don't see a problem with `open` and `close` on the same day.

dnicolodi (changes requested):

beancount/plugins/close_tree.py
blais: +1

beancount/plugins/close_tree.py
blais: +1

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+    close_count = 0
+    new_entries = []
+    errors = []
+
+    opens = [e.account for e in entries if isinstance(e, Open)]
+    closes = [e.account for e in entries if isinstance(e, Close)]
+
+    for entry in entries:
+        if isinstance(entry,  Close):
+            subaccounts = [a for a in opens if entry.account + ':' in a and a not in closes]
```
blais: Use a set for speed

blais: a.startswith(entry.account+ ":") or use a regexp

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+            for s in subaccounts:
```
blais: Don't use single letter variables; rename to "account" or "sub" or something.

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,98 @@
...
+    Returns:
+      A tuple of entries and errors. """
+
+    start_time = time.time()
+    close_count = 0
+    new_entries = []
+    errors = []
+
+    opens = [e.account for e in entries if isinstance(e, Open)]
+    closes = [e.account for e in entries if isinstance(e, Close)]
```
blais: If you wanted to be more precise about closing really just the subaccounts that are open at the time of closing, you should keep track of the tree of open accounts as you process the directives and close the subaccounts that are open then. You'd also have to handle subsequent closing directives colliding with your inserted ones. But I think that would be the most correct algorithm.

blais (changes requested):

beancount/plugins/close_tree.py
blais: To be more clear what I mean is that it's possible for accounts to be opened, closed and then reopened (I think, I forget a bit).

beancount/plugins/close_tree.py
redstreet: Speed is not a concern, so leaving this as such (technically, I believe a list might be slightly faster in this case as the hashing of sets is avoided, and we don't do the expensive lookups or removals).

redstreet: Changed.

beancount/plugins/close_tree.py
redstreet:
```
2014-01-01 open Assets:XBank
2015-01-01 close Assets:XBank

2016-01-01 open Assets:XBank
2017-01-01 close Assets:XBank   
```

This doesn't seem to be supported (in v2, at least). I get:
```
Duplicate open directive for Assets:XBank
Duplicate close directive for Assets:XBank
```

Do we still need to address this case? Is there a plan to support it in v3 or in the future?

beancount/plugins/close_tree.py
redstreet: 1. Fixed Beancount's error message.
2. Do we need to remove the pair, if there wasn't a transaction found to the account in question on that date? We would save the potentially expensive cost of checking if a removal is needed, if we instead always inserted the `open`/`close` pair (and sometimes ended up with this additional pair).

beancount/ops/validation.py
```diff
@@ -87,7 +87,7 @@ def validate_open_close(entries, unused_options_map):
             else:
                 try:
                     open_entry = open_map[entry.account]
-                    if entry.date <= open_entry.date:
+                    if entry.date < open_entry.date:
```
dnicolodi, line 90: This should be a separate commit with a commit message that explain what is being changed and why. That said, I haven't checked if there is a reason for not allowing `open` and `close` directives on the same day. @blais do you remember anything about this?

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Invoke this plugin _after_ any plugins that generate `open` directives for account trees that you
+want to auto close. An example is the `auto_accounts` plugin that ships with Beancount:
+
+```
+plugin "beancount.plugins.auto_accounts"
+plugin "beancount.plugins.close_tree"
+```
+"""
```
dnicolodi, line 24: Documentation strings do not use Markdown syntax. If anything it should be reST.

beancount/plugins/close_tree.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def close_tree(entries, unused_options_map):
+    """Insert close entries for all subaccounts of a closed account.
+
+    Args:
+      entries: A list of directives. We're interested only in the Open/Close instances.
+      unused_options_map: A parser options dict.
+    Returns:
+      A tuple of entries and errors. """
```
dnicolodi: The closing `"""` should go on a line by themselves.

beancount/plugins/close_tree_test.py
```diff
@@ -0,0 +1,192 @@
...
+            2014-01-01 open Assets:XBank
+            2014-01-01 open Assets:XBank:AAPL
+            2015-01-01 close Assets:XBank
+            2015-01-01 close Assets:XBank:AAPL
+        """, dedent=True)
+
+        actual, _ = close_tree.close_tree(entries, {})
+        for a, e in zip(s(actual), s(expected)):
+            self.assertEqual(a.date, e.date)
+            self.assertEqual(a.account, e.account)
```
dnicolodi: Here and everywhere else, is there a reason why `self.assertEqual(actual, expected)` does not work? Also, why is there the need to sort the actual and expected entries, can't the expected entries be in the right order?

dnicolodi (changes requested):

beancount/plugins/close_tree_test.py
redstreet: 1. Doesn't work because metadata is not the same
2. Sorting the expected entries as the actual would have been sorted makes it unintuitive IMO.
```
            2019-01-01 open Expenses:Non-Retirement:Auto:Fit:Gas
            2019-01-01 open Expenses:Non-Retirement:Auto:Fit:Insurance
            ....
            2021-06-17 open Expenses:Non-Retirement:Auto:Fit
            2021-06-17 close Expenses:Non-Retirement:Auto:Fit:Insurance
            2021-06-17 close Expenses:Non-Retirement:Auto:Fit:Gas
            2021-06-17 close Expenses:Non-Retirement:Auto:Fit
```
The above is neither alphabetical, nor consistent (ordering of opens and closes), nor does the open/close pair intuitively appear next to each other. In addition, changes to the plugin's algorithm might require all expected cases to be reordered, which is messy.

beancount/plugins/close_tree.py
redstreet: Ah, found the [documentation](https://beancount.github.io/docs/beancount_language_syntax.html#directives):
> _At the moment, once an account is closed, you cannot reopen it after that date._

@blais, do you still see this to be a point of concern?

beancount/plugins/close_tree_test.py
dnicolodi: Please have a look, for example, at how the tests for the `auto_accounts.py` plugin are structured and follow the same pattern.

beancount/ops/validation.py
blais: I don't remember.
Il'l review now.

blais: It does make sense to me that we should allow entries on the same day as the open though. Accepting.

blais (approved):

PR #758, opened 2023-06-01
v2: use new `NamedTuple` syntax

trim21:


trim21:
tests passed locally, is ci broken?

blais:
The master branch is already doing this. 
No need to make changes to v2.

PR #765, opened 2023-07-25
Adds a hermetic setup for python 3.10

filmil:
This allows downloading and compiling beancount latest without any special python package installation.

Hopefully that's something that we want to have.


filmil:
Not sure if this is something of value, though. Please let me know.

2 tests in `//experiments/...` fail, for non-obvious reasons. Also, Not sure what would happen when CI runs.

filmil:
I'm not sure if the tests //experiments/{protos,pbpb}/... should pass today or not.

filmil:
Yup, the same tests fail on test run on clean repo:

```
env PYTHON_BIN_PATH=$(which python3) bazel test //...
```

Results in:

```
┬─[filmil@instance-3:~/code/beancount2]─[08:59:54]                                                                                                                                                                                                                                        
│ (g/b:master)                                                                                                                                                                                                                                                                            
╰─>$ env PYTHON_BIN_PATH=(which python3) bazel test //...                                                                                                                                                                                                                                 
INFO: Analyzed 274 targets (0 packages loaded, 0 targets configured).                                                                                                                                                                                                                     
INFO: Found 177 targets and 97 test targets...                                                                                                                                                                                                                                            
FAIL: //experiments/v3/protos:expose_protos_test (see /home/filmil/.cache/bazel/_bazel_filmil/05b429cd0a0ec0b2a4eb6a2faee8edc1/execroot/beancount/bazel-out/k8-fastbuild/testlogs/experiments/v3/protos/expose_protos_test/test.log)
INFO: From Testing //experiments/v3/protos:expose_protos_test:                                                                               
==================== Test output for //experiments/v3/protos:expose_protos_test:       
EE                                                                                                                                           
======================================================================                 
ERROR: test_parse_and_convert (__main__.ExportProtosTests)                                                                                   
----------------------------------------------------------------------                 
  File "/home/filmil/.cache/bazel/_bazel_filmil/05b429cd0a0ec0b2a4eb6a2faee8edc1/sandbox/linux-sandbox/842/execroot/beancount/bazel-out/k8-fastbuild/bin/experiments/v3/protos/expose_protos_test.runfiles/beancount/experiments/v3/protos/expose_protos_test.py", line 37, in test_parse_
and_convert                                                                                                                                  
    ledger = ep.parse(self.filename)                                                                                                         
AttributeError: 'NoneType' object has no attribute 'parse'                                                                                   
                                                                                                                                             
======================================================================                 
ERROR: test_time_parsing (__main__.ExportProtosTests)                                                                                        
----------------------------------------------------------------------                 
  File "/home/filmil/.cache/bazel/_bazel_filmil/05b429cd0a0ec0b2a4eb6a2faee8edc1/sandbox/linux-sandbox/842/execroot/beancount/bazel-out/k8-fastbuild/bin/experiments/v3/protos/expose_protos_test.runfiles/beancount/experiments/v3/protos/expose_protos_test.py", line 31, in test_time_p
arsing                                                                                                                                       
    ledger = ep.parse(self.filename)                                                                                                         
AttributeError: 'NoneType' object has no attribute 'parse'                                                                                   
                                                                                                                                             
----------------------------------------------------------------------                 
Ran 2 tests in 0.001s                                                                                                                        
                                                                                                                                             
FAILED (errors=2)                                                                                                                            
================================================================================       
FAIL: //experiments/v3/pbpb:pbpb_protos_test (see /home/filmil/.cache/bazel/_bazel_filmil/05b429cd0a0ec0b2a4eb6a2faee8edc1/execroot/beancount/bazel-out/k8-fastbuild/testlogs/experiments/v3/pbpb/pbpb_protos_test/test.log)
INFO: From Testing //experiments/v3/pbpb:pbpb_protos_test:                                                                                   
==================== Test output for //experiments/v3/pbpb:pbpb_protos_test:           
  File "/home/filmil/.cache/bazel/_bazel_filmil/05b429cd0a0ec0b2a4eb6a2faee8edc1/sandbox/linux-sandbox/843/execroot/beancount/bazel-out/k8-fastbuild/bin/experiments/v3/pbpb/pbpb_protos_test.runfiles/beancount/experiments/v3/pbpb/pbpb_protos_test.py", line 17, in <module>
    from experiments.v3.protos_new import pbpb_protos as ep                                                                                  
ModuleNotFoundError: No module named 'experiments.v3.protos_new'                                                                             
================================================================================       
INFO: Elapsed time: 0.702s, Critical Path: 0.40s                                                                                             
INFO: 3 processes: 4 linux-sandbox.                                                                                                          
INFO: Build completed, 2 tests FAILED, 3 total actions                                      
```

requirements/dev.txt
```diff
@@ -1,4 +1,12 @@
...
+click>=7.0
+google-api-python-client>=1.8.2
+google_auth_oauthlib>=0.4.2
+httplib2>=0.10
+lxml>=3.0
+oauth2client>=4.0
+ply>=3.4
+python-dateutil>=2.6.0
+requests>=2.0
 pylint>=2.6.0
```
dnicolodi: Why? The point of this is to do not to have to keep the dependencies in sync in two different files. I suspect that Bazel does not support the `requirements.txt` file format, but I don't think this is a thing that we should work around here.

filmil: I haven't found a way to pass tools.txt to `compile_pip_requirements` sandbox. [See here][cpr]. The `extra_deps` parameter can not be it, as it requires a `PyPi` provider, which a source file of course doesn't have

[cpr]: https://github.com/bazelbuild/rules_python/blob/main/docs/pip.md#compile_pip_requirements

Even though the underlying `pip` invocation will look for it, there currently does not seem to be a way to add that file to the sandbox where `compile_pip_requirements` is executing. This means that the rule will fail when executed with `File not found: requirements/tools.txt`.  

Since in my view not having a hermetic build is a bigger issue than some duplication, I opted to go this route.

filmil: Filed https://github.com/bazelbuild/rules_python/issues/1347

dnicolodi: Using bazel for deploying Python applications is an extremely niche use case, therefore I'm not very sympathetic to the idea of making the life of the Beancount maintainers harder to satisfy this use case. Bazel is developed by an organization with very deep pockets that employs a huge number of engineers. Ask them to fix their tools.

dnicolodi: Also, why do you install the development dependencies? It seems an odd choice for what I understand are the goals of an hermetic installation, at least as a default. Also, please note that the Bazel build definition pulls in a lot of components and compiles code that is not used anywhere. I would expect that an hermetic build wants to cut dependencies to a minimum, not pull in a truckload of unused components.

filmil: > Using bazel for deploying Python applications is an extremely niche use case

I suppose it is, and yet beancount has bazel support right there. Perhaps you want to add a `CONTRIBUTING.md`  to clarify your expectations?

Else, if it is a niche use case, why even use bazel?

> I'm not very sympathetic to the idea of making the life of the Beancount maintainers harder to satisfy this use case.

Sorry, there was only so much I could do in an evening after work.

> Bazel is developed by an organization with very deep pockets that employs a huge number of engineers. Ask them to fix their tools.

I filed a bug and will look into this when I have the time. Is there more I could be doing here?

filmil: > Also, why do you install the development dependencies?

My goal here was to just rewrite what was already there in starlark. If there is a different way one should set this up, I welcome a pointer. A difference between `dev.txt` and `tools.txt` is not obvious to me.

filmil: OK, I learned how to remove the duplication in dev.txt vs tools.txt, and the original issue in this thread is now fixed.

blais:
> Not sure if this is something of value, though. Please let me know.
> 

It's interesting indeed!

filmil:
> It's interesting indeed!

Tell us more?

Would you be game to accept the PR, in this form, or with edits? Is this something of interest to you regarding bazel support in beancount?

blais:
Yes, isolating the environment is great.
I just have to find some free time to test it out.

filmil:
No worries, I'm not in a rush.

I just wanted to get some signal of whether this was something of interest. Happy to wait as long as it takes now that it seems that it is.

blais:
Yes, I'm very interested. Bazel provides good isolation.
Another thing I'm interested in is setting up a working build for Rust.
I was before the summer toying around with the idea of using Rust instead of C++ and was having problems getting things to build properly. (I'm not going to use cargo).

filmil:
> Yes, I'm very interested.

:+1: 

> Another thing I'm interested in is setting up a working build for Rust. [...] (I'm not going to use cargo).

Here's my rust build setup at https://github.com/filmil/rust-bazel-example.
It is based on [cargo-raze](https://github.com/google/cargo-raze) (so, bazel-native rust build rules, not `cargo`-based).
IDE integration is based on a `rust-project.json` generator, also bazel-native. You can then use any LSP-compatible editor to get the IDE goodies.

The downside of using `rust-project.json` vs. cargo is that by default `rust-analyzer` will turn off any checks that it would otherwise run `cargo check` for if it gets the workspace info via `rust-project.json` instead of `Cargo.toml`.
So the IDE support is partial (e.g. will not alert you if you mistype a variable; it will not alert you on a type mismatch which is mildly annoying). But it works fairly well otherwise.

PR #775, opened 2023-09-18
Use the regex module for Unicode character class aware regular expr

dnicolodi:


dnicolodi:
This Unicode Technical Standard https://www.unicode.org/reports/tr55/#Identifiers seems to be relevant. But I'm still trying to wrap my head around it.

blais:
> This Unicode Technical Standard https://www.unicode.org/reports/tr55/#Identifiers seems to be relevant. But I'm still trying to wrap my head around it.

That's a full time job.

PR #780, opened 2023-10-02
parser: Allow single uppercase character flags without quote prefix

dnicolodi:


dnicolodi:
The different syntax between beancount v2 and v3 and having to prefix single uppercase character flags with the quote symbols annoyed it me enough to come up with a solution. This adds a rule to the lexer to match single uppercase characters and modifies the grammar to expect these at the right locations. The bulk of the changes are updates to the tests. If we agree that this is something we want, I'll update the C++ parser accordingly.

blais:
Thank you Daniele. I'll try to review this soon.

beancount/parser/lexer.l
```diff
@@ -230,9 +230,16 @@ NULL {
...
     return FLAG;
 }
-'[A-Z] {
-    yylval->character = yytext[1];
-    return FLAG;
+ /* The whitespace lookahead is not necessary to correctly parse valid
+  * ledgers, but it enables the lexer to throw an error sooner on
+  * invalid input. For example, without the lookahead, an invalid
+  * account name like "Assets:invalid" is tokenized into CAPITAL
+  * ('A'), KEY ('ssets:'), erorr ('invalid'). With the lookahead the
```
blais, line 237: tpyo

blais (approved):

PR #790, opened 2023-11-23
setup: Fix typo for extras_require

davide125:
This is a fixup for dc761f986c32bbb109cb719969ae55d6eb7247ae to make extras work correctly.

PR #791, opened 2023-11-25
Make file type logic work with both magic libraries

davide125:
The current implementation only works with python-magic because it uses its specific API. However, python-magic includes a [compatibility API](https://github.com/ahupp/python-magic/blob/master/COMPAT.md) as of 0.4.19, so we can use that instead to make the logic compatible with both python-magic and python-file-magic.

### DIFF
--- a/beancount/scripts/deps.py
+++ b/beancount/scripts/deps.py
@@ -111,8 +111,9 @@ def check_python_magic():
 
     python-magic is an interface to libmagic, which is used by the 'file' tool
     and UNIX to identify file types. Note that there are two Python wrappers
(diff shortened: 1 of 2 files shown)


davide125:
I didn't update the python-magic version in `setup.py` in here to avoid merge conflicts with https://github.com/beancount/beancount/pull/790, but also because we may want to move the magic requirement to an extra as it's an optional dependency.

dnicolodi:
I am  not sure I follow what the intent. beancount depends on `python-magic`. Making the API compatible with `python-file-magic` while still requiring `python-magic` does not seem to achieve anything.

PR #815, opened 2024-03-30
Fix implicit price date from cost

aldur:
I noticed that when transferring assets having a cost AND a date, the plugin would wrongly infer the transfer date as the price date, instead of the one specified in the cost.

In the following example, it would add a new price for 1 EUR dated 2023-02-14.

```beancount
2023-02-14 * "Move"
    Assets:A                              -50 B {1 EUR, 2021-05-23}
    Assets:C
```

If this is wrong, the fix is pretty simple.

### DIFF
--- a/beancount/plugins/implicit_prices.py
+++ b/beancount/plugins/implicit_prices.py
@@ -79,7 +79,7 @@ def add_implicit_prices(entries, unused_options_map):
                     # reducing leg?  Check.
                     meta = data.new_metadata(entry.meta["filename"], entry.meta["lineno"])
                     meta[METADATA_FIELD] = "from_cost"
-                    price_entry = data.Price(meta, entry.date,
+                    price_entry = data.Price(meta, cost.date,
                                              units.currency,
                                              amount.Amount(cost.number, cost.currency))
                 else:


wzyboy:
This is a valid use case.

Currently if you purchased some stocks/cryptocurrencies in 2020 at $200/share and move the lots in 2025, the `implicit_price` plugin generates a price directive of $200 in 2025. But the current market price for the stock/crypto might be $1000+, making the $200 price odd.

This patch would remove these odd directives.

But the example file you provided does not work. To make your patch actually takes effect, you need to either transfer the lots with dates manually attached to the destination lots, or use [a plugin](https://github.com/hoostus/beancount-asset-transfer-plugin).

What I mean by manually copy the lots, including the dates:

```
2023-02-14 * "Move"
  Assets:A     -50 B {1 EUR, 2021-05-23}
  Assets:C      50 B {1 EUR, 2021-05-23}
```

@blais What do you think?

wzyboy:
A less intrusive way would be to edit lot transfer txns to have `FLAG_TRANSFER` ([like this](https://github.com/hoostus/beancount-asset-transfer-plugin/blob/a627ca9abc2de27fa23d6c89d187e745ce057999/transfer.py#L111)) and skip these transactions in `implict_prices` plugin.

aldur:
> But the example file you provided does not work.

Right, I went back and looked at my ledger file and in my case it was working because I was also using a `transer_lots` plugin (I think taken from [here](https://groups.google.com/g/beancount/c/vDX1oA2mJXA), which seems like a less polished version of the plugin you linked).

Also TIL about transfer flag :)

blais:
Hmm, this seems like a bug indeed, but now every time that lot will show up the same price directive will be inserted. We will need to deduplicate them price directives somehow.

wzyboy:
I just tested and there are no duplicate entries with this patch.

Minimal file to test:

```beancount
plugin "beancount.plugins.auto_accounts"
plugin "beancount.plugins.implicit_prices"

2020-01-01 * "Buy BTC"
  Assets:Bank                               -18000.00 USD
  Assets:Wallet1                                2 BTC { 9000.00 USD }

2023-01-01 * "Move BTC to another wallet"
  Assets:Wallet1                               -2 BTC { 9000.00 USD, 2020-01-01 }
  Assets:Wallet2                                2 BTC { 9000.00 USD, 2020-01-01 }

2025-01-01 * "Sell BTC"
  Assets:Wallet2                               -2 BTC {} @ 100,000.00 USD
  Assets:Bank                             200,000.00 USD
  Income:PnL
```

Print out price entries:

```python
#!/usr/bin/env python

from beancount.loader import load_file

entries, _, _ = load_file('./gh_pr_815.bean')
for e in entries:
    if isinstance(e, data.Price):
        print(e)
```


Without patch:

```
Price(meta={'filename': '/home/wzyboy/Documents/Ledger/gh_pr_815.bean', 'lineno': 4, '__implicit_prices__': 'from_cost'}, date=datetime.date(2020, 1, 1), currency='BTC', amount=9000.00 USD)
Price(meta={'filename': '/home/wzyboy/Documents/Ledger/gh_pr_815.bean', 'lineno': 8, '__implicit_prices__': 'from_cost'}, date=datetime.date(2023, 1, 1), currency='BTC', amount=9000.00 USD)
Price(meta={'filename': '/home/wzyboy/Documents/Ledger/gh_pr_815.bean', 'lineno': 12, '__implicit_prices__': 'from_price'}, date=datetime.date(2025, 1, 1), currency='BTC', amount=100000.00 USD)
```

With patch:

```
Price(meta={'filename': '/home/wzyboy/Documents/Ledger/gh_pr_815.bean', 'lineno': 4, '__implicit_prices__': 'from_cost'}, date=datetime.date(2020, 1, 1), currency='BTC', amount=9000.00 USD)
Price(meta={'filename': '/home/wzyboy/Documents/Ledger/gh_pr_815.bean', 'lineno': 12, '__implicit_prices__': 'from_price'}, date=datetime.date(2025, 1, 1), currency='BTC', amount=100000.00 USD)
```

Works as expected. Because only [unique](https://github.com/wzyboy/beancount/blob/994c8a705aa724501c49a5c7ab2df9f6d7c6c456/beancount/plugins/implicit_prices.py#L92-L97) price directives are produced by the plugin.

blais:
I printed before and after on my file and I get diffs: 
image

I'll have to investigate deeper.

blais:
Please note: when testing, be wary of the cache.

wzyboy:
I just tested again with cache explicitly disabled (`export BEANCOUNT_DISABLE_LOAD_CACHE=True`). The test result is the same:

image

The test file has only 3 transactions, it should not trigger cache anyway.

Did I miss anything?

blais:
I tested using my own ledger. There are differences. I'll need to isolate and understand better the effect before merging.

PR #816, opened 2024-04-01
docs: fix module docstring

william-davies:
I think `In a sense, this is the inverse of "pedantic." This is useful when doing some types of quick and dirty tests.` was copied from the `beancount.plugins.auto` docstring


blais:
tyty!

PR #834, opened 2024-07-05
Bug fix: keep entries when doing beancount.ops.balance check

derpferd:
The check op was removing balance entries in the case where the account does not exist. I think it is better to keep the entry and it to fail validation, this makes it more consistent with other entry types.


blais:
Tests are failing

PR #838, opened 2024-07-17
Bug fix: keep entries when doing beancount.ops.balance check

derpferd:
The check op was removing balance entries in the case where the account does not exist. I think it is better to keep the entry and it to fail validation, this makes it more consistent with other entry types.


blais (approved):

blais:
Why do you need to change v2?

derpferd:
I am using fava and this is the version that fava uses. Is v3 stable?

blais:
Yes.

PR #840, opened 2024-08-12
Propagate cost basis through "transfer" transactions

ericaltendorf:
As discussed various times on the beancount mailing list, some users find it desirable that when a transaction "transfers" assets from one account to another, cost basis is brought along.  Concretely, something like

        2015-10-01 *
          Assets:Account1          1 HOOL {100.00 USD, 2010-01-01}
          Assets:Account1          1 HOOL {120.00 USD, 2012-01-01}

        2015-10-02 *
          Assets:Account1         -1 HOOL {}
          Assets:Account2          1 HOOL {}

booked under FIFO should result in "1 HOOL {100.00 USD, 2010-01-01}") in Account1.

This PR attempts to implement this.  This happens in two places: currency categorization, where we need to infer from account balances the cost currencies of the transfer postings, and in booking reductions, where we need to identify the reduced units and match them with the corresponding augmentations.

NOTE: This is for preliminary review, feedback, and testing only.  It is not ready to submit.  At the very least, the current implementation only works for transfers with two equal and opposite augment/reduce postings, but does not gracefully detect and reject more complex forms (e.g., a transfer from one account into two others, or from two others into one).


ericaltendorf:
Removing this PR as it requires more changes and discussion

PR #841, opened 2024-08-15
Generalize balance_by_account() with key fn

ericaltendorf:
Allow caller to pass in a function which can mark the point at which balances should be computed.  This was a TODO in the code, and I needed the functionality to compute balances at timestamps (which are meta flags on my transactions).  Happy to adjust naming and do more tests if you like.


dnicolodi:
Avoid merge commits in a PR. You should rebase not merge your patches on top of master.

This is an API change, thus it requires very good motivation to be merged. And I don't quite see the need for this extension. The same result can be obtained with code that looks like this
```python
import itertools
balance_by_account(itertools.takewhile(lambda x: not stop_fn(x), entries), date)
```
which simplifies a bit if `stop_fn()` gets its return value reversed. What am I missing?

blais:
Change is not breaking the API, seems fine to me. I think more importantly
is there are/were a few similar-ish functions around the codebase which
probably stand to be reviewed and reconciled into one.

dnicolodi:
> Change is not breaking the API, seems fine to me.

The patch modifies the signature of the function. How is this not an API change? I would agree with you if the patch added a keyword argument at the end of the arguments list, but here it is added in the middle. However, the main objection is that this addition does not generalize and the same issue can be solved with `itertools.takewhile()`.

The patch states that it solves a `TODO` comment which I believe means something completely different.

blais:
Sure, move the argument at the end of the function or make it keyword only.

It's true it doesn't fix the larger issue but that will involve a lot more
API changing things.

blais:
FWIW I don't expect contributors to refactor things and fix up my messes
(or should I say, my evolving APIs). OTOH an incremental contribution is
often an indicator of something that's needed by someone and that's
valuable input, as I never needed to do this myself. So the right thing
here is probably to allow the function to become slightly more complex now
and temporarily and add the stop_fn so when I do get to refactor things the
behavior will be included in a newer/better api (or removed altogether,
including removing the date, and force users to compose as you suggested).

dnicolodi:
I don't understand: you suggest to add this (IMO unnecessary) feature to then remove it again in the future as part of refactoring Does this benefit anyone? The user proposing adding the functionality will have to deal with the missing functionality till the next release of beancount (most likely in the way I suggest above) then they will need to modify their code to make use of the new functionality (if they want) but finally they will need to rework their code once more to adapt to the refactored code that we will be eventually release.

I am definitely in favor of adding functionality that enables new use cases, and do not block evolution waiting for redesigns, refactors, and cleanups that may never happen. But, in this case, the changes proposes AFAICT do not enable anything that is not already possible today simply composing two functions.

ericaltendorf:
There is more for me to do and say here, but for now just two quick things:

Thank you for the guidance on git usage and preparing PRs, I am new to contributing in this way and probably should have studied best practices before doing this.

I had thought that I was addressing the TODO, if I am not can you help me better understand the bigger issue the todo described?

Issue #332, opened 2018-09-09
Add date + time parsing to parser and grammar


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/332) by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

Parse ISO8601 date+time format, convert to datetime instance, and add to metadata.
No other support (perhaps add optional metadata to key sorting function).
See this thread:
https://groups.google.com/d/msgid/beancount/d6a65c2c-3ee2-4d69-b3e2-b5c94be62645%40googlegroups.com?utm_medium=email&utm_source=footer

PR #852, opened 2024-09-15
Sort entries by time field in metadata

hsiuhsiu:
This PR provides a temporary quick fix for the issue #332 [(More discussion)](https://groups.google.com/g/beancount/c/zZiFBPU1V8Y). It offers a simple way to add time support to transactions using metadata, addressing a specific scenario where the order of transactions matters across different files.

## Example Use Case
Consider the following two transactions in different .bean files:

First .bean file:
```beancount
2020-03-01 * "transfer"
  Assets:Crypto:Coinbase                   -1 ETH {164.06915 USD, 2019-04-13}
  Assets:Crypto:CoinbasePro                 1 ETH {164.06915 USD, 2019-04-13}
```
Second .bean file:
```beancount
2020-03-01 * "transfer"
  Assets:Crypto:CoinbasePro                -1 ETH {164.06915 USD, 2019-04-13}
  Assets:Crypto:CoinbasePro              1000 USD
  Income:Crypto
```
The first transaction needs to be processed before the second one. If these transactions are in the same file, the order is preserved, and no error occurs. However, when the transactions are in separate files, they are sorted by their line numbers within each file, which can cause processing errors.

## Proposed Solution
With this change, users can add a time field to the transaction metadata, allowing finer control over transaction order across multiple files. For example:

First .bean file:
```beancount
2020-03-01 * "transfer"
  time: "22:47:17"
  Assets:Crypto:Coinbase                   -1 ETH {164.06915 USD, 2019-04-13}
  Assets:Crypto:CoinbasePro                 1 ETH {164.06915 USD, 2019-04-13}
```
Second .bean file:
```beancount
2020-03-01 * "transfer"
  time: "22:57:08"
  Assets:Crypto:CoinbasePro                -1 ETH {164.06915 USD, 2019-04-13}
  Assets:Crypto:CoinbasePro              1000 USD
  Income:Crypto
```

## Timezone Handling (updated 2025 February)
Following the ISO 8601 standard, there are three types of time strings supported:
1. Explicit offset – e.g., `12:34:56+08:00`
2. Zulu time (UTC, denoted by Z) – e.g., `12:34:56Z`
3. Local time – The default timezone specified in the default_timezone option will be used – e.g., `12:34:56`

In the future, support for account-level or even entry/posting-level time and timezone settings may be added.

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -24,6 +24,7 @@ from beancount.core.number import D
 from beancount.core.position import Cost
 from beancount.core.position import CostSpec
 from beancount.utils.bisect_key import bisect_left_with_key
+from beancount.utils.date_utils import parse_time
 
 # Type declarations.
 Account = str
@@ -705,18 +706,32 @@ def get_entry(posting_or_entry: Directive | TxnPosting) -> Directive:
 SORT_ORDER = {Open: -2, Balance: -1, Document: 1, Close: 2}
 
 
-def entry_sortkey(entry: Directive) -> tuple[datetime.date, int, int]:
+DEFAULT_ENTRY_TIME = datetime.time(12, 0, 0, tzinfo=datetime.timezone.utc)
+
+def entry_sortkey(entry):
     """Sort-key for entries. We sort by date, except that checks
     should be placed in front of every list of entries of that same day,
-    in order to balance linearly.
+    in order to balance linearly. We also sort by time if the entry has
+    a time field in the metadata.
 
     Args:
       entry: An entry instance.
     Returns:
-      A tuple of (date, integer, integer), that forms the sort key for the
-      entry.
+      A tuple of (date, integer, datetime.time, integer), that forms the
+      sort key for the entry.
     """
-    return (entry.date, SORT_ORDER.get(type(entry), 0), entry.meta["lineno"])
+    time_value = DEFAULT_ENTRY_TIME
+    if "time" in entry.meta:
+        time_value = parse_time(entry.meta["time"])
+
+    # We sort by time after the type to ensure Open, Close, and Document
+    # directives appear in the right order.
+    return (
+        entry.date,
+        SORT_ORDER.get(type(entry), 0),
+        time_value,
+        entry.meta["lineno"],
+    )
 
 
 def sorted(entries: Directives) -> Directives:
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -24,7 +24,6 @@ from beancount.core.number import D
 from beancount.core.position import Cost
 from beancount.core.position import CostSpec
 from beancount.utils.bisect_key import bisect_left_with_key
-from beancount.utils.date_utils import parse_time
 
 # Type declarations.
 Account = str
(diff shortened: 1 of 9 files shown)


blais:
Hmm, tempting. 

- Implementation seems okay on the surface, but the format that allows time only ideally shouldn't allow timezone offset (-4:00) or timezone records (EDT), but only timezone identifiers (America/New_York), as these can be gotten wrong (maybe it wasn't EDT on that date, or maybe you adjust the date and then it's invalid). FYI see https://github.com/pola-rs/polars/issues/8289. Also, a user might be surprised that 17:00:00 and 17:10:00+2:00 don't sort correctly. Granted it would be their fault but still not ideal. I'm not sure how to enter these timezone ids in a more terse way though. Maybe each account could have a timezone associated with it? Or maybe a default timezone for the notebook as a whole (easier and more constrained and consistent)?

- Note: Allowing the timezone to be inferred from the local machine is fraught with peril: you travel and when you process your ledger it's all different. We don't want that (and I believe you're not including support for this here).

- I'm wondering if taking another approach and just allowing a generalized string key might not be the better thing to do here. WDYT?

@dnicolodi   WDYT?

hsiuhsiu:
Thanks for the comment!

1. I wasn't aware of that timezone identifiers are preferred. I can see the point. Yeah, I also think putting "America/New_York" will be too annoying and error-prone. Conceptually, I think linking account and timezones is fine, but I guess the implementation will be much more involved. If I understand correctly, that basically solving the "Posting vs. Settlement Dates" issue as well? since there will be multiple timezones involve in a single transaction.
2. Agree, many people move across timezone temporarily and permanently often.
3. Sorry, not following what does "generalized string key" mean here?

So far, I feel that, making everything timezone-agnostic (or equivalently having a single fixed timezone for the whole ledger) will be simpler and cleaner. And that will be an improvement without causing an issue anyway?

blais:
> Thanks for the comment!
> 
> 1. I wasn't aware of that timezone identifiers are preferred. I can see the point. Yeah, I also think putting "America/New_York" will be too annoying and error-prone. 

I agree, not too sure what to do. I think we should solve the problem so that in the common case where all the timezones are the same the user doesn't have to insert them, but may override them. So maybe a global default (a new option) and an optional inconvenient timezone id that one usually doesn't have to provide.

> Conceptually, I think linking account and timezones is fine, but I guess the implementation will be much more involved. If I understand correctly, that basically solving the "Posting vs. Settlement Dates" issue as well? since there will be multiple timezones involve in a single transaction.

No, I don't think these are related. Posting vs. Settlement dates is a problem that requires having a date for each posting--it's unrelated. (I'm not sure why you bring that up.)


> 2. Agree, many people move across timezone temporarily and permanently often.

I think more often is that you might be traveling and your local timezone will be different. It's happened to me frequently while on the road running unit tests that used to work consistently (when I ran them from NYC) and all of a sudden they break (when I'm running them from elsewhere).

> 3. Sorry, not following what does "generalized string key" mean here?

That would be taking a more naive approach and not even trying to interpret the time and date, e.g., just using a string and "que sera sera." I guess if we need timezones and stuff to sort things out properly we can't really have that option but at the same time I don't think the use cases require to solve that problem. Maybe just a secondary key containing any string is good enough...

> 
> So far, I feel that, making everything timezone-agnostic (or equivalently having a single fixed timezone for the whole ledger) will be simpler and cleaner. And that will be an improvement without causing an issue anyway?

hsiuhsiu:
> > Conceptually, I think linking account and timezones is fine, but I guess the implementation will be much more involved. If I understand correctly, that basically solving the "Posting vs. Settlement Dates" issue as well? since there will be multiple timezones involve in a single transaction.
> 
> No, I don't think these are related. Posting vs. Settlement dates is a problem that requires having a date for each posting--it's unrelated. (I'm not sure why you bring that up.)

Because if each account has its own timezone, then one transaction can involve multiple accounts with multiple timezones, then we either (1) need to tell the interpreter which timezone to use, or (2) specify different time strings for each account
 

Ah, I see what you meant. And I like it. Whatever date/time solution solves two issues (1) accurate event logging, (2) transaction ordering (cross files). We indeed don't have to solve them together -- (1) is less important, and people can always do that via plugin and and comments. For (2), we just need a secondary key to be used manually when necessary.

Do you think I can instead, introduce the field "sort_key" to the metadata?

trim21:
is it possible to extend current date to date[time]? 🤔

for example, `2020-03-01T00:01:02 * "transfer"`

blais:
This has been discussed before; it would drag timezones into it. I chose to keep things at the civil date level on purpose.
It's been discussed to add a time field without interpretation to the syntax and I agreed to this a while back;
never moved on it though.

I'm designing a new schema now, will consider adding it in.

trim21:
> This has been discussed before; it would drag timezones into it. I chose to keep things at the civil date level on purpose. It's been discussed to add a time field without interpretation to the syntax and I agreed to this a while back; never moved on it though.
> 
> I'm designing a new schema now, will consider adding it in.

A default tz option and a optional tz part after time may do?

If we have seprated parsing/transforming/booking steps this is a easy thing to do, even for file level tz option

hsiuhsiu:
> I'm designing a new schema now, will consider adding it in.

In that case, another thought is, can we have a "borrow" a commodity and have a negative position? At least temporary -- meaning only require the balance at the end of a day to be non-negative. Speaking of which, I personally never have a short position, but curious if allowing negative also makes that easier to track.

dnicolodi:
> In that case, another thought is, can we have a "borrow" a commodity and have a negative position? At least temporary -- meaning only require the balance at the end of a day to be non-negative.

A loan is modeled booking against a `Liability` account. I don't see the need to allow negative balances, but probably I'm missing something. Can you please describe in more details what you mean? But please open a new issue for that, as it has no connection with this PR.

> Speaking of which, I personally never have a short position, but curious if allowing negative also makes that easier to track.

Why do you thing that short positions hard to track now? How are negative balances simplifying anything?

hsiuhsiu:
> > In that case, another thought is, can we have a "borrow" a commodity and have a negative position? At least temporary -- meaning only require the balance at the end of a day to be non-negative.
> 
> A loan is modeled booking against a `Liability` account. I don't see the need to allow negative balances, but probably I'm missing something. Can you please describe in more details what you mean? But please open a new issue for that, as it has no connection with this PR.

In the example I gave at the very beginning, the point of introducing "time" is in order to sort the transaction "correctly".
And in my last message, I was hoping that, allowing negative balance will make either order acceptable. (now it will compile error as there's no slot for the CoinbasePro account to sell.

> > Speaking of which, I personally never have a short position, but curious if allowing negative also makes that easier to track.
> 
> Why do you thing that short positions hard to track now? How are negative balances simplifying anything?

Maybe not, I never have to track short positions and haven't thought about it seriously. So probably that's non-sense. But anyway, like in the previous paragraph mentioned, I'm thinking if allowing negative balance can also solve what this PR is trying to solve at the beginning, without worrying about the order.

SuibianP:
> This has been discussed before; it would drag timezones into it. I chose to keep things at the civil date level on purpose. It's been discussed to add a time field without interpretation to the syntax and I agreed to this a while back; never moved on it though.

For anyone looking for the context: <https://groups.google.com/g/beancount/c/zZiFBPU1V8Y/m/z4_jN340EwAJ>

This would be nice. As a newcomer planning to use Beancount as my *canonical* book, my main concern is more about the loss of time precision thus canonicality, for which I still have to keep the original documents around. Metadata certainly works, but does not feel natural when there is already the ISO8601 date field that can be extended to full ISO8601/RFC3339 while remaining extensible and unambiguous in intended behaviour.

I am frankly not yet familiar enough with Beancount to understand the issue with timezone from the discussion (thought the timestamps are transformed to a totally ordered representation and then sorted), but would it be possible to support parsing the subset without tz? That format would be backward compatible as well should the support lands some time later on.

wzyboy:
FWIW, I've been using [sort_by_time](https://gist.github.com/dallaslu/431d23f4174ed74874e803dc5eca7a3c) plugin for years and it meets my occanical needs of intra-day sorting.

hsiuhsiu:
@blais had some new thoughts about this. Also updated the description. Basically, I'm trying to
1. keep the actual time rather than a arbitrary sort keys
2. balance between ISO8601 and the timezone identifier

I think another potential future work could be supporting time format like `12:34:56[America/New_York]` But anyway, this is the direction I'm trying.

PR #860, opened 2024-11-03
Modernizations

dnicolodi:


Makefile
```diff
@@ -40,28 +29,16 @@ $(CROOT)/lexer.c $(CROOT)/lexer.h: $(CROOT)/lexer.l $(CROOT)/grammar.h
...
-	$(CROOT)/macros.h			\
-	$(CROOT)/tokens.h
-
 .PHONY: build
-build: $(SOURCES)
-	$(PYTHON) setup.py build_ext -i
-
-$(CROOT)/tokens_test: $(CROOT)/tokens_test.o $(CROOT)/tokens.o $(CROOT)/decimal.o
+build:
+	meson setup --reconfigure build/
```
blais: I'm all for modernization, but this does include a new dependency on meson though.
How reliable is that? Will it work on Windows? On macOS?

dnicolodi: Yes, it works on Windows and macOS. It is how the binary Python packages we publish are built and also how the package is build if someone installs it on a platform for which we don't publish a binary. I think it is more reliable than setuptools. The build system we use is the same used by numpy, scipy, pandas, matplotlib and others.

dnicolodi:
Converting this as a draft for now. The Makefile does not regenerate the parser from the bison and flex sources anymore when running `make build`, therefore working on the parser is more cumbersome now. I would remove the generated sources from version control and have them generated at build time. We publish binary packages for the major platforms, thus it should not be a problem for users.

blais (approved):

blais:
Another thing to modernize Beancount would be to make it use `uv`

dnicolodi:
Use `uv` for doing what? You can build and install Beancount via `uv` just like any other Python package.

dnicolodi:
@blais please have a look. I think this is ready to be merged now.

blais:
Changes look good.
I want to test those out on my machine later today before merging.

I'm not sure however that it's the best idea to require flex/bison just to build; I was checking in the C code on purpose to avoid those dependencies. We can try it, but if it prevents too many people from installing I'd consider bringing it back in the future.

More later.

dnicolodi:
I would be surprised to find that anyone installing Beancount is building from source, unless they want an unreleased version. We provide binary packages for all major platforms. The only platform that supports building from from source without too many problems is Linux and even there someone would need to install a C compiler and the Python development package. Adding flex and bison on top does not seem much of a problem to me. I don't seem to find a way to look at statistics of dowloaded source distributions vs binary distributions, but I think there must be a way to get at it.

blais:
Ok, this works on my box. Looks elegant as well.

blais:
One more thing: my logic for baking the revision into the source code is gone it seems.
This was useful when developing and using git commands to detect a bad version.
I'm referring to VC_CHANGESET and VC_TIMESTAMP.
Not super critical, but I wonder if it'd be easy to reinstall.
I guess with Windows it would be difficult to run git as an external program from meson.build.

dnicolodi:
Some of the logic is still there https://github.com/beancount/beancount/blob/af2ba01e71865a76d42bf70b0fe7519cd9a5f087/meson.build#L203-L214
however I did not test it. The one setting `VC_CHANGESET` and `VC_TIMESTAMP` is gone indeed. I don't remember why I didn't implement it. Most likely because it does not work when building from a source distribution and because the Bazel build also dropped support for it. I can restore it.

blais:
Nah that's alright, don't worry about it. If it gets annoying - e.g. we end up often using the wrong build - then we'll build it again. Let's leave the ifdefs for now they're harmless. 
tyty Daniele,

PR #861, opened 2024-11-06
[WIP]: rust parser

trim21:
- [X] includes
- [X] options
- [ ] directives
  - [ ]     Transaction
  - [ ]  Price
  - [ ] Balance
  - [ ]  Open
  - [ ]  Close
  - [ ]  Pad
  - [ ]  Commodity
  - [ ] Event


blais:
Thanks for this draft.
I gave this a very brief review for now, here is my first reaction:
- The pyi changes are something that should be done separately from this feature
- I have to review the reformatting changes, surprised ruff doesn't align the same, but also would be done in separate feature
- modernization of build, e.g. deletion of setup.py: dnicolodi has a more in-depth review of that in a different PR, I will review and likely accept, would conflict
- Rust parser: I think what I'd like to do is to keep the .proto files from the cpp feature and write a converter into the equivalent Rust (with no comments). Your Rust code and the LLM-generated code are good examples of the kind of output this program would generate. I think I'd like to make it a simple Python program that leverages the protobuf library and uses reflection on the protos to produce all the Rust code for all the protos.

trim21:
> The pyi changes are something that should be done separately from this feature

It's for convenient of developing, maybe we can move it to a seprated PR before this PR is ready.

> I have to review the reformatting changes, surprised ruff doesn't align the same, but also would be done in separate feature

That's because I didnt' run `ruff format` 😅

and `make lint` doesn't run `ruff format` eigher.

I run `ruff format` and looks like there are some files not formatted by ruff, and still got some reformatting changes.

> modernization of build, e.g. deletion of setup.py: dnicolodi has a more in-depth review of that in a different PR, I will review and likely accept, would conflict

Sorry, I didnt' get this. I think what bulding tool/how extensions are built is depend on which language they are, right? And the modernization of build is for c/c++ only.

So with rust there is no need for setuptools/cmake/meson/bazel, only cargo and maturin.


> Rust parser: I think what I'd like to do is to keep the .proto files from the cpp feature and write a converter into the equivalent Rust (with no comments). Your Rust code and the LLM-generated code are good examples of the kind of output this program would generate. I think I'd like to make it a simple Python program that leverages the protobuf library and uses reflection on the protos to produce all the Rust code for all the protos.

I don't thinks this is very doable. If we want to impl these types in rust and make them useable in python we will need `#[pyclass]` macro, and `#[pymethods]` for python's `__init__` method. I don't think proto can help with that.

dnicolodi:
> I don't thinks this is very doable. If we want to impl these types in rust and make them useable in python we will need `#[pyclass]` macro, and `#[pymethods]` for python's `__init__` method. I don't think proto can help with that.

What @blais is saying is that instead of writing the Rust code we should write a Python tool that generates the Rust code from the protobuf description of the data structures. This would use the protobuf library only to read the description.

trim21:
> > I don't thinks this is very doable. If we want to impl these types in rust and make them useable in python we will need `#[pyclass]` macro, and `#[pymethods]` for python's `__init__` method. I don't think proto can help with that.
> 
> What @blais is saying is that instead of writing the Rust code we should write a Python tool that generates the Rust code from the protobuf description of the data structures. This would use the protobuf library only to read the description.


For example, use some library like https://github.com/criccomini/proto-schema-parser and write a generator?

I prefer hand written structs by now and make this branch runnable first, maybe handle type defs in multiple branch later.

dnicolodi:
What is the advantage of rewriting the parser in Rust? Why it there the hope that this will improve the performance of Beancount? It may make parsing faster (although I don't think time spent parsing the input is the bottleneck or that the parsing can be made much faster) but all operations on the ledger will still be carried out in Python, with the additional penalty that any operation would need to convert from Rust types to Python types and vice versa. This has maybe low overhead for basic types (it still requires memory allocation though), but Beancount makes large use of `decimal.Decimal` and `datetime.date` which need to be constructed every time a field carrying one of these data types is accessed. I don't think that the speed up that may be offered from storing the fields in Rust containers rather than Python containers is going to be enough to offset the costs of the domain conversions.

trim21:
> What is the advantage of rewriting the parser in Rust? Why it there the hope that this will improve the performance of Beancount? It may make parsing faster (although I don't think time spent parsing the input is the bottleneck or that the parsing can be made much faster) but all operations on the ledger will still be carried out in Python, with the additional penalty that any operation would need to convert from Rust types to Python types and vice versa. This has maybe low overhead for basic types (it still requires memory allocation though), but Beancount makes large use of `decimal.Decimal` and `datetime.date` which need to be constructed every time a field carrying one of these data types is accessed. I don't think that the speed up that may be offered from storing the fields in Rust containers rather than Python containers is going to be enough to offset the costs of the domain conversions.

Personally I'm hoping we can rewrite core in rust and provide python api for current plugin. in that case we only need to convert python type to rust type at edge, not at whole parser.

But even using python conainers with c-api if faster than use them in python.

blais:
> > The pyi changes are something that should be done separately from this feature
> 
> It's for convenient of developing, maybe we can move it to a seprated PR before this PR is ready.
> 
> > I have to review the reformatting changes, surprised ruff doesn't align the same, but also would be done in separate feature
> 
> That's because I didnt' run `ruff format` 😅
> 
> and `make lint` doesn't run `ruff format` eigher.
> 
> I run `ruff format` and looks like there are some files not formatted by ruff, and still got some reformatting changes.

I accepted your other feature with ruff fmt. Thank you!


> 
> > modernization of build, e.g. deletion of setup.py: dnicolodi has a more in-depth review of that in a different PR, I will review and likely accept, would conflict
> 
> Sorry, I didnt' get this. I think what bulding tool/how extensions are built is depend on which language they are, right? And the modernization of build is for c/c++ only.

No I'm referring to the Python setup.py and such.
The C++ build is probably going to die.
I'm going to look at recycling all the work done there for a Rust parser instead (more below).

> 
> So with rust there is no need for setuptools/cmake/meson/bazel, only cargo and maturin.

Yes, but there would also be some Python wheels.
I'd like to try to use `uv` for that (by the same folks who are making ruff).
So there would be a toolchain for the Rust build, and a toolchain for the Python build.
I'm not too sure how to integrate them, Polars probably provides a good example.


> 
> > Rust parser: I think what I'd like to do is to keep the .proto files from the cpp feature and write a converter into the equivalent Rust (with no comments). Your Rust code and the LLM-generated code are good examples of the kind of output this program would generate. I think I'd like to make it a simple Python program that leverages the protobuf library and uses reflection on the protos to produce all the Rust code for all the protos.
> 
> I don't thinks this is very doable. If we want to impl these types in rust and make them useable in python we will need `#[pyclass]` macro, and `#[pymethods]` for python's `__init__` method. I don't think proto can help with that.
dnicolodi:
> 
> What @blais is saying is that instead of writing the Rust code we should write a Python tool that generates the Rust code from the protobuf description of the data structures. This would use the protobuf library only to read the description.

That's correct.
What protobufs provide is an ascii representation of the data structure that's nice for writing tests.
I already have a lot of tests written with it (see the tests in https://github.com/beancount/beancount/blob/cpp/beancount/cparser/parser_test.cc).

I think it might be possible, however, to rewrite these tests in Rust itself, but I'm not sure how convenient that is.
If we did that instead and it turns out looking nice, maybe I'd consider ditching the .proto files and turning them into source in Rust (and moving all the comments there), but even then, honestly I think generating all that code is probably going to be nicer. The Rust code probably requires a lot of annotations and the proto syntax is pretty minimal in comparison.  I'm not sure yet.

Yet another idea would be to do this (that is, convert the schema to Rust files with all comments there) but to implement a serde from/to an ascii format much like the protobuf ascii pb. It shouldn't be too hard to write.

A downside of protos is that they don't express all the things we want, in particular, civil dates (year/mth/day) and decimal numbers. It'd be nicer to just have those in native Rust in the schema.

In this feature I've been toying around the idea a bit:
https://github.com/beancount/beancount/tree/rust
- added all the schemas there
- had them converted by Claude (from A\)
- also used prost to see what that generates (still have to try to add PyO3)

In summary, ultimately my goal is to test
- benchmarking PyO3 to Python and back with mutability, ideally using data of size O(my large personal ledger)
- introduce some serde from protobuf ascii to leverage the new cpp parser tests (or rewrite them in Rust literal syntax with cmp)
If this works well (very fast?) this would be a good starting point for a Rust rewrite.
Writing a parser has been done by a number of others, I think we'd want to have an official parser in here.
I worked with jcornaz some time in the past to make sure his parser was comprehensive and could parse the whole syntax, this could be a starting point, but I'm not sure which parser generator libraries / framework I'd want to use (there are many).
(I know some years ago PEG parsers were all the rage, wondering if maybe those are in Rust as well)
In any case, I'd want all the tests preserved, I view tests as precious.

> 
> For example, use some library like https://github.com/criccomini/proto-schema-parser and write a generator?
> 
> I prefer hand written structs by now and make this branch runnable first, maybe handle type defs in multiple branch later.

You don't need an external library to parse protobuf schemas. 
The official protobuf package has its own parsing and reflection which is sufficient and should be what we use.

Once again I'm not attached to protobuf, but I like the idea of a very simple/minimal language for schema specification and autogeneration of the actual code to any language. This opens the door to supporting multiple languages eventually (which is really nice), and if we need to add lots of ugly annotations in Rust if it's autogenerated I don't care (if the Rust file is the pristine source, then it makes the schema ugly).


> What is the advantage of rewriting the parser in Rust? Why it there the hope that this will improve the performance of Beancount? 

Yes

> It may make parsing faster (although I don't think time spent parsing the input is the bottleneck or that the parsing can be made much faster) but all operations on the ledger will still be carried out in Python, with the additional penalty that any operation would need to convert from Rust types to Python types and vice versa. This has maybe low overhead for basic types (it still requires memory allocation though), but Beancount makes large use of `decimal.Decimal` and `datetime.date` which need to be constructed every time a field carrying one of these data types is accessed. I don't think that the speed up that may be offered from storing the fields in Rust containers rather than Python containers is going to be enough to offset the costs of the domain conversions.

No we would rewrite the entire codebase of the core beancount repo in Rust.
Python would be just at the API level, like in Polars.
Beancount would become a Rust project with a Python API that is cared for first class (because that's how users would interact with it and all other Beancount org projects would remain in Python). I can imagine most users writing plugins in Python, but the most savvy writing plugins in Rust (maybe we can figure out how to do that with a dlopen and registration mechanism).

Basically we would give Beancount a similar treatment as Pydantic or Polars.

You are correct that parsing is not the main slowness in processing ledgers--though it's not insignificant. Applying the booking rules is (last time I checked). This would all be done in Rust, and off of the improved schema from the cpp branch (that schema separates the intermediate AST and the finalized, booked and interpolated schema--they are distinct schemas, unlike in current Beancount. The types are more explicit).

The dream is for instant processing, and to build interactive interfaces on top of that. Imagine if while you're editing a transaction it was so fast that you could update another Emacs buffer with the output of bean-doctor context on every character, while you're typing. You could even do things like allow the user to select a particular Inventory lot from the calculated inventory just before the transaction, e.g., select lot by number, interactively. I would *love* instant bean-check. I think we could take Beancount to another level of usability. I'm currently waiting >5secs every time I process. It's not terrible, but it changes the way I interact with it. Ideally I'd want flymake to run bean-check in the background.

Also, I would like to make the API more capable, so that processing sequences of trades using our data structures (for matching lots) could be reusable outside the context of Beancount. Those data structures would all be implemented in Rust, but with a nice API. And it would implement the newer matching semantics I'm proposing in https://docs.google.com/document/d/1H0UDD1cKenraIMe40PbdMgnqJdeqI6yKv0og51mXk-0/


> 
> Personally I'm hoping we can rewrite core in rust and provide python api for current plugin. in that case we only need to convert python type to rust type at edge, not at whole parser.

Precisely.

trim21:
> No I'm referring to the Python setup.py and such.
> The C++ build is probably going to die.
> I'm going to look at recycling all the work done there for a Rust parser instead (more below).

maturin will be the tool to build python wheel. `pip wheel .`/`uv wheel .`/`build .` is also supported by maturin.

setup.py is also going to die in any way.

trim21:
so you plan to provide both rust crate and pypi wheel? 

maybe we can do more like pydantic, provide python api and write test with python 🤔

dnicolodi:
> Yes, but there would also be some Python wheels.
> I'd like to try to use `uv` for that (by the same folks who are making ruff).

`uv` has a Python package build front-end (`uv build`) that just invokes the Python package build back-end used by a specific package. In the case of Beancount that is now `meson-python` which itself is based on Meson. `uv` is a Python environment and Python package manager. You can use it to create an virtual environment and install Beancout:
```sh
$ uv venv --python=python3.13 beans
$ . beans/bin/activate
$ uv pip install beancount
```
Or, if you want to use `uv` project management capabilities:
```sh
$ uv init beans
$ cd beans
$ uv add beancount
```
But still you need to have a Beancount Python package.

trim21:
I didn't get the point on the discussion about packaging here. `setup.py` is used by `setuptools`, the old build tools we used which support c/c++. And now we are using meason, so setuptools and setup.py have already be removed.

If we use rust, then we will obviously use maturin (underhood it call cargo to build dll/so, like setuptools will call gcc/ld). and there is no need to add setup.py back.

uv/pip/build don't really care what build tool we are using to build wheel, as long as it support pep517.

blais:
> > No I'm referring to the Python setup.py and such.
> > The C++ build is probably going to die.
> > I'm going to look at recycling all the work done there for a Rust parser instead (more below).
> 
> maturin will be the tool to build python wheel. `pip wheel .`/`uv wheel .`/`build .` is also supported by maturin.
> 
> setup.py is also going to die in any way.

SGTM


> so you plan to provide both rust crate and pypi wheel?

I... guess?  Definitely a PyPI wheel; whether there's a Rust crate would depend on whether we want people to write Rust against the codebase. Maybe later?


> 
> maybe we can do more like pydantic, provide python api and write test with python 🤔

That's what I started to do for the C++ code. 
I agree it's a lot more elegant. There are a lot of tests. I'd rather not convert all to Rust for now.



> > Yes, but there would also be some Python wheels.
> > I'd like to try to use `uv` for that (by the same folks who are making ruff).
> 
> `uv` has a Python package build front-end (`uv build`) that just invokes the Python package build back-end used by a specific package. In the case of Beancount that is now `meson-python` which itself is based on Meson. `uv` is a Python environment and Python package manager. You can use it to create an virtual environment and install Beancout:
> 
> ```shell
> $ uv venv --python=python3.13 beans
> $ . beans/bin/activate
> $ uv pip install beancount
> ```
> 
> Or, if you want to use `uv` project management capabilities:
> 
> ```shell
> $ uv init beans
> $ cd beans
> $ uv add beancount
> ```
> 
> But still you need to have a Beancount Python package.

SGTM
I have little experience with `uv`. I prodded at it a bit and it felt very fast.
See https://www.youtube.com/watch?v=gSKTfG1GXYQ



> I didn't get the point on the discussion about packaging here. `setup.py` is used by `setuptools`, the old build tools we used which support c/c++. And now we are using meason, so setuptools and setup.py have already be removed.

I think that's right. We need meson for the C build.
That's not going away in master.
I see Rust as an experiment at the moment. 
If it were to supplant the C parser and core, we could get rid of the C code and a lot of Python code.
But that's for much, much later... and tentative.


> 
> If we use rust, then we will obviously use maturin (underhood it call cargo to build dll/so, like setuptools will call gcc/ld). and there is no need to add setup.py back.
> 
> uv/pip/build don't really care what build tool we are using to build wheel, as long as it support pep517.

SGTM

I admit I'm a tad worried about the large number of dependencies brought in by default but in the process of getting comfortable with Rust I'll avoid looking at that for a while.

blais:
Look here for some experiments: https://github.com/beancount/beancount/tree/rust/experiments/rust

trim21:
> I think that's right. We need meson for the C build.
> That's not going away in master.
> I see Rust as an experiment at the moment.
> If it were to supplant the C parser and core, we could get rid of the C code and a lot of Python code.
> But that's for much, much later... and tentative.

In this case I think I'll just go with hand written structs by now, we can adapter proto -> rust struct later.

I understand proto is more readable in defining struct, but I don't know it very well and I don't know how to make it work with pyo3. except LLM generated struct, and it's output is not stable 🤔

My initial plan is to replase `parse_file` (_parser and `grammar.Builder`) with rust

https://github.com/beancount/beancount/blob/064163aa429e2a043f743d1eba91fafd01d554d0/beancount/parser/parser.py#L182-L184

blais:
> > I think that's right. We need meson for the C build.
> > That's not going away in master.
> > I see Rust as an experiment at the moment.
> > If it were to supplant the C parser and core, we could get rid of the C code and a lot of Python code.
> > But that's for much, much later... and tentative.
> 
> In this case I think I'll just go with hand written structs by now, we can adapter proto -> rust struct later.
> 
> I understand proto is more readable in defining struct, but I don't know it very well and I don't know how to make it work with pyo3. except LLM generated struct, and it's output is not stable 🤔
> 
> My initial plan is to replase `parse_file` (_parser and `grammar.Builder`) with rust
> 
> https://github.com/beancount/beancount/blob/064163aa429e2a043f743d1eba91fafd01d554d0/beancount/parser/parser.py#L182-L184

Let's not have plans just yet. 
I'd like for us to do some benchmarking over a large file back and forth through Python to see if this is feasible in the first place.

Prototyping will inform what we end up with.

trim21:
> Let's not have plans just yet. I'd like for us to do some benchmarking over a large file back and forth through Python to see if this is feasible in the first place.
> 
> Prototyping will inform what we end up with.

Is it possible to have a shareable large beancount data files from generated fake data for testing/benchmarking?

I have a profiling branch on my local but my beancount is not large or complex enough, normally it only take only 500ms for bean-check.

dnicolodi:
You can generate one with `bean-example`

trim21:
I need reconsider this PR since we didn't plan to rewrite parser only, close it for now.

PR #870, opened 2024-11-14
fix parser test on windows and improve DX for windows users

trim21:
looks like we do not need to link libm manually in meson.build and also msvc can't find it on windows, msvc will link it by default, try to find it will fail.

File created by NamedTemporaryFile on Windows can't be shared, it can't be opened many times. use `TemporaryDirectory` and create a seprated file to bypass this.


dnicolodi:
Can you please don't reformat the code? It is very hard to look at the patch in the current form.

.github/workflows/tests.yaml
```diff
@@ -36,8 +36,13 @@ jobs:
       - run: echo "/opt/homebrew/opt/bison/bin:$PATH" >> $GITHUB_PATH
         if: runner.os == 'macOS'
 
-      - run: python -m pip install meson-python ninja pytest
+      - run: python -m pip install meson-python ninja pytest pytest-github-actions-annotate-failures
```
dnicolodi: I hate `pytest-github-actions-annotate-failures`. Please don't add it

trim21: any reason you hate it?

trim21: not try to argue about it, just curious

trim21:
> Can you please don't reformat the code? It is very hard to look at the patch in the current form.

looks like ruff format doesn't have stable output, I have black enabled in my IDE and these diff can't be reverted by ruff format. I'll revert them manually 🤔

meson.build
```diff
@@ -256,7 +256,7 @@ if get_option('tests').enabled()
         'beancount/parser/decimal.c',
         'beancount/parser/tokens.c',
         'beancount/parser/tokens_test.c',
-        dependencies: [libm, py.dependency(embed: true)],
+        dependencies: [py.dependency(embed: true)],
```
dnicolodi: Why this change?

`libm` is needed on some systems and Meson should do the right thing on systems where it is not.

beancount/parser/printer_test.py
```diff
@@ -230,7 +232,7 @@ def test_Transaction(self, entries, errors, __):
           Assets:Account1         1 USD @ 0 OTHER
           Assets:Account2         1 CAD @ 0 OTHER
 
-        2014-06-10 * "Entry with escaped \\"symbols\\" \\ \\r \\n"
+        2014-06-10 * "Entry with escaped \"symbols\" \ \r \n"
```
dnicolodi: Why this change?

beancount/parser/printer_test.py
```diff
@@ -89,15 +89,17 @@ def assertRoundTripViaRealFile(self, entries1, errors1, test_write_source=True):
...
 
         oss1.write('option "plugin_processing_mode" "raw"\n')
         printer.print_entries(entries1, file=oss1)
 
         oss1.close()
 
-        with open(oss1.name, "r") as file:
-            oss1_value = file.read()
+        with open(oss1.name, "rb") as file:
+            oss1_value = file.read().decode()
```
dnicolodi: Why this change? As far as I can tell the old and new code does exactly the same thing.

beancount/parser/printer_test.py
```diff
@@ -89,15 +89,17 @@ def assertRoundTripViaRealFile(self, entries1, errors1, test_write_source=True):
         # Contex manager can be used as of python 3.12
         # see https://github.com/python/cpython/issues/58451
         # https://github.com/beancount/beancount/issues/222
-        oss1 = tempfile.NamedTemporaryFile("w", suffix=".beancount", delete=False)
+        oss1 = tempfile.NamedTemporaryFile(
+            "w", suffix=".beancount", delete=False, newline="\n"
+        )
```
dnicolodi: I would use the native line ending for the platform. Instead of changing the newline style here, change the code where more lines are appended to use native newlines.

beancount/parser/printer.py
```diff
@@ -349,7 +350,11 @@ def Note(self, entry, oss):
         self.write_metadata(entry.meta, oss)
 
     def Document(self, entry, oss):
-        oss.write('{e.date} document {e.account} "{e.filename}"'.format(e=entry))
+        oss.write(
+            "{e.date} document {e.account} {filename}".format(
+                e=entry, filename=json.dumps(entry.filename, ensure_ascii=False)
+            )
+        )
```
dnicolodi: Why `json.dumps()`?

beancount/parser/parser_test.py
```diff
@@ -82,18 +83,18 @@ def test_parse_string(self):
...
 
     def test_parse_file(self):
-        with tempfile.TemporaryFile("w+b", suffix=".beancount") as file:
-            file.write(self.INPUT.encode("utf-8"))
-            file.seek(0)
-            entries, errors, _ = parser.parse_file(file)
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_file_path = Path(temp_dir, "temp.beancount")
+            temp_file_path.write_bytes(self.INPUT.encode("utf-8"))
+            entries, errors, _ = parser.parse_file(str(temp_file_path))
```
dnicolodi: This changes what the test does.

beancount/parser/printer_test.py
```diff
@@ -106,7 +108,7 @@ def assertRoundTripViaRealFile(self, entries1, errors1, test_write_source=True):
         self.assertEqualEntries(entries1, entries2)
         self.assertFalse(errors)
 
-        oss2 = io.StringIO()
+        oss2 = io.StringIO(newline="\n")
```
dnicolodi: Same here as in the other similar hunk.

dnicolodi (changes requested):
Some comments, but mostly questions.

beancount/parser/printer.py
trim21: for double quoted escaped string, there is no way(no easy way) to do this. `repr(s)` is single quoted

beancount/parser/parser_test.py
trim21: Oh, yes

beancount/parser/printer.py
dnicolodi: Doing what exactly? Escaping double quotes in `filename`? I don't think `filename` is expected to contain double quotes, but anyway: `filename.replace('"', '\\"')` does it.

trim21: I find a problem here, we can not use `\` as escaping because windows use it as filepath seprator, even `json.dumps` is also wrong 🤔

dnicolodi: Why does `filename` contain double quotes?

dnicolodi: Are double quotes a valid filename character on Windows?

beancount/parser/printer_test.py
trim21: we already use native newlines, add argument here is to get a platform-independent result in tests

dnicolodi: I'm not sure I understand. Where is the platform dependent value compared to a platform independent one?

beancount/parser/printer_test.py
dnicolodi: This is marked as resolved, but the change is still there.

beancount/parser/printer.py
trim21: good question, looks like it's not

beancount/parser/printer_test.py
trim21: `oss2` here is a `io.StringIO` and doesn't do any newline handling, do we need to let `tempfile.NamedTemporaryFile` do not change any newlines.

trim21: I find the reason, it's in escape_string

trim21: looks like this is expected? `\n` is not escaped.

https://github.com/beancount/beancount/blob/07406fd58de27deb8920663af85659e0e2c81eba/beancount/utils/misc_utils_test.py#L236-L240

blais:
Q1. Can you put the original reason for the problem in the description? This was broken because... NamedTemporaryFile doesn't work properly on Windows? (the file is locked?)

Q2. Is this ready to merge?

trim21:
file created by NamedTemporaryFile on Windows can't be shared, it can't be opened many times.

maybe, I may find a bug in parser, and I think we should fix it in another PR

```py
from beancount.loader import load_string

row = r"""
2014-06-10 * "Entry with escaped \"symbols\" \ \r \n"
  Assets:Account1       111.00 BEAN
  Assets:Cash          -111.00 BEAN

2014-06-20 custom "budget" Assets:Account2 "balance < 200.00 USD" 200.00 10.00 USD
"""

assert load_string(row)[0][0].narration == 'Entry with escaped "symbols"  \r \n'
```

`load_string(row)[0][0].narration == 'Entry with escaped "symbols"  \r \n'` should be false here

beancount/parser/printer_test.py
trim21: for better readability

dnicolodi:
> `load_string(row)[0][0].narration == 'Entry with escaped "symbols" \r \n'` should be false here

Why? I would expect it to work exactly as it does.

beancount/parser/printer_test.py
trim21: it's a bug of parser, can't be fixed in this PR

trim21:
> > `load_string(row)[0][0].narration == 'Entry with escaped "symbols" \r \n'` should be false here
> 
> Why? I would expect it to work exactly as it does.

Oh, you are right... I may need to get some sleep.

beancount/parser/printer_test.py
dnicolodi: Please drop this change. Readability is subjective. Especially in this case.: for me, having to scan several lines up to realize that this is a raw string and that the back-slash escaping is not interpreted by Python is much less readable than the double escapes.

trim21:
it's a bug in `print_entries`

trim21:
we parse `r"\n"` as `"\n"` but we didn't escape it back when printing. even we have a `test_escape_string` for that but it has wrong assert.

trim21:
> Q1. Can you put the original reason for the problem in the description? This was broken because... NamedTemporaryFile doesn't work properly on Windows? (the file is locked?)
> 
> Q2. Is this ready to merge?

when you merge would you mind merge it with squash? commit log here doesn't make much sence.

beancount/parser/printer.py
dnicolodi: Why the use of `json.dump()` is still there? Why was this comment marked as resolved when it is clearly not?

beancount/utils/misc_utils.py
```diff
@@ -281,15 +282,15 @@ def compute_unique_clean_ids(strings):
...
+def escape_string(string: str) -> str:
     """Escape quotes and backslashes in payee and narration.
 
     Args:
       string: Any string.
     Returns.
       The input string, with offending characters replaced.
     """
-    return string.replace("\\", r"\\").replace('"', r"\"")
+    return json.dumps(string, ensure_ascii=False)[1:-1]
```
dnicolodi: Why?

meson.build
```diff
@@ -251,6 +250,8 @@ py.extension_module(
 )
 
 if get_option('tests').enabled()
+    libm = cc.find_library('m')
+
```
dnicolodi: This is a change for the sake of change, and it has no relation with the focus of the patch. Please discard.

dnicolodi:
> when you merge would you mind merge it with squash? commit log here doesn't make much sence

Please fix this on your side and force push the resulting commit to your branch.

.github/workflows/tests.yaml
dnicolodi: Because it adds noise to the PR that does not help in the slightest to understand why a test failed.

beancount/parser/printer.py
trim21: sorry， forget to revert this

beancount/utils/misc_utils.py
trim21: > we parse `r"\n"` as `"\n"` but we didn't escape it back when printing. even we have a `test_escape_string` for that but it has wrong assert.

dnicolodi: I don't think there is a problem here. If there is a problem, how does the test work on platforms using `\n` as native newline? Anyhow, why use `json.dumps()`? I start to have the feeling that there is AI crap involved here...

dnicolodi: Please look at your patch: you change the implementation of `escape_string()` and then change the test to match your implementation. There test is correct for the current implementation. The current code implements the desired behavior. Why is this behavior not the correct one, in your opinion?

.gitignore
```diff
@@ -1,7 +1,7 @@
 *.DS_Store
 *.a
 *.picklecache
-*.pyc
+*.py[cd]
```
dnicolodi, line 4: This has nothing to do with the description of the PR.m Please revert or include this in a commits of its own with a suitable commit message.

beancount/utils/misc_utils.py
trim21: because I don't want to write double quoting escaping logic again.

And this is not a problem before because we didn't notice here, no one realy use `\\n` inside double quoted string.

`print_entries` doesn't revers paring as expected, try this example and see output from your console (different from before)

```
from beancount import print_entries
from beancount.loader import load_string

row = r"""
2014-06-10 * "Entry with escaped \"symbols\" \ \r \n"
  Assets:Account1       111.00 BEAN
  Assets:Cash          -111.00 BEAN
"""

print_entries(load_string(row)[0])
```

dnicolodi: What the function does is explained in the doc string? Did you read it? At a minimum, if you change what the function does you need to change the doc string as well.

.gitignore
trim21: python on windows use `pyd` as binary extension ext name, why it's not related in this PR?

beancount/parser/printer.py
dnicolodi: Why in hell are you changing this to escape double quotes? How the heck are you getting a filename that contains double quotes on Windows? I thought that we agreed that `"` is not a valid character in filenames on Windows. I'm getting tired of pointing out changes that have nothing to do with the stated purpose of the PR. Either you start to split up your changes in meaningful commits with commit messages explaining them, or I'll stop looking at the patches all together.

beancount/utils/misc_utils.py
dnicolodi: > because I don't want to write double quoting escaping logic again

What makes you think that the escape sequences accepted by the `json` module are the same accepted by the beancount parser?

.gitignore
dnicolodi: Because the title of the PR is "fix parser test on windows" and this does not fix any test.

meson.build
trim21: I't not a "change for the sake of change", it's a change doesn't affect unix user and affect windown users. we should not even try to find libm on windows.

.gitignore
trim21: this is easy to fix

beancount/utils/misc_utils.py
trim21: because it realy does, when json escaped in string `\"` `\r` `\n` `\b` `\f` are all handled with in parser, so escape it with `json.dumps` works.

unless it's a `"\你好"`， it get ignored by parser and parsed as `"你好"` then handled also as expected.

beancount/parser/printer.py
trim21: ok, i now why `json.dumps` here at the first place. becasue we parse filename wrong and still consider `\` as escaping char, so only `json.dumps` can revert that. 😅

since it's a bug of parser itself I'll just skip related tests on win32

meson.build
trim21: Also the reason it doesn't fail in ci because it use gcc which has libm, but it's not true for clang/clang-cl/msvc.

dnicolodi: Why moving looking up the `libm` within the `if get_option('tests').enabled()` section solves this problem?

trim21: yes it doens't, I have fixed this. see latest diff

dnicolodi: If this is required, it is a Meson regression. Meson has special handling of `libm` when compiling with MSVC. Can you please share verbose logs of the build failing?
```sh
python -m pip install -v .
```

trim21: I find the PR where meson has special handling of libm, which is somehow broken.

I have tried `meson setup build --wipe -Dbuildtype=release` with CC=clang, CC=clang-cl and `meson setup build --wipe -Dbuildtype=release --backend vs` and they all failed with `meson.build:6:10: ERROR: C shared or static library 'm' not found`, only working compiler is GCC on windows.

```text
Collecting meson>=1.2.1
    Obtaining dependency information for meson>=1.2.1 from https://files.pythonhosted.org/packages/76/73/3dc4edc855c9988ff05ea5590f5c7bda72b6e0d138b2ddc1fab92a1f242f/meson-1.6.0-py3-none-any.whl.metadata
    Using cached meson-1.6.0-py3-none-any.whl.metadata (1.8 kB)
  Collecting packaging>=19.0 (from meson-python>=0.14.0)
    Obtaining dependency information for packaging>=19.0 from https://files.pythonhosted.org/packages/88/ef/eb23f262cca3c0c4eb7ab1933c3b1f03d021f2c48f54763065b6f0e321be/packaging-24.2-py3-none-any.whl.metadata
    Using cached packaging-24.2-py3-none-any.whl.metadata (3.2 kB)
  Collecting pyproject-metadata>=0.7.1 (from meson-python>=0.14.0)
    Obtaining dependency information for pyproject-metadata>=0.7.1 from https://files.pythonhosted.org/packages/e8/61/9dd3e68d2b6aa40a5fc678662919be3c3a7bf22cba5a6b4437619b77e156/pyproject_metadata-0.9.0-py3-none-any.whl.metadata
    Using cached pyproject_metadata-0.9.0-py3-none-any.whl.metadata (6.9 kB)
  Using cached meson_python-0.17.1-py3-none-any.whl (27 kB)
  Using cached meson-1.6.0-py3-none-any.whl (973 kB)
  Using cached packaging-24.2-py3-none-any.whl (65 kB)
  Using cached pyproject_metadata-0.9.0-py3-none-any.whl (18 kB)
  Installing collected packages: packaging, meson, pyproject-metadata, meson-python
    Creating C:\Users\Trim21\AppData\Local\Temp\pip-build-env-3oiubf7d\overlay\Scripts
  Successfully installed meson-1.6.0 meson-python-0.17.1 packaging-24.2 pyproject-metadata-0.9.0
  Installing build dependencies ... done
  Running command Getting requirements to build wheel
  Getting requirements to build wheel ... done
  Running command Preparing metadata (pyproject.toml)
  + meson setup C:\Users\Trim21\proj\beancount C:\Users\Trim21\proj\beancount\.mesonpy-_3dwmx4z -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=C:\Users\Trim21\proj\beancount\.mesonpy-_3dwmx4z\meson-python-native-file.ini
  The Meson build system
  Version: 1.6.0
  Source dir: C:\Users\Trim21\proj\beancount
  Build dir: C:\Users\Trim21\proj\beancount\.mesonpy-_3dwmx4z
  Build type: native build
  WARNING: Project does not target a minimum version but uses feature introduced in '1.1': meson.options file. Use meson_options.txt instead
  Project name: beancount
  Project version: 3.1.0.dev0
  C compiler for the host machine: sccache clang (clang 19.1.3 "clang version 19.1.3")
  C linker for the host machine: clang link 14.41.34123.0
  Host machine cpu family: x86_64
  Host machine cpu: x86_64
  Program bison found: YES 3.8.2 3.8.2 (C:\Users\Trim21\scoop\shims\bison.EXE)
  Program flex found: YES 2.6.4 2.6.4 (C:\Users\Trim21\scoop\shims\flex.EXE)

  ..\meson.build:6:10: ERROR: C shared or static library 'm' not found
```

dnicolodi: Ah. `required: false` is missing. Just change `libm = cc.find_library('m')` to `libm = cc.find_library('m', required: false)`

trim21: yes, I already did that.

~~And I think nowadays even on posix libm is linked by default. I remove libm and compile it with gcc 12, everything is fine.~~

Ok I'm wrong about this, it's still need.

.gitignore
dnicolodi: How do you get a `.pyd` file in the source directory?

trim21: because I call meson manually

beancount/utils/misc_utils.py
```diff
@@ -281,15 +282,15 @@ def compute_unique_clean_ids(strings):
     return idmap
 
 
-def escape_string(string):
-    """Escape quotes and backslashes in payee and narration.
+def escape_string(string: str) -> str:
+    """Escape double quoted string, the returned str itself not quoted.
```
dnicolodi: This does not describe what the function does.

.gitignore
dnicolodi: I'm getting tired of your half answers. If you invoke meson manually you do not have build artifacts in the source directory. The only way to get a `.pyd` file in the source directory is if you copy it there.

What about you don't do that and instead use an editable install?
```
$ python -m pip install --no-build-isolation -e .
```
This is the recommended way to install a project for development.

beancount/utils/misc_utils.py
```diff
@@ -281,15 +282,16 @@ def compute_unique_clean_ids(strings):
     return idmap
 
 
-def escape_string(string):
-    """Escape quotes and backslashes in payee and narration.
+def escape_string(string: str) -> str:
+    """Escape string in double quoted style, the returned str itself not quoted.
+    for example: `escape_string("1\n\"2") -> "1\\n\"2"`
```
dnicolodi: This is not the style in which docstrings need to be written: the first line needs to be a self standing paragraph. Anyhow "double quoted style" is not a term with a recognizable meaning.
```suggestion
    """Escape backslashes, double quotes, and control characters."""
```

beancount/utils/misc_utils.py
```diff
@@ -281,15 +282,9 @@ def compute_unique_clean_ids(strings):
...
 
 
-def escape_string(string):
-    """Escape quotes and backslashes in payee and narration.
-
-    Args:
-      string: Any string.
-    Returns.
-      The input string, with offending characters replaced.
-    """
```
dnicolodi: Why are you removing most of the docstring?

trim21: it's from your suggestion https://github.com/beancount/beancount/pull/870#pullrequestreview-2438495445 

you add a full string.

.gitignore
trim21: thanks for your recommendation. but I still think we should add pyd, to match so in .gitignore file.

beancount/utils/misc_utils.py
dnicolodi: Sorry, my bad. I thought that it was evident that the suggestion applied only to the marked lines. Please fix this. Now I understand why most people find it tolerable to interact with LLMs... They apply the same level of critical thinking.

trim21: It's also evident that and docstring is end with `"""` and rest of docstring doesn't provide any extra information

dnicolodi: There is a reason if the docstring had that format. Please restore the truncated part.

dnicolodi:
We may be approaching something that can be merged.

There are at lest five things that the patch does:
1. adjust the tests so that they can run on windows,
2. modify the `escape_string()` function to escape more characters,
3. modify the serialization of the `filename` part of `document` directives in `print_entry()` to escape the string,
4. run the parser tests on Windows in the CI setup,
5. add `.pyd` to `.gitignore`.

These need to be 5 separate commits with proper commit messages.

1, 4, and 5 are uncontroversial. However, 2 and 3 are backward incompatible changes that need to be considered with more attention. In particular `escape_string()` is used to escape narration and description fields of transaction directives when they are serialized with `print_entry()`. Among other things, the changes proposed here would make it such that a transaction like this:
```beancount
2024-11-15 * "Foo" "Foo
bar baz"
    Assets:Foo  1.00 X
    Expenses:Foo
```
would be written back as:
```
2024-11-15 * "Foo" "Foo\nbar baz"
    Assets:Foo  1.00 X
    Expenses:Foo
```
I haven't thought it through completely, but another consequence of this may be that Beancount looses support for universal newlines in these fields: the serialization format depends on which system the code is run.

3 is probably the right thing to do. However, because Windows style file paths break in the current state of things, there may be code out there that works-around the current limitation. Once this change lands, that would result in double-escaping, and broken code.

@blais What is your opinion?

beancount/utils/misc_utils.py
```diff
@@ -281,15 +282,16 @@ def compute_unique_clean_ids(strings):
...
+def escape_string(string: str) -> str:
+    """Escape backslashes, double quotes, and control characters.
 
     Args:
       string: Any string.
     Returns.
       The input string, with offending characters replaced.
     """
-    return string.replace("\\", r"\\").replace('"', r"\"")
+
```
dnicolodi: This adds a spurious newline.

blais:
> I'm getting tired of your half answers.  [...]
and
> Why in hell are you changing this to escape double quotes? How the heck are you  [...]

@dnicolodi   I understand you're getting frustrated, but let's please keep the tone polite and encouraging. This language is crossing the line a bit for me and I'd like to discourage writing to our new contributors in this way.  Trim21 contributed an entire new parser in Rust, which will be useful in something that's coming up and I do appreciate the enthusiasm.  I do agree with the technical content of your reactions, but we should aim to bring new contributors in line with our conventions without detracting them from contributing. It's probably sufficient here to request disjoint changes be made in separate PRs.

> 1, 4, and 5 are uncontroversial. However, 2 and 3 are backward incompatible changes that need to be considered with more attention. In particular escape_string() is used to escape narration and description fields of transaction directives when they are serialized with print_entry(). Among other things, the changes proposed here would make it such that a transaction like this:

Agreed.

> @blais What is your opinion?

@trim21 Can you add a unit test for the Windows filename case that's breaking?  Is the issue that we need escaping of \ in our existing function, but that would break for \n and \r?

blais:
Just saw #871.
Okay, so after reviewing #871, I think given that Beancount escapes backslashes in strings, you'd have to type:
```
2014-06-08 document Assets:Account1 "D:\\a\\beancount\\beancount\\path\\to\\document.csv" 
```

It's not ideal, but it doesn't change the current semantics. In theory we could have unescaped strings for documents as a special case, I don't think it would be great to have different semantics in different contexts for strings. I think you should complete the refactoring and we submit changes for 1,4,5, so Windows tests works (thanks for submitting a fix, we don't run Windows much around here.)

For the future--and I think this applies not just for this project but also in the workplace--when you want to submit a change that does multiple things and needs review by others, it's always best to submit each conceptual change in isolation (i.e., using distinct PRs). It adds a bit of friction for sure, but you will find it helps reviewers and you will move changes through faster overall.

trim21:
> Can you add a unit test for the Windows filename case that's breaking? Is the issue that we need escaping of \ in our existing function, but that would break for \n and \r?

do you mean something like this?

```py

def test_win32_filename():
    row = r'2014-06-08 document Assets:Account1 "D:\a\beancount\beancount\path\to\document.csv"'

    file = io.StringIO()

    print_entries(load_string(row)[0], file=file)
    assert (
        load_string(row)[0]
        != r'2014-06-08 document Assets:Account1 "D:\\a\beancount\beancountpath\todocument.csv"'
    )

```

blais:
> > Can you add a unit test for the Windows filename case that's breaking? Is the issue that we need escaping of \ in our existing function, but that would break for \n and \r?
> 
> do you mean somethink like this?
> 

Yes. Actually given the comments I just made, it would be more useful to escape the backslashes; a future person reading the code will see your unit test and understand they're suppose to escape the backslashes in filenames.

(Once again though, doesn't Windows accept fwd slash as valid filename separators?)

trim21:
I'm not suggesting support single backslash if it's intend to be doule backslash escaped.

tests are actually using with `/`, but we use `os.path` to handle them and they will become `\`.

trim21:
> For the future--and I think this applies not just for this project but also in the workplace--when you want to submit a change that does multiple things and needs review by others, it's always best to submit each conceptual change in isolation (i.e., using distinct PRs). It adds a bit of friction for sure, but you will find it helps reviewers and you will move changes through faster overall.

I understand, the initial plan for this PR is to fix tests only, change the formatter is not my intention.

But with @dnicolodi 's suggestion I find that there are some tests failed because the bug of formatting and can't be fixed in tests only if we want to run them on windows. (also got confused in the process thought that it's a bug of parser)

beancount/parser/parser_test.py
```diff
@@ -12,9 +12,13 @@
 import sys
 from pathlib import Path
 
+import pytest
```
dnicolodi: While we use pytest to run the tests in CI, the Beancount test suite is kept compatible with unittest from the standard library. Please adjust the test accordingly.

dnicolodi (changes requested):

beancount/parser/parser_test.py
```diff
@@ -67,6 +69,19 @@ def test_parse_doc__noerrors(self, _, __, ___):
           Assets:US:Cash             -100 USD
         """
 
+    @unittest.skipIf(sys.platform != "win32", reason="test for win32 file separator")
```
dnicolodi, line 70: Why do you skip this test when executed on platforms other than Windows? As far as I know, the parser does not care if the path in the `document` directive is valid or not.

beancount/parser/parser_test.py
```diff
@@ -67,6 +69,19 @@ def test_parse_doc__noerrors(self, _, __, ___):
...
+    def test_win32_filename(self):
+        """filepath in document are expected to be escaped"""
+        row = r'2014-06-08 document Assets:Account1 "D:\a\beancount\beancount\path\to\document.csv"'
+
+        file = io.StringIO()
+
+        print_entries(load_string(row)[0], file=file)
+        assert (
+            load_string(row)[0]
+            != r'2014-06-08 document Assets:Account1 "D:\\a\beancount\beancountpath\todocument.csv"'
```
dnicolodi: Can you please explain in prose what this test is trying to test?

Unless I am missing something, the test as written now checks that the result of the parse, print, parse round trip gives you a string that is different from the string you started from, which is true in all cases, unless it exists a bug in the parser or in the printer that precisely maps the string you start form to a different string.

beancount/parser/parser_test.py
```diff
@@ -67,6 +69,19 @@ def test_parse_doc__noerrors(self, _, __, ___):
...
 
+    @unittest.skipIf(sys.platform != "win32", reason="test for win32 file separator")
+    def test_win32_filename(self):
+        """filepath in document are expected to be escaped"""
+        row = r'2014-06-08 document Assets:Account1 "D:\a\beancount\beancount\path\to\document.csv"'
+
+        file = io.StringIO()
+
+        print_entries(load_string(row)[0], file=file)
+        assert (
```
dnicolodi: This would need to be `self.assertNotEqual()` (yes, I know that there are a couple of tests in the codebase that use plain `assert`s, these would need to be fixed too...). But I don't see why you don't compare to what you expect to get, instead than comparing to an arbitrary different result.

dnicolodi (changes requested):

beancount/parser/parser_test.py
trim21: Our grammer care, parser doesn't work alone.
 https://github.com/beancount/beancount/blob/2a455c7987c8ae81bba761f5da3d74e4711d9e8e/beancount/parser/grammar.py#L773-L776

beancount/parser/parser_test.py
trim21: my bad, should not run printer here

trim21: is it clear what we are testing now?

beancount/parser/parser_test.py
dnicolodi: Thanks. I didn't remember about that.

beancount/parser/parser_test.py
dnicolodi: Now it is clearer. But I don't understand why you think that adding this test is appropriate. The test gives the same result before and after your changes. Do you think it helps extend the test coverage of the parser? Why is this case more important to test than others? Handling of escape sequences is done in the lexer, for which the `filename` field of the `document` directive is just a string like another. I think we agree on the fact that we don't want to change the parser to handle the `filename` field differently than other strings.

trim21: did I misunderstand what @blais means in https://github.com/beancount/beancount/pull/870#issuecomment-2479112420 ?

Or you disagree with him ?

trim21:
👀

.gitignore
dnicolodi: This hunk needs to be its own commit with a suitable commit message.

beancount/parser/printer.py
```diff
@@ -349,7 +349,11 @@ def Note(self, entry, oss):
         self.write_metadata(entry.meta, oss)
 
     def Document(self, entry, oss):
-        oss.write('{e.date} document {e.account} "{e.filename}"'.format(e=entry))
+        oss.write(
+            '{e.date} document {e.account} "{filename}"'.format(
+                e=entry, filename=misc_utils.escape_string(entry.filename)
+            )
+        )
```
dnicolodi: This introduces a backward incompatible change to the serialization code. At the very minimum it needs to be in a separate PR all together, where the merits of the change can be discussed. This change is not required to fix the tests. See #876

beancount/utils/misc_utils.py
```diff
@@ -281,15 +282,15 @@ def compute_unique_clean_ids(strings):
...
+def escape_string(string: str) -> str:
+    """Escape backslashes, double quotes, and control characters.
 
     Args:
       string: Any string.
     Returns.
       The input string, with offending characters replaced.
     """
-    return string.replace("\\", r"\\").replace('"', r"\"")
+    return json.dumps(string, ensure_ascii=False)[1:-1]
```
dnicolodi: This introduces a backward incompatible change to the serialization code. At the very minimum it needs to be in a separate PR all together, where the merits of the change can be discussed. This change is not required to fix the tests. See #876.

beancount/utils/test_utils.py
```diff
@@ -187,11 +188,11 @@ def new_function(self):
...
         if any(key not in allowed for key in kwargs):
             raise ValueError("Invalid kwarg to docfile_extra")
-        with tempfile.NamedTemporaryFile("w", **kwargs) as file:
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_file_path = Path(temp_dir, "temp.txt")
             text = contents or function.__doc__
-            file.write(textwrap.dedent(text))
-            file.flush()
-            return function(self, file.name)
+            temp_file_path.write_text(textwrap.dedent(text), encoding="utf-8")
```
dnicolodi: This now silently ignores the `buffering`, `encoding`, `newline`, `dir`, `prefix`, and `suffix` arguments that can be passed to the `docfile_extra()` decorator. If these are not used anywhere in the code, support for them can most likely be removed, but ignoring them silently is not ok.

dnicolodi (changes requested):

beancount/utils/test_utils.py
trim21: is it ok now?

beancount/parser/printer.py
trim21: 🤔 #876 is basically just skip testing failed case, right?

beancount/utils/misc_utils.py
```diff
@@ -282,15 +281,15 @@ def compute_unique_clean_ids(strings):
     return idmap
 
 
-def escape_string(string: str) -> str:
-    """Escape backslashes, double quotes, and control characters.
+def escape_string(string):
+    """Escape quotes and backslashes in payee and narration.
```
dnicolodi: I don't get this docstring change. Where is this function checking that it is being applied to payee and narration?

dnicolodi (changes requested):

beancount/utils/misc_utils.py
trim21: this is a revert, this is the old docstring.

trim21:
backward incompatible change has been removed

beancount/parser/parser_test.py
dnicolodi: I cannot know what @blais was asking for. My guess is that he was asking for a test that verifies that the entry printer does not escspe backslashes in the input strings. This test may do that, but it does so in the most counter intuitive way I can imagine and it actually checks that non-escaped backslashes produce unwanted results.

If we want to add a test like this (whch IMHO is not really necessary) I would write it in a very different way: just instantiate a `beancount.core.data.Document` object wit escaped backslashes in the file path, serialize it with `beancount.parser.printer.print_entry()` and check that the serialized file path has exactly the same form. If you craft the file path in the proper way, the test does not need to be platform dependent.

This test would need to live with the tests for the printer, though. Not with the tests for the parser.

beancount/parser/printer.py
dnicolodi: Wrong. The test is failing because it assumes something that is not true: that file paths do not contain characters that need to be escaped to be serialized as Beancount strings. This is not true in general on Windows. #876 constructs the test input data in a way that satisfies the assumption across platforms.

trim21: This is just another way of saying "skip testing failed case"

dnicolodi: No it is not. "skip testing failed case" assumes that there is something wrong with the code being tested. In this case, as I explained, there is something wrong with the test. What it is done is to fix the test to test the right thing. If you still don't see the difference, I don't know how else to explain it.

dnicolodi:
Changes should be split into commits that perform one single thing and each commit needs to come with a commit message that explain why a change is being made. Here we still have commits, fixup commits, reverts, and no meaningful commit message. I don't know how else I should ask for it and I don't like you ignoring this request.

beancount/parser/printer.py
trim21: Ok, now I can understand why you think they are different.

The assumption that file paths / narration do not contain characters that need to be escaped is simply wrong, and obviously not "the right thing".

This assumption not only exists in testing code but also implement code.

If this assumption is right, then we should not need to escape backward slashes in windows filepath（https://github.com/beancount/beancount/issues/871#issuecomment-2478455204 ）, right?

---

What is done in #876 is to replace rel path with abs path, so `os.path.join` in grammer won't be called. In this case filepath will only contain `/`.


By removing `\r` `\n` and `\` from testing, https://github.com/beancount/beancount/pull/876 the tests are fixed, but failed 
 case like `print_entries(load_string(r'2014-06-08 document Assets:Account1 "./document.csv"')[0])` and `2014-06-10 * "Entry with escaped \"quotes\" \r \n"` is no longer tested.

trim21: "there is something wrong with the code being tested"

This is obvirously true, for a beancount like this on windows:

```beancount
row = r'2014-06-08 document Assets:Account1 "./document.csv"'
```

printer will output something like this:

```beancount
2014-06-08 document Assets:Account1 "C:\Users\Trim21\proj\beancount\document.csv"
```

trim21: off course there is something wrong with the code being tested

did you try this on windwos:

```beancount
from beancount import print_entries
from beancount.loader import load_string

row = r'2014-06-08 document Assets:Account1 "./document.csv"'

print(load_string(row)[0])

print_entries(load_string(row)[0])
```

what should be expected output is currently output `2014-06-08 document Assets:Account1 "C:\Users\Trim21\proj\beancount\document.csv"` right?

trim21:
sorry for that, I'll squash the commit and force push them in following changes.

dnicolodi:
> sorry for that, I'll squash the commit and force push them in following changes.

We seem to have a communication problem. I thought that this

> 👀

was a weird way to ask for another review of this PR. This would have implied that you think that the PR is ready to be merged. What was that emoji supposed to mean?

beancount/parser/printer.py
dnicolodi: > The assumption that file paths / narration do not contain characters that need to be escaped is simply wrong

The existing code expects that characters that need escaping are escaped in the data passed to `beancount.parser.printer.print_entry()`. This is how it works. This is how it was designed. It cannot be changed. As I have explained multiple times, this would be an incompatible change that would break existing code.

As I have written multiple times, if you want to change this, please propose this change in a new PR where the merits of breaking backward compatibility can be discussed. It has nothing to do with fixing the tests to run on Windows.

> By removing `\r` `\n` and `\` from testing, #876 the tests are fixed, but failed
> case like `print_entries(load_string(r'2014-06-08 document Assets:Account1 "./document.csv"')[0])` and `2014-06-10 * "Entry with escaped \"quotes\" \r \n"` is no longer tested.

Of course removing the newline and form feed characters from the test string they are not tested anymore. I suspect that you have not read the commit message of the commit that changes the test. Have you? What is not clear in it? Do you understand why the test currently succeeds on Unix platforms but fails on Windows?

trim21: This is really funny. Before you continue this discussion or write something about how it expects, I would suggest you to read the code we are talking about first.

https://github.com/beancount/beancount/blob/2a455c7987c8ae81bba761f5da3d74e4711d9e8e/beancount/parser/printer.py#L432-L436

It currently doesn't escape filename doesn't mean it "expects that characters that need escaping are escaped." There is a third option: forget to handle, which is a common case in multiple-platform code.

These lines live here for 10 years, unless "format an entry into a string in the same input syntax the parser accepts" mean something else in your language, it expects `entries = load_string('...')[0])` get same entries with `load_string(print_entries(entries)`, and that's how we are using it in the same codebase.

It's designed to generate output that parser will parse it as same result. If a user really input a entry with escaped string, then it should output double escaped string so the parser could escape back, that's what we called "same input syntax the parser accepts"

I very much agree that change it is a breaking change and need discussion. But first stop make assumption that it's expected to be broken on win32 or "Escaping other special characters is responsibility of the user".

This is how it works but it's not how it's designed. 

If you do have critial thinking, please make a difference between the status quo and the intention. You are mixing how it designed with how it works or how it's implemented.

> As I have written multiple times, if you want to change this, please propose this change in a new PR where the merits of breaking backward compatibility can be discussed

> Do you understand why the test currently succeeds on Unix platforms but fails on Windows?

No one is talking about changing it in this PR anymore.

Right now we are talking about how to fix tests, and that you did in #876 is just "skip testing failed case", even yourself know the failed case only exists on win32 with relative file path.

trim21:
> > sorry for that, I'll squash the commit and force push them in following changes.
> 
> We seem to have a communication problem. I thought that this
> 
> > 👀
> 
> was a weird way to ask for another review of this PR. This would have implied that you think that the PR is ready to be merged. What was that emoji supposed to mean?

That's just asking for another review of this PR.

beancount/parser/parser_test.py
trim21: not really, I think he mean that a test show that filepath in document are expected to be escaped. printer is not involved here.

meson.build
```diff
@@ -1,7 +1,7 @@
 project('beancount', 'c', version: files('beancount/VERSION'), meson_version: '>=1.1')
 
-bison = find_program('bison', 'win_bison', version: '>=3.8.0')
-flex = find_program('flex', 'win_flex', version: '>=2.6.4')
+bison = find_program('bison', version: '>=3.8.0')
+flex = find_program('flex', version: '>=2.6.4')
```
dnicolodi: What is this?

trim21: squash based on wrong commit, reverted.

dnicolodi:
I asked several time to produce commits with individual changes and appropriate commit messages. It seems that either you do not care or you do not understand what this means. Which one is it? Please see #876 for an example of what I am asking.

We are now at 113 messages exchanged for a 53 lines patch. I'll not be looking at this PR anymore. The tests you have been trying to fix are fixed in an IMO better way in #876.

trim21:
For what you write in  #876, my opinion is  https://github.com/beancount/beancount/pull/870#discussion_r1860197083


> I asked several time to produce commits with individual changes and appropriate commit messages. It seems that either you do not care or you do not understand what this means. Which one is it? Please see https://github.com/beancount/beancount/pull/876 for an example of what I am asking.

because of https://github.com/beancount/beancount/pull/870#issuecomment-2478123132

PR #876, opened 2024-11-26
Fix parser tests on Windows

dnicolodi:


blais:
There's really no way to encapsulate this different in a newer, better context manager?
This is pretty ugly. I can't imagine I'd remember when writing a new test to use these.

beancount/utils/misc_utils_test.py
```diff
@@ -234,10 +234,8 @@ def test_is_sorted(self):
         self.assertTrue(misc_utils.is_sorted([7, 6, 3, 1], cmp=operator.gt))
 
     def test_escape_string(self):
-        assert (
-            misc_utils.escape_string('Entry with escaped "symbols" \\ \r \n')
-            == 'Entry with escaped \\"symbols\\" \\\\ \r \n'
-        )
+        escaped = misc_utils.escape_string('Entry with escaped "symbols" \\')
```
blais: Why remote \r and \n from this test?

dnicolodi: Because they are not escaped anyway and their presence seems to trip up readers of this code. I'll put them back and add a comment instead.

dnicolodi:
> There's really no way to encapsulate this different in a newer, better context manager?
> This is pretty ugly. I can't imagine I'd remember when writing a new test to use these.

I think the code looks much uglier than what it is because how ruff-format forces it to be indented. It is definitely possible to hide some of the complexity in a context manager, but then you need to remember to use it instead of the `tempfiles` context manager... The real solution would to use `io.BytesIO` instead that temporary files for all but an handful of tests, but that would likely require a bit more refactoring of the tests. I haven't tried, though. I'll add the context manager to hide away the complexity.

dnicolodi:
Instead of turning that pattern for the use of temporary files into something nice with a context manager I opted for using the context manager introduced by @trim21 that creates a file in a temporary directory (Thanks @trim21!). It is a bit more wasteful as the directory creation is not strictly necessary and it requires a few extra syscalls, but this is test code that is definitely not performance critical.

dnicolodi:
@blais I addressed your comments. Please have a look.

beancount/parser/printer_test.py
```diff
@@ -285,17 +285,36 @@ def test_Note(self, entries, errors, __):
...
+        # However ``beancount.parser.print_entries()`` does not do that,
+        # breaking the round-trip test.  To change ``print_entries()`` to escape
+        # special characters would be an incompatible change.
+        #
+        # The solution could be to use a simple file name in the test, however,
+        # the beancount loader transforms all relative paths into absolute
+        # paths, introducing path separator characters.  Windows also accepts
+        # forward slashes as path separators, therefore, the solution is to
+        # craft the test to use absolute paths with forward slashes on all
+        # platforms.
```
blais, line 305: Amazing description! Thank you!

PR #877, opened 2024-11-27
replace `chdir` with thread-safe root_dir aware glob search  

trim21:
python 3.10 add `glob.glob(root_dir=...)`, before we drop py<3.10 we need to build abs path from root.

We should delay this after we can run loader_test on win32 to see if it works on win32 too

### DIFF
--- a/beancount/utils/file_utils.py
+++ b/beancount/utils/file_utils.py
@@ -4,7 +4,6 @@ __copyright__ = "Copyright (C) 2014-2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
 from os import path
-import contextlib
 import logging
 import os
 import time
@@ -99,20 +98,3 @@ def touch_file(filename, *otherfiles):
         new_stat = os.stat(filename)
         if new_stat.st_mtime_ns > orig_mtime_ns:
             break
-
-
-@contextlib.contextmanager
-def chdir(directory):
-    """Temporarily chdir to the given directory.
-
-    Args:
-      directory: The directory to switch do.
-    Returns:
-      A context manager which restores the cwd after running.
-    """
-    cwd = os.getcwd()
-    os.chdir(directory)
-    try:
-        yield cwd
-    finally:
-        os.chdir(cwd)
--- a/beancount/utils/file_utils.py
+++ b/beancount/utils/file_utils.py
@@ -4,9 +4,12 @@ __copyright__ = "Copyright (C) 2014-2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
 from os import path
+import contextlib
 import logging
 import os
 import time
+
+import typing_extensions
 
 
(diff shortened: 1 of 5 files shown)


blais:
Your tests pass on windows. Do you still want to wait?

blais (approved):
SGTM, I don't like changing the global state either.

trim21:
> Your tests pass on windows. Do you still want to wait?

Ok, then there is no need to wait.

beancount/utils/file_utils.py
```diff
@@ -101,6 +103,7 @@ def touch_file(filename, *otherfiles):
             break
 
 
+@typing_extensions.deprecated("use contextlib.chdir or community-backport instead")
 @contextlib.contextmanager
 def chdir(directory):
```
blais: Why not just remove chdir? If we don't use it elsewhere in this repo, remove.

blais: I don't see it used anywhere else in the beancount repo. I think it's safe to remove.

trim21: > If we don't use it elsewhere in this repo, remove.

I did remove it in 5e288d5a97e4072164a67a75b785cb595f7ffc5b, then I found someone is using it so I add it back.

https://github.com/alexjj/money-scripts/blob/ddaee26d61092bee1db84e9e9034a20a39309a1b/bean-bake.py#L216

I have a general questions about this situation. I find there are severial things in thie repo we don't use. For example, `date_utils.intimezone` and `snoop.Snoop` and I didn't find it used in other public repos. Is is OK we just remove them in the future?

blais: If it's used in another repo like this + we're on the master branch, we can remove it here. The beancount repo should contain only the minimum  it needs. That other repo can switch to using contextlib.chdir.

To your second question, yes, we should probably remove old things that aren't public and aren't in use anymore, but it's not super important. Those exist because all the bean* repos used to be in here, and they were forked to other repositories under github.com/beancount. Now, it might be worthwhile to move that code to those repos if it's still used there, except maybe if they're used broadly in a bunch of repos and related to beancount functionality, in which case we'd consider them publicly provided by this repo.  

snoop: This is useful for regexps but don't use it anymore (it was used by bean-bake iirc). We can remove.

intimezone: Can also remove if it's not used. This was used for some tests (surprised we're not using it anymore).

blais: Also, it's best to do these removals in dedicated PRs each, so we have a history of them.

trim21: ok, done

trim21: > intimezone: Can also remove if it's not used. This was used for some tests (surprised we're not using it anymore).

I guess the reason is it can't work on windows

blais (approved):

blais (approved):

PR #878, opened 2024-11-27
fix beancount/{utils,core,ops,projects,tools} tests on win32

trim21:


beancount/utils/test_utils.py
```diff
@@ -78,6 +80,18 @@ def subprocess_env():
     return {"PATH": binpath, "PYTHONPATH": find_python_lib()}
 
 
+@contextlib.contextmanager
+def temp_file(suffix: str = ".txt") -> Generator[Path, None, None]:
+    """A context manager that return a filepath inside inside a temporary directory and
+    deletes this directory unconditionally once done.
+
```
blais, line 89: Can you add under which conditions this is required to use over the vanilla function NamedTemporaryFile() from tempfile?

beancount/utils/file_utils_test.py
```diff
@@ -42,23 +43,31 @@ def walk(fords):
         self.assertEqual(
             sorted(
                 [
-                    "alice/bottle.txt",
-                    "rabbit/suit/glasses.txt",
-                    "caterpillar/who-are-you.txt",
+                    Path(self.tempdir, "alice/bottle.txt"),
```
blais: I don't understand why this change is necessary.

beancount/utils/test_utils.py
trim21: done

beancount/utils/file_utils_test.py
trim21: `walk` may yield abs path or rel path on different OS. And also, `os.sep` problem.

trim21:
```text
beancount/utils/file_utils_test.py::TestFileUtilsFind::test_find_files FAILED

========================================================================== FAILURES ===========================================================================
______________________________________________________________ TestFileUtilsFind.test_find_files ______________________________________________________________

self = <beancount.utils.file_utils_test.TestFileUtilsFind testMethod=test_find_files>

    def test_find_files(self):
        def walk(fords):
            return sorted(clean(self.tempdir, file_utils.find_files(fords)))

>       self.assertEqual(
            sorted(
                [
                    "alice/bottle.txt",
                    "rabbit/suit/glasses.txt",
                    "caterpillar/who-are-you.txt",
                ]
            ),
            walk([self.tempdir]),
        )
E       AssertionError: Lists differ: ['alice/bottle.txt', 'caterpillar/who-are-y[30 chars]txt'] != ['C:\\Users\\Trim21\\AppData\\Local\\Temp\\[241 chars]txt']
E
E       First differing element 0:
E       'alice/bottle.txt'
E       'C:\\Users\\Trim21\\AppData\\Local\\Temp\\[41 chars].txt'
E
E       - ['alice/bottle.txt', 'caterpillar/who-are-you.txt', 'rabbit/suit/glasses.txt']
E       + ['C:\\Users\\Trim21\\AppData\\Local\\Temp\\TestFileUtilsFind.d1t917_c\\alice\\bottle.txt',
E       +  'C:\\Users\\Trim21\\AppData\\Local\\Temp\\TestFileUtilsFind.d1t917_c\\caterpillar\\who-are-you.txt',
E       +  'C:\\Users\\Trim21\\AppData\\Local\\Temp\\TestFileUtilsFind.d1t917_c\\rabbit\\suit\\glasses.txt']

beancount\utils\file_utils_test.py:43: AssertionError
```

trim21: Ah, I know. I should fix this in `clean` instead of tests

beancount/ops/documents_test.py
```diff
@@ -42,7 +47,7 @@ def test_process_documents(self):
 
               2014-07-10 document Liabilities:US:Bank  "does-not-exist.pdf"
 
-            """).replace("ROOT", self.root)
+            """).replace("ROOT", self.root.replace("\\", r"\\"))
```
dnicolodi, line 46: Why modify a test that you skip anyway?

beancount/ops/documents_test.py
```diff
@@ -61,7 +66,7 @@ def test_process_documents(self):
           option "plugin_processing_mode" "raw"
           2014-06-08 document Assets:US:Bank:Checking "ROOT/Assets/US/Bank/Checking/2014-06-08.bank-statement.pdf"
           2014-07-10 document Liabilities:US:Bank "ROOT/does-not-exist.pdf"
-        """).replace("ROOT", self.root)
+        """).replace("ROOT", self.root.replace("\\", r"\\"))
```
dnicolodi: Also here.

beancount/core/account.py
```diff
@@ -195,6 +195,9 @@ def walk(root_directory: Account) -> Iterator[Tuple[str, Account, List[str], Lis
     This only yields directories that are accounts... it skips the other ones.
     For convenience, it also yields you the account's name.
 
+    Note:
+        this only work as expected on POSIX.
+
```
dnicolodi: Why is it so? I don't see anything POSIX specific in this function.

dnicolodi (changes requested):

beancount/core/account.py
trim21: os.walk

beancount/ops/documents_test.py
trim21: make it easier to enable after we fix printer

beancount/core/account.py
dnicolodi: `os.walk` is available on Windows too.

Do you realize that I am putting it time and effort to review your patches and that this style of replied to my comments is really annoying? I have tried to convey this several times now. I understand that English is not your native language, it is not mine either. For this reason it would be better to spend more words, not less, to explain your position. From your reply above the only conclusion I can draw is that you have no idea of what you are writing about.

beancount/ops/documents_test.py
dnicolodi: Which fix to the printer do you have in mind? How would changing the test in this way help?

If the change to the printer you have in mind it the one that would make it escape file paths, as we discussed in the other PR, the change here would actually produce wrong results, no more correct results: backslashes in the file path would end up being escaped twice. If you have something else in mind, please elaborate.

trim21: no matter we fix printer or not the input is wrong on Windows

trim21: I think you mis-read the code here, this is the input of parser, and backward slashes are expected to be escaped. `self.root` is the fileapth with single backslashe, which need to be replace by double backslashes.


I mis-read the code too, `assertEqualEntries` doesn't compare entries with printer.

Now I find a way to workaround `assertEqualEntries` here, it's no longer skipped.

beancount/core/account.py
trim21: Ah, you are right. the path seprator of root by yield `account.walk` doesn't matter here.

At least we do handle it as filepath this time.

beancount/utils/test_utils.py
trim21: I think this can be used to fix `test_util.docfile` but it will conflict with other PR.

blais (approved):
Thanks for fixing tests on Windows.

blais:
Thanks for fixing tests on Windows.
Please rebase and I'll merge.

PR #879, opened 2024-11-29
match pre-commit hooks version and lock schema-store

trim21:
`validate-pyproject` may validate all ruff rules in ignore/selected are defined in json schema and it may not match the ruff version we are using in pre-commit hooks, and it will fail even ruff config is still valid.

for example, ruff will rename rule TCH to TC in v0.8, and using TC will make ruff v0.7.2 not happy, use TCH will make `validate-pyproject` not happy.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,10 +18,13 @@ repos:
     rev: v0.22
     hooks:
       - id: validate-pyproject
-        additional_dependencies: [ "validate-pyproject-schema-store[all]" ]
+        # schema-store include schema for ruff config,
+        # need to match ruff hook version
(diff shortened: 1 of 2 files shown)


blais:
Why can't we match the versions of ruff with the ones used here?
Wouldn't it be more appropriate to switch to using ruff v0.8?

trim21:
> Why can't we match the versions of ruff with the ones used here? 
> Wouldn't it be more appropriate to switch to using ruff v0.8?

Yes, another option is to do this:

```yaml
  - repo: https://github.com/abravalheri/validate-pyproject
    rev: v0.22
    hooks:
      - id: validate-pyproject
        additional_dependencies: [ "validate-pyproject-schema-store[all]==2024.11.25" ]
```

PR #882, opened 2024-12-04
tests: fix beancount/loader_test.py on win32

trim21:


trim21:
🤔 There is a problem in loader test, we only cache beancount file need 1s to load, so the tests will pass only on high performance hardware, and will fail on low performance hardware.

beancount/loader.py
```diff
@@ -206,6 +206,7 @@ def wrapped(toplevel_filename, *args, **kw):
                 try:
                     result = pickle.load(file)
                 except Exception as exc:
+                    print("failed to load cache", exc)
```
blais: remove print

trim21: Yes it't not ready for review yet. I use this to find why test failed in ci.

blais:
Please rebase, I just merged another 876.

beancount/loader_test.py
```diff
@@ -479,8 +492,15 @@ def test_load_glob_relative_abs(self):
...
-            ["apples.beancount", "bananas.beancount", "oranges.beancount", "top.beancount"],
-            list(map(path.basename, options_map["include"])),
+            sorted(
+                [
+                    "apples.beancount",
+                    "bananas.beancount",
+                    "oranges.beancount",
+                    "top.beancount",
+                ]
+            ),
```
dnicolodi: It does not matter much, but sorting a literal list seems unnecessary: if the order matters, just write the literal in the correct order, if the order does not matter (and there aren't duplicates) use a set.

beancount/loader_test.py
```diff
@@ -559,7 +579,8 @@ def test_load_cache(self):
                 },
             )
             top_filename = path.join(tmp, "apples.beancount")
-            entries, errors, options_map = loader.load_file(top_filename)
+            with mock_time_time():
+                entries, errors, options_map = loader.load_file(top_filename)
```
dnicolodi: I'm a bit suspicious of this change: I don't think that this code has always been run on system where it takes more than 1.2 seconds to execute. Or is it the other way around and it takes more than 1.2 seconds to execute it on the Windows CI machines and you are artificially capping the execution time to 1.2 seconds? I agree that having the test depend on the execution time is not good, but this deserves a comment explaining what is going on.

beancount/loader_test.py
```diff
@@ -512,8 +530,8 @@ def test_load_glob_relative_mixed(self):
         self.assertEqual(2, len(entries))
         self.assertTrue(all(path.isabs(filename) for filename in options_map["include"]))
         self.assertEqual(
-            ["1.apples.beancount", "2.oranges.beancount", "top.beancount"],
-            list(map(path.basename, options_map["include"])),
+            set(["1.apples.beancount", "2.oranges.beancount", "top.beancount"]),
```
dnicolodi:
```suggestion
            {"1.apples.beancount", "2.oranges.beancount", "top.beancount"},
```

beancount/loader_test.py
```diff
@@ -704,7 +733,8 @@ def test_load_cache_disable(self):
                     },
                 )
                 filename = path.join(tmp, "apples.beancount")
-                entries, errors, options_map = loader.load_file(filename)
+                with mock_time_time():
```
dnicolodi: This test function is executed with `PICKLE_CACHE_THRESHOLD` set to 0. Thus this should not be necessary.

beancount/loader_test.py
```diff
@@ -680,7 +708,8 @@ def test_load_cache_override_filename_pattern_by_argument(self):
                 },
             )
             filename = path.join(tmp, "apples.beancount")
-            entries, errors, options_map = loader.load_file(filename)
+            with mock_time_time():
```
dnicolodi: Also here.

beancount/loader_test.py
```diff
@@ -660,7 +687,8 @@ def test_load_cache_override_filename_pattern_by_env_var(self):
                     },
                 )
                 filename = path.join(tmp, "apples.beancount")
-                entries, errors, options_map = loader.load_file(filename)
+                with mock_time_time():
```
dnicolodi: And here.

beancount/loader_test.py
trim21: Looks like it's the other way around

https://github.com/beancount/beancount/blob/793d8be9f3e7f47230c270d6f0faf015ccf4cdab/beancount/loader.py#L236-L238

We only cache beancount take more than 1s to load, and we have test to check if cache file is generated.

tests pass on POSIX and windows ci(py<3.13), so I think it normally take more than 1s to execute.

dnicolodi (changes requested):
I think some of the mocking of `time.time()` is not necessary. I don't think the others may have something to do with the resolution of the file modification times on Windows filesystems, or something like that, more than of the execution time.

beancount/loader_test.py
dnicolodi: I still don't get it. Are you saying that the test works only if the loafing takes more than 1 seconds? I'm pretty sure that it takes less than than on my machine.

trim21: > I still don't get it. Are you saying that the test works only if the loafing takes more than 1 seconds? I'm pretty sure that it takes less than than on my machine.

I need to take a deeper look into it, I guess it also has something to do with the time resolution.

trim21: OK, I know. on windows the time resolution is not small enough to distinguish time_before and time_after, and the time delta we have will be `0.0`.

So even we have `PICKLE_CACHE_THRESHOLD=0` we will need to mock time.

trim21: and we should use `time.monotonic()` here, even though it can't avoid time resolution problem.

dnicolodi: Somewhere there must be a condition equivalent to `(time_after - time_before) > PICKLE_CACHE_THRESHOLD` which fails when `time_after == time_before`. What about turning `>` into `>=`? That should solve the issue without any practical consequence on real usage (on Windows the default resolution for `time.time()` is 13 ms and `PICKLE_CACHE_THRESHOLD` is much larger than than in any practical scenario). Would this fix the issue?

trim21:
> I think some of the mocking of `time.time()` is not necessary. I don't think the others may have something to do with the resolution of the file modification times on Windows filesystems, or something like that, more than of the execution time.

they are necessary, see https://github.com/beancount/beancount/pull/882#discussion_r1869626554

beancount/loader_test.py
dnicolodi: Namely, turn this line https://github.com/beancount/beancount/blob/793d8be9f3e7f47230c270d6f0faf015ccf4cdab/beancount/loader.py#L242
into
```python
    if time_after - time_before >= time_threshold:
```

trim21: > Somewhere there must be a condition equivalent to `(time_after - time_before) >PICKLE_CACHE_THRESHOLD` which fails when `time_after == time_before`. What about turning `>` into `>=`? That should solve the issue without any practical consequence on real usage (on Windows the default resolution for `time.time()` is 13 ms and `PICKLE_CACHE_THRESHOLD` is much larger than than in any practical scenario. Would this fix the issue?

yes, that's one option

dnicolodi: Another option is to use an higher resolution clock via `time.perf_counter()`.

dnicolodi:
Thanks @trim21. Looks good

dnicolodi (approved):

beancount/loader_test.py
trim21: that's `time.monotonic`

blais (approved):

PR #884, opened 2024-12-04
fix!: escape filename in document filename

trim21:
The intention of our printer is to generate same syntax the parser accept.

Since we expect backslash in filename to be escaped, it's rational that we also escape it in our printer.

This should not have huge impact because it's current output in this case is already broken.

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -351,8 +351,8 @@ class EntryPrinter:
 
     def Document(self, entry: Document, oss):
         oss.write(
-            '{e.date} document {e.account} "{filename}"'.format(
-                e=entry, filename=entry.filename.replace("\\", r"\\")
+            "{e.date} document {e.account} {filename}".format(
+                e=entry, filename=misc_utils.quote_string(entry.filename)
             )
         )
         if entry.tags or entry.links:
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -21,6 +21,7 @@ from beancount.core import account
 from beancount.core import data
 from beancount.core import interpolate
 from beancount.core import display_context
+from beancount.core.data import Document
 from beancount.utils import misc_utils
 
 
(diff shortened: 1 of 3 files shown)


dnicolodi:
> This should not have huge impact because it's current output in this case is already broken.

I disagree. There can be code that relies on the current `beancount.parser.printer.print_entry()` behavior and escapes the file path component of the `document` directive before passing it to `print_entry()`. This code works fine now and it will be broken by this change.

Going beyond the backward compatibility, it is a bad idea to escape only backslashes. Any character is allowed in an Unix file name, including newlines and other non printable characters. Before this change is possible to represent these in a Beancount ledger simply passing them escaped to `print_entries()`. After this change there is no way to pass them because the backslashes in the escape sequences would be doubled.

It is also undesirable to have only a few strings being subject to escaping but not others. This requires the users of the API to remember which fields needs to be escaped before passing them to `print_entry()` and which do not. I know that this is already a bit the case with some escaping being applied to payee and narration, but I don't see this as a justification for adding more special cases.

I'm not opposed to this change, but the pro and cons needs to be presented correctly.

trim21:
> I disagree. There can be code that relies on the current beancount.parser.printer.print_entry() behavior and escapes the file path component of the document directive before passing it to print_entry(). This code works fine now and it will be broken by this change.

>  it is a bad idea to escape only backslashes. Any character is allowed in an Unix file name, including newlines and other non printable characters. Before this change is possible to represent these in a Beancount ledger simply passing them escaped to print_entries(). After this change there is no way to pass them because the backslashes in the escape sequences would be doubled.

Thanks for reminding, forget that some characters like `\n` is allowed in POSIX filename.

I think we should escape all characters here.

> It is also undesirable to have only a few strings being subject to escaping but not others. This requires the users of the API to remember which fields needs to be escaped before passing them to print_entry() and which do not. I know that this is already a bit the case with some escaping being applied to payee and narration, but I don't see this as a justification for adding more special cases.

Ideally users should not need to escape anything, and printer should escape anything needed.

what we didn't escape yet is note.comment, event name, description, query name and query string.

dnicolodi:
This is not a fix, the commit message is misleading on this aspect. The reason behind the change should be noted in the commit message. Explicitly warning about the backward compatibility issue would also be nice.

trim21:
for someone who need this can parse beancount as AST to preserve the original content: https://github.com/trim21/beancount-ast

PR #885, opened 2024-12-04
tests: fix script tests on win32

trim21:


beancount/scripts/directories_test.py
```diff
@@ -55,7 +55,8 @@ def test_validation(self):
                 any(
                     expected_account in message
                     for expected_account in expected_error_accounts
-                )
+                ),
+                f"{expected_error_accounts=!r} not in {message=!r}",
```
dnicolodi: This test error message is misleading as it can be read as the literal string to the left hand side of the `not in` operator is searched in the literal string on the right hand side. If you would like a better error message, the test can be rewritten as something like:
```python
for error in errors:
    for expected_account in expected_accounts:
        self.assertNotIn(expected_account, str(error))
```
or something like this.

beancount/scripts/doctor_test.py
```diff
@@ -189,7 +188,7 @@ def test_context(self, filename):
         self.assertRegex(rv.stdout, "50.02")
 
     @test_utils.docfile
-    def test_context_multiple_files(self, filename):
+    def test_context_multiple_files(self, filename: str):
```
dnicolodi, line 191: Spurious change.

dnicolodi (changes requested):

beancount/scripts/directories_test.py
trim21: I reverted this line

trim21:
@blais hi, would you mind review this first? I'm hoping to fix conflict in 886 after this is merged

dnicolodi:
@trim21 It is impolite to ask another review when comments from previous reviews have not been addressed.

trim21:
I'm wishing to get some comments from blais.

Also I didn't ignore your comments if you feel that way.

dnicolodi:
> I'm wishing to get some comments from blais.
> 
> Also I didn't ignore your comments if you feel that way.

One of the comments I left has not been addressed and you didn't react in any way to it. How you define this other than ignoring?

trim21:
> > I'm wishing to get some comments from blais.
> > Also I didn't ignore your comments if you feel that way.
> 
> One of the comments I left has not been addressed and you didn't react in any way to it. How you define this other than ignoring?

I'm sorry you feel this way, there will be conflicts in 886, and also 886 is still a draft. So I'm proposaling a priority between these 2 PRs.

dnicolodi:
There is a comment on this PR that has not been addressed. Do you see it?

beancount/scripts/doctor_test.py
blais: Bah that's no big deal

dnicolodi: Sure. But it is @trim21 approach at dealing with review comments that is really annoying. Not even after pointing it out that there is an unaddressed comment I have been able to have they recognize it. I wonder why I should spend time to review the PR if my parts of my reviews are simply ignored.

blais: Hey Daniele, if it becomes unpleasant for you, you shouldn't have to spend time on it.  
This is a free for-fun project. You should only spend time if you enjoy it. This should be fun.
Don't worry I'll be as detail oriented with all contributors as possible, even if it takes more time.

PR #886, opened 2024-12-04
make sure open file in binary mode or encoding specified text mode

trim21:
There is a preview ruff rule

https://docs.astral.sh/ruff/rules/unspecified-encoding/#unspecified-encoding-plw1514

### DIFF
--- a/beancount/projects/export.py
+++ b/beancount/projects/export.py
@@ -26,6 +26,7 @@ Query Language query. However, BQL is not there yet.
 __copyright__ = "Copyright (C) 2018  Martin Blais"
 __license__ = "GNU GPLv2"
 
+import sys
 from typing import NamedTuple, Tuple, List, Set, Any, Dict
 from decimal import Decimal
 import csv
@@ -252,6 +253,15 @@ def write_table(table: Table, outfile: str):
     writer.writerows(table.rows)
 
 
+# python 3.15 will make utf-8 mode enabled by default
+# so use a locale encoding for backward compatibility
+if sys.version_info < (3, 10):
+    # locale encoding is added in 3.10
+    OUTPUT_ENCODING = None
+else:
+    OUTPUT_ENCODING = "locale"
+
+
 @click.command(help=__doc__)
 @click.argument("filename")
 @click.option(
@@ -272,37 +282,37 @@ def write_table(table: Table, outfile: str):
 @click.option(
     "--output",
     "-o",
-    type=click.File("w"),
+    type=click.File("w", encoding=OUTPUT_ENCODING),
     help="CSV filename to write out the final joined table to.",
 )
 @click.option(
     "--output_commodities",
     "-c",
-    type=click.File("w"),
+    type=click.File("w", encoding=OUTPUT_ENCODING),
     help="CSV filename to write out the commodities table to.",
 )
 @click.option(
     "--output_accounts",
     "-a",
-    type=click.File("w"),
+    type=click.File("w", encoding=OUTPUT_ENCODING),
     help="CSV filename to write out the accounts table to.",
 )
 @click.option(
     "--output_prices",
     "-p",
-    type=click.File("w"),
+    type=click.File("w", encoding=OUTPUT_ENCODING),
     help="CSV filename to write out the prices table to.",
 )
 @click.option(
     "--output_rates",
     "-r",
-    type=click.File("w"),
+    type=click.File("w", encoding=OUTPUT_ENCODING),
     help="CSV filename to write out the rates table to.",
 )
 @click.option(
     "--output_postings",
     "-m",
-    type=click.File("w"),
+    type=click.File("w", encoding=OUTPUT_ENCODING),
     help="CSV filename to write out the postings table to.",
 )
 def main(
--- a/beancount/scripts/example.py
+++ b/beancount/scripts/example.py
@@ -1836,7 +1836,13 @@ class LiberalDate(click.ParamType):
 @click.option("--date-birth", type=LiberalDate(), help="Fictional date of birth.")
 @click.option("--seed", "-s", type=int, help="Random seed.")
 @click.option("--no-reformat", is_flag=True, help="Do not reformat the output.")
-@click.option("--output", "-o", type=click.File("w"), default="-", help="Output filename.")
+@click.option(
+    "--output",
+    "-o",
+    type=click.File("w", encoding="utf-8"),
+    default="-",
+    help="Output filename.",
+)
 @click.option("--verbose", "-v", type=bool, help="Produce logging output.")
 @click.version_option(message=VERSION)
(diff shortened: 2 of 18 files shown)


dnicolodi:
It would be nice to have the commit message provide a reason for this change.

I think that using UTF-8 encoding (the use of the "utf8 mode" in the commit message is imprecise) is correct in all instances. Beancount ledgers are expected to use UTF-8 encoding. However, I don't see why we need to override the environment encoding for all other kind of files. If the user uses whatever Windows encoding is used these days, I'm quite sure they would like Beancount to emit CSV files in that encoding, so that they can use them without issues on their system, not UTF-8. If they want to use UTF-8 for everything they would have chosen to use this encoding for their environment.

What am I missing?

trim21:
Python will make utf-8 mode by default in python3.15, then all `open` without encoding will means utf-8 encoding.


Yes, the csv output may be better to use `locale` encoding.

dnicolodi:
> Python will make utf-8 mode by default in python3.15

Do you care to elaborate on why that refutes what I wrote above? If Python will use UTF-8 encoding in 3.15 (which is a couple of years away, at least three years from widespread adoption) why do we need to change the encoding we use for writing CSV files today? This is going to break things now for users of systems with different default encoding.

dnicolodi:
This will conflict with the changes in the other PRs. Maybe hold on reworking this till the other PRs are merged.

trim21:
> Do you care to elaborate on why that refutes what I wrote above? If Python will use UTF-8 encoding in 3.15 (which is a couple of years away, at least three years from widespread adoption) why do we need to change the encoding we use for writing CSV files today? This is going to break things now for users of systems with different default encoding.

No, I agree with you

beancount/utils/encryption.py
```diff
@@ -33,7 +33,7 @@ def is_encrypted_file(filename):
     if ext == ".gpg":
         return True
     if ext == ".asc":
-        with open(filename) as encfile:
+        with open(filename, encoding="utf-8") as encfile:
```
blais: That one should also be None or "locale" I think

blais (approved):

beancount/utils/encryption.py
trim21: Asc file is pure ascii file, right?

blais: I think the encoding of the contents are base64 (off the top of my head)

trim21: Then it's ok to use utf-8 I think

trim21:
Oh sorry click wrong button

beancount/projects/export.py
```diff
@@ -252,6 +253,15 @@ def write_table(table: Table, outfile: str):
     writer.writerows(table.rows)
 
 
+# python 3.15 will make utf-8 mode enabled by default
+# so use a locale encoding for backward compatibility
+if sys.version_info < (3, 10):
+    __output_encoding = None
```
dnicolodi: Why do you prefix the variable name with two underscores? I would not use even one, but two just look odd.

beancount/scripts/example.py
```diff
@@ -1836,7 +1836,13 @@ def convert(self, value, param, ctx):
...
 @click.option("--seed", "-s", type=int, help="Random seed.")
 @click.option("--no-reformat", is_flag=True, help="Do not reformat the output.")
-@click.option("--output", "-o", type=click.File("w"), default="-", help="Output filename.")
+@click.option(
+    "--output",
+    "-o",
+    type=click.File("w", encoding="utf-8"),
+    default="-",
+    help="Output filename.",
+)
```
dnicolodi, line 1845: Can we keep this on one line? This black-style formatting is really ugly.

beancount/utils/encryption.py
dnicolodi: I think it should be `"ascii"`.

beancount/utils/encryption_test.py
```diff
@@ -141,7 +141,7 @@ def encrypt_as_file(self, string, encrypted_filename):
             "--output=-",
             stdin=string.encode("utf8"),
         )
-        with open(encrypted_filename, "w") as encfile:
+        with open(encrypted_filename, "w", encoding="utf-8") as encfile:
```
dnicolodi: `"ascii"` here too

beancount/projects/export.py
blais: According to the style guide, globals should be in caps, e.g. OUTPUT_ENCODING or _OUTPUT_ENCODING

beancount/scripts/example.py
trim21: Can't. It reach current length limit.

beancount/utils/encryption.py
trim21: we should open file in ascii encoding and suppress UnicodeDecodeError here.

beancount/projects/export.py
trim21: done

beancount/utils/encryption.py
```diff
@@ -33,10 +34,12 @@ def is_encrypted_file(filename):
...
     if ext == ".asc":
-        with open(filename) as encfile:
-            head = encfile.read(1024)
-            if re.search("--BEGIN PGP MESSAGE--", head):
-                return True
+        # python will still raise UnicodeDecodeError if file content is not in ascii encoding
+        with contextlib.suppress(UnicodeDecodeError):
+            with open(filename, encoding="ascii") as encfile:
+                head = encfile.read(1024)
+                if re.search("--BEGIN PGP MESSAGE--", head):
```
trim21, line 41: This regex is unnecessary

beancount/utils/encryption.py
dnicolodi: An armored GPG encrypted file is a base64 encoded binary. It is guaranteed to contain only ascii characters. Do you have evidence of the contrary? Why do you want to suppress decoding errors?

trim21: > An armored GPG encrypted file is a base64 encoded binary. It is guaranteed to contain only ascii characters. Do you have evidence of the contrary? Why do you want to suppress decoding errors?

This is function `def is_encrypted_file(filename):`, at this point we do not know if this `.asc` file is GPG encrypted file or not.

If someone use `.asc` file extension and it's not really GPG encrypted file, we open it with `encoding=ascii`, it may cause UnicodeDecodeError here. So we should supress UnicodeDecodeError and it will go into another code path and return False.

dnicolodi: Oh. Ok. This function is probing whether the file is or is not an armored GPG encrypted file. It makes sense to do not raise an exception if the file does not have the expected encoding. It looks ok.

blais:
Please rebase

blais (approved):

PR #891, opened 2024-12-07
parser: Port the scanner generation from Flex to RE/flex

dnicolodi:


blais:
So this would turn the C + Python Beancount into one that depends on C++?
Idea being we get part of the benefits of the work we did on the cpp branch, but in the master branch?

blais:
(It should probably be mentioned I'm more or less decided to work on a Rust port by now.
Have done some work already, not checked in yet.)

dnicolodi:
This is a rebase of an old PR of mines, #581. The code evolved a bit between then and now and what I hoped to be a trivial rebase turned out to be a bit more involved. There is still something wrong, but I run out of time yesterday.

At the time I didn't pursue it further because it seemed imminent that we would land the C++ work in v3 and we were still building with setuptools, which made using external dependencies a pain. Meson solves the build issues and I wanted to see if we can use some of the work done in the C++ branch.

Part of the reasons for dusting this off is some profiling I did which shows that parsing takes 22% of the time required to run `bean-check` on a ledger generated by `bean-example` and I'm curious to see how much faster it can be made. One of the low hanging fruits is to remove some regular expression matching done in the parser in favor of some tighter lexing. But that requires a better lexer generator.

blais:
Given parsing is onlty ~20% of the total time, wouldn't it be better to focus on the rest?
How much faster is it?

Here's what I see:
Pros:
+ can do unicode properly
+ potentially faster
Cons:
- adds dependency on C++
-

trim21:
> Given parsing is onlty ~20% of the total time, wouldn't it be better to focus on the rest? How much faster is it?
> 
> Here's what I see: Pros:
> 
> * can do unicode properly
> * potentially faster
>   Cons:
> 
> * adds dependency on C++

If we are now using meson I think add dependency on C++ is not a liability anymore, meson handle it really well. No extra system package is needed.

trim21:
If we are using meson wrap it would be nice to add these to `.gitignore`:

```gitignore
/subprojects/*

!/subprojects/*.wrap
!/subprojects/packagefiles
```

otherwise after build:

image

PR #895, opened 2024-12-15
typings: improve; add py.typed marker

yagebu:
- use future import to avoid interpretation of types outside of type
  checking
- some minor fixes in types and/or docstrings as well
- mypy infers some weird stuff for Decimal.as_tuple() - hence some
  ignores for those

### DIFF
--- a/beancount/core/account.py
+++ b/beancount/core/account.py
@@ -5,6 +5,8 @@ of their associated postings. This is achieved by building a realization; see
 realization.py for details.
 """
 
+from __future__ import annotations
+
 __copyright__ = "Copyright (C) 2013-2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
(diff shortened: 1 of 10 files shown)


trim21:
should we consider merge https://github.com/beancount/fava/tree/main/stubs/beancount into this repo?

blais:
Yes and no. I think we should add type annotations everywhere.
Last week I did an experiment to do this with LLMs and it ... worked. Amazing.
I then setup aider and aider.el and then I got busy.
I think this could be easy, but I want to check all the diffs.

trim21:
> Yes and no. I think we should add type annotations everywhere. Last week I did an experiment to do this with LLMs and it ... worked. Amazing. I then setup aider and aider.el and then I got busy. I think this could be easy, but I want to check all the diffs.

I cause some misunderstanding, I mean merge pyi type annotation back to source python file, not as seprated pyi type stub.

blais:
Sure, but I feel that having an LLM generate high quality versions is equivalent to me.
Sometimes even better, I've seen it fix errors!

PR #898, opened 2024-12-21
refactor: rewrite NamedTuple with class inheritance

trim21:


blais:
SGTM. Though we'll looking pretty immodern with NamedTuple, all the cool kids are using dataclasses now :-)

trim21:
> SGTM. Though we'll looking pretty immodern with NamedTuple, all the cool kids are using dataclasses now :-)

Yes, especially we need to overwrite `__init__`/`__new__`

PR #899, opened 2024-12-21
refactor: cleanup some really old code

trim21:


beancount/__init__.py
```diff
@@ -7,9 +7,6 @@
 # Check the version requirements.
 import sys
 
-if (sys.version_info.major, sys.version_info.minor) < (3, 3):
```
blais, line 10: A check is nice, why remove it?

blais (approved):

beancount/__init__.py
trim21: I don't think it's possible to install beancount in that old python versions

https://github.com/beancount/beancount/blob/876b86950e6129a505867fba9bfb78b7a9c413d9/pyproject.toml#L26

blais: I think it's more for development.
Can you keep the check and update the version number?

trim21: I think that's also not possible since now we use `pip install --no-isolate -e .` to install this package for development?

blais: I don't pip install anything, I mangle my `PYTHONPATH` like a dinosaur (it works well!)

trim21: So you also build meson manually and copy binary extension back to source like I do?

blais: Yes

trim21: Then I have a snippet in meson.build for this case, I'll send it later 😁

trim21: added it back with updated version number

PR #901, opened 2024-12-21
build: add a build_ext meson target for old-style developers

trim21:
Add a extra ninja target for developer using old-style in-tree development.
it doesn't contribute to the building but developing.

A developer can do this to get save result like `python setup.py build_ext --inplace`
to copy binary extension back to source tree.

```shell
meson setup build # --wipe
ninja -C build build_ext
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -244,7 +244,7 @@ parser = py.extension_module(
     subdir: 'beancount/parser',
 )
 
-# this is a extra ninja target for developer using old-style in-tree development.
+# This is a extra ninja target for developer using old-style in-tree development.
 # it doesn't contribute to the building but developing.
 # A developer can do this to get save result like `python setup.py build_ext --inplace`
 # to copy binary extension back to source tree.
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -244,6 +244,28 @@ py.extension_module(
     subdir: 'beancount/parser',
 )
 
+# this is a extra ninja target for developer using old-style in-tree development.
```
blais: Nit: please capitalize

trim21:
```suggestion
# This is a extra ninja target for developer using old-style in-tree development.
```

blais:
I just used `make build`
```
lumia [git|master]:~/p/beancount$ make -n build
meson setup --reconfigure -Dtests=enabled build/
ninja -C build/
cp build/_parser.cpython-312-x86_64-linux-gnu.so beancount/parser/
```

trim21:
> I just used `make build`
> 
> ```
> lumia [git|master]:~/p/beancount$ make -n build
> meson setup --reconfigure -Dtests=enabled build/
> ninja -C build/
> cp build/_parser.cpython-312-x86_64-linux-gnu.so beancount/parser/
> ```

I also have a script for this in my local taskfile.yaml, but since it's a common use case I think it would be more nice to have one in meson.build, this works on all python versions and OSs

PR #902, opened 2024-12-21
style: enable ruff isort rule

trim21:


blais:
Is there a way to configure isort so that it doesn't import multiple symbols on the same line?
I've avoided that historically in this codebase.
Not a huge deal to change but if it's possible I'd like it better that way.

trim21:
> Is there a way to configure isort so that it doesn't import multiple symbols on the same line? I've avoided that historically in this codebase. Not a huge deal to change but if it's possible I'd like it better that way.

yes

blais:
Please make the change then, much appreciated.

trim21:
> Please make the change then, much appreciated.

done

PR #903, opened 2024-12-21
refactor: improve typing

trim21:
1. we can inhert from `NamedTuple` directly.
2. Original type of `Position.cost` doesn't match, it allow `Cost | CostSpec | None` but typed as `Cost`

### DIFF
--- a/beancount/core/amount.py
+++ b/beancount/core/amount.py
@@ -32,10 +32,7 @@ CURRENCY_RE = "|".join(
 )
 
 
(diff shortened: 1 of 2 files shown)


trim21:
There are many code using `MISSING` as `None` or conversely, should we unify them and only use one sentinel value?

(there is even also a `UNSET`)

blais:
I think we should ignore those Python incongruities for now.
The Rust schema which will go a lot farther into clarifying the types and values between parser and final.

PR #905, opened 2024-12-22
Further types

yagebu:
Forgot to actually include the py.typed marker in #895.
With these changes, it will be possible to remove the stubs that I've so far been using to typecheck Fava, see https://github.com/beancount/fava/pull/1920 :)


trim21:
we should consider enabled mypy on beancount someday？

beancount/core/account_types.py
```diff
@@ -14,10 +14,13 @@
 __license__ = "GNU GPLv2"
 
 import re
+from typing import TYPE_CHECKING
 from typing import NamedTuple
 
 from beancount.core import account
-from beancount.core.account import Account
+
+if TYPE_CHECKING:
```
trim21, line 22: Is thre a circle imports here?

beancount/core/data.py
```diff
@@ -31,6 +33,18 @@
 Flag = str
 Meta = Dict[str, Any]
 
+if TYPE_CHECKING:
```
trim21: `BeancountError` looks usefull, I suggest we add it as real runtime Protocol class so downsteam project can import it without the need of `typing.TYPE_CHECKING`

beancount/parser/_parser.pyi
```diff
@@ -4,4 +4,4 @@ from beancount.parser.grammar import Builder
 
 class Parser:
     def __init__(self, builder: Builder, debug: bool) -> None: ...
```
trim21: Some minor problem here.

```suggestion
    def __init__(self, builder: Builder, debug: bool = ...) -> None: ...
```

And `builder` actually take a `lexer.LexBuilder`, which is the base class of `Builder`, not other way around

so it should be 

```suggestion
    def __init__(self, builder: lexer.LexBuilder, debug: bool = ...) -> None: ...
```

https://github.com/beancount/beancount/blob/859505420f3b3eaf6ecc99f7aaf2126784164b98/beancount/parser/lexer.py#L63-L65

https://github.com/beancount/beancount/blob/859505420f3b3eaf6ecc99f7aaf2126784164b98/beancount/parser/parser_test.py#L327-L328

https://github.com/beancount/beancount/blob/859505420f3b3eaf6ecc99f7aaf2126784164b98/beancount/parser/parser_test.py#L334-L335

beancount/core/account_types.py
yagebu: no circle in this case I think, but for clarity and avoiding runtime overhead, I think it's preferable to have type-only imports in these conditional blocks, see also https://docs.astral.sh/ruff/rules/typing-only-first-party-import/

beancount/parser/_parser.pyi
```diff
@@ -4,4 +4,4 @@ from beancount.parser.grammar import Builder
...
-    def parse(self, file: io.IOBase, filename: str, lineno: int) -> None: ...
+    def parse(self, file: io.IOBase, filename: str | None, lineno: int) -> None: ...
```
trim21: what we actually have:

```suggestion
    def parse(self, file: io.IOBase, filename: str | None = ..., lineno: int = ...) -> None: ...
```

beancount/core/account_types.py
trim21: I don't think the runtime overhead may be a problem here? In most case `beancount.core.account` will exist in module cache

beancount/loader.py
```diff
@@ -76,9 +75,20 @@ class LoadError(NamedTuple):
...
+_load_file: Callable[
+    [str, Any, list[Any] | None, str | None],
+    tuple[data.Directives, list[data.BeancountError], OptionsMap],
+]
+
+
 def load_file(
-    filename, log_timings=None, log_errors=None, extra_validations=None, encoding=None
-):
+    filename: str,
```
trim21, line 86: I think it can be `pathlib.Path` too

beancount/loader.py
```diff
@@ -115,13 +125,13 @@ def load_file(
...
 
 def load_encrypted_file(
-    filename,
-    log_timings=None,
-    log_errors=None,
-    extra_validations=None,
-    dedent=False,
-    encoding=None,
-):
+    filename: str,
```
trim21, line 128: also `pathlib.Path`

beancount/ops/summarize.py
```diff
@@ -29,16 +33,23 @@
 from beancount.parser import options
 from beancount.utils import bisect_key
 
+if TYPE_CHECKING:
+    from beancount.core.account import Account
+    from beancount.core.account_types import AccountTypes
+    from beancount.core.data import Currency
```
trim21, line 39: we already have `from beancount.core import data`, should we just reuse it?

beancount/parser/booking_method.py
```diff
@@ -19,6 +20,11 @@
 from beancount.core.number import ZERO
 from beancount.core.position import Cost
 
+if TYPE_CHECKING:
+    from decimal import Decimal
+
+    from beancount.core.position import Cost
```
trim21: we already have one.

and I think `decimal.Decimal` will definitely exists in module cache, so no need to avoid import overhead here.

beancount/utils/encryption.py
```diff
@@ -9,20 +9,22 @@
 from os import path
 
 
-def is_gpg_installed():
+def is_gpg_installed() -> bool:
     """Return true if GPG 1.4.x or 2.x are installed, which is what we use and support."""
     try:
-        pipe = subprocess.Popen(
+        pipe = subprocess.Popen(  # type: ignore[call-overload]
```
trim21: This is caused by `shell=0`, I think it whould be better to change it to `shell=False` instead of suppress type error

beancount/utils/encryption.py
```diff
@@ -9,20 +9,22 @@
...
-        return pipe.returncode == 0 and re.match(r"gpg \(GnuPG\) (1\.4|2)\.", version_text)
+        return pipe.returncode == 0 and bool(
+            re.match(r"gpg \(GnuPG\) (1\.4|2)\.", version_text)
+        )
     except OSError:
         return False
 
 
-def is_encrypted_file(filename):
+def is_encrypted_file(filename: str) -> bool:
```
trim21: it also take `pthlib.Path`, which is what we used in tests in some case.

beancount/utils/encryption.py
```diff
@@ -43,7 +45,7 @@ def is_encrypted_file(filename):
     return False
 
 
-def read_encrypted_file(filename):
+def read_encrypted_file(filename: str) -> str:
```
trim21: I think some caller also use a `pathlib.Path` as filename

beancount/core/account_types.py
yagebu: No, there's probably no noticeable runtime overhead. I think it's easier to not decide this on a case by case basis, why take the time to evaluate whether it has any overhead. So as written above I prefer to just follow that lint rule

beancount/parser/booking_method.py
yagebu: hah, interesting that this wasn't flagged by ruff

beancount/ops/summarize.py
yagebu: these ones are used a lot in this file, so I added the explicit imports

beancount/parser/_parser.pyi
yagebu: I didn't modify this line - this PR is not intended to fix all types, I think we can do this incrementally

beancount/parser/booking_method.py
trim21: Is there any ruff rule exists for this case?

beancount/core/account_types.py
trim21: Fair enough. I personally think it add mental overhead of having to distinguish between runtime imports and type-only imports

beancount/loader.py
yagebu: there's various places where stringiness of paths is assumed in the Beancount code base (in the cache hashing for example or when they are added to metadata). I think allowing paths in more places is good but that's a bigger change I don't want to do in this PR

trim21: We are already using `pathlib.Path` it in beancount tests

https://github.com/beancount/beancount/blob/859505420f3b3eaf6ecc99f7aaf2126784164b98/beancount/loader_test.py#L114-L116

I mean it's `filename: str | pathlib.Path`, I'm not suggesting `filename: pathlib.Path`

beancount/core/account_types.py
yagebu: That's why I let ruff distinguish this for me ;)

trim21: > That's why I let ruff distinguish this for me ;)

Never tried it, does ruff know a type-only import become runtime import and move it outside of `TYPE_CHECKING`?

beancount/loader.py
trim21: Even filename is a `pathlib.Path` here, it will be convert to str before we feed it to cache or parser, no need to worry about that.

beancount/core/account_types.py
yagebu: yes, with the `TC` rules

blais:
> we should consider enabled mypy on beancount someday？

Absolutely. Or other variants.

beancount/core/data.py
blais: Yes, this should be public.

blais (approved):
LGTM
I'll leave a few hours in case you want to address additional changes.
Thank you!

beancount/core/data.py
yagebu: This is public in any case but I'll move it outside the TYPE_CHECKING block (and also make it runtime_checkable)

trim21: I'm +1 on public and -1 on `runtime_checkable`, maybe we should defer  `runtime_checkable` and wait for need of `isinstance(err, BeancountError)`

beancount/loader.py
yagebu: > I mean it's filename: str | pathlib.Path, I'm not suggesting filename: pathlib.Path

That's what I understood as well :) But mainly due to this being a bigger change in the API, I'll leave it as is for this PR

trim21: `filename` here is actually will be passed to `is_encrypted_file(filename, ...)` and `load_encrypted_file(filename, ...)`...

beancount/parser/lexer.py
```diff
@@ -5,38 +5,41 @@
...
 
 
 class LexBuilder:
     """A builder used only for building lexer objects."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         # Errors that occurred during lexing and parsing.
-        self.errors = []
+        self.errors: list[BeancountError] = []
```
trim21, line 34: do we need `from __future__ import annotations` in this file？

yagebu: I guess since the older Python builds aren't failing, we don't need it, but I think we should have it in any case, thanks :)

yagebu:
@blais I think this is ready :)

blais:
> @blais I think this is ready :)

Thank you!

blais:
Just for fun I tried asking gemini-exp-1206 to update type annotations on account.py.
Here's what it found: 
https://github.com/beancount/beancount/commit/6980ccef6a1322d9f8db7321f99379aadd13d8e1

yagebu:
> Just for fun I tried asking gemini-exp-1206 to update type annotations on account.py.

Not bad :)

PR #907, opened 2024-12-22
refactor: cleanup some unused old code

trim21:
looks like they are from the era of python2?


blais:
This is great.
Only one thing though: while `idify` should be removed here, removing it will break `beangulp`.
A copy of the util should be migrated to its repo.

PR #910, opened 2024-12-23
enable mypy

trim21:
enable mypy on beancount

1. Add `mypy` and `types-regex` to `types` extra `project.optional-dependencies`
2. add mypy ci job to run mypy checks

this PR add mypy options `check_untyped_defs = false` so it only check function body with arguments/return type checks. untyped function like `def test_a(self):` will not be checked.

use `uv` in ci job there is no way to only install deps without building project from pyproject.toml with pip.

### DIFF
--- a/beancount/core/getters.py
+++ b/beancount/core/getters.py
@@ -10,6 +10,8 @@ __license__ = "GNU GPLv2"
 from collections import OrderedDict
 from collections import defaultdict
 from typing import TYPE_CHECKING
+from typing import Dict
+from typing import Tuple
 
 from beancount.core import account
 from beancount.core.data import Close
@@ -32,7 +34,7 @@ if TYPE_CHECKING:
     from beancount.core.data import Note
     from beancount.core.data import Pad
 
-    AccountsUseMap = tuple[dict[Account, datetime.date], dict[Account, datetime.date]]
+    AccountsUseMap = Tuple[Dict[Account, datetime.date], Dict[Account, datetime.date]]
 
 
 class GetAccounts:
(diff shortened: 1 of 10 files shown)


beancount/core/getters.py
```diff
@@ -32,7 +34,7 @@
     from beancount.core.data import Note
     from beancount.core.data import Pad
 
-    AccountsUseMap = tuple[dict[Account, datetime.date], dict[Account, datetime.date]]
+    AccountsUseMap = Tuple[Dict[Account, datetime.date], Dict[Account, datetime.date]]
```
blais, line 37: Why not keep the new modern types?

blais (approved):
LGTM (except the use of the older types)

beancount/core/getters.py
blais: I'll assume there's a mypy related reason; I'll go ahead and merge this, we can separately modernize all the type annotations after (there are plenty others).

trim21: 1. we target py38 
2. in mypy `typing.TYPE_CHECKING` is always true, so mypy take it as runable code

blais: Let's bump it up to 3.9 and require modern type annotations.
(In general I feel in branch `master` it's fine for us to require more modern Python. 
Even 3.9 is old by now, if 3.12 had something we care about I'd bump the req.)

blais: Oh I wonder if gemini can modernize all the type annotations for me... I'm on an Aider kick...

trim21: I'm still using py310...

blais: AFAIK generics in standard type annotations became possible in 3.9, so you should be alright:
https://peps.python.org/pep-0585/

PR #911, opened 2024-12-23
minor type fix

trim21:
add `pathlib.Path` on `load_file`, which we are already using it in our tests.

remove `@runtime_checkable` from `BeancountError`

Our errors do not actually inherit BeancountError, and `runtime_checkable` has very stanage runtime behavior (in my opinion) and there is also no need to do `isinstance(..., BeancountError)`

```python
from typing import Any
from typing import NamedTuple

from beancount.core.data import BeancountError


class AnySomeOtherNamedTupleWithSameName(NamedTuple):
    source: Any
    message: Any
    entry: Any


assert isinstance(AnySomeOtherNamedTupleWithSameName(1, 2, 3), BeancountError)
```

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -18,7 +18,6 @@ from typing import NamedTuple
 from typing import Protocol
 from typing import Union
 from typing import overload
-from typing import runtime_checkable
 
 from beancount.core.account import has_component
 from beancount.core.amount import Amount
@@ -34,9 +33,8 @@ Flag = str
 Meta = Dict[str, Any]
 
 
-@runtime_checkable
 class BeancountError(Protocol):
-    """Beancount errors are classes with these attributes."""
+    """Beancount errors are objects with these attributes"""
 
     @property
     def source(self) -> Meta: ...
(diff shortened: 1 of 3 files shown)


beancount/core/data.py
```diff
@@ -34,9 +33,8 @@
 Meta = Dict[str, Any]
 
 
-@runtime_checkable
 class BeancountError(Protocol):
-    """Beancount errors are classes with these attributes."""
+    """Beancount errors are namedtuple with these attributes"""
```
yagebu: With this protocol, we don't need them to be namedtuples, right? (so e.g. plugins are free to use plain classes or dataclasses). If you don't like the word "classes", maybe just "objects"?

trim21: good idea

beancount/loader.py
```diff
@@ -35,6 +36,9 @@
 from beancount.utils import encryption
 from beancount.utils import misc_utils
 
+if TYPE_CHECKING:
+    import pathlib
```
blais: `Path` is one of those imports I'd make an exception from the Google import rule for and import as 
`from pathlib import Path` due to its ubiquity

PR #912, opened 2024-12-23
drop py<3.9

trim21:


pyproject.toml
```diff
@@ -70,9 +70,6 @@ line-length = 92
 # See 'rust-format.el' and .dir-locals.el at the root of the project.
 
 [tool.ruff]
```
trim21, line 72: ruff can understand `project.requires-python`

yagebu:
Why drop Python 3.9? I think it's good to support all officially supported Python versions - unless there's something we really need in 3.10+?

trim21:
> Why drop Python 3.9? I think it's good to support all officially supported Python versions - unless there's something we really need in 3.10+?

not really IMO

https://github.com/beancount/beancount/pull/910#discussion_r1895804150

trim21:
> Why drop Python 3.9? I think it's good to support all officially supported Python versions - unless there's something we really need in 3.10+?

OH, sorry. we do not need to drop 39. I'll add 39 back

blais:
I think I've done this an hour ago more or less

PR #915, opened 2024-12-24
Fix typings which broke beanquery

yagebu:
- position.cost should always be a cost (on postings, we need to have CostSpecs before booking)
- all transactions should have a flag

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -260,7 +260,7 @@ class Transaction(NamedTuple):
 
     meta: Meta
     date: datetime.date
-    flag: Flag | None
+    flag: Flag
     payee: str | None
     narration: str | None
     tags: frozenset[str]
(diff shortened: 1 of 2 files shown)


beancount/core/data.py
```diff
@@ -260,7 +260,7 @@ class Transaction(NamedTuple):
 
     meta: Meta
     date: datetime.date
-    flag: Flag | None
+    flag: Flag
```
trim21, line 263: should we also update here?

https://github.com/beancount/beancount/blob/4b0c4af727c8538e370fb813eb5b4c8532d02d43/src/python/beancount/core/data.py#L288

blais: Hmm. I think you can technically construct one programmatically without a flag.
I'm okay with requiring it (even if an empty string).

PR #923, opened 2025-01-02
Use `Optional[T]` instead of `T | None` for Python 3.9 compatibility

dnicolodi:
This converts the spelling to the latter form only for type declarations consumed in beanquery, the others are left as an exercise for someone else.

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -13,6 +13,7 @@ from decimal import Decimal
 from typing import Any
 from typing import Iterator
 from typing import NamedTuple
+from typing import Optional
 from typing import Protocol
 from typing import Union
 from typing import overload
@@ -111,7 +112,7 @@ class Open(NamedTuple):
     date: datetime.date
     account: Account
     currencies: list[Currency]
-    booking: Booking | None
+    booking: Optional[Booking]
 
 
 class Close(NamedTuple):
@@ -198,8 +199,8 @@ class Balance(NamedTuple):
     date: datetime.date
     account: Account
     amount: Amount
-    tolerance: Decimal | None
-    diff_amount: Amount | None
+    tolerance: Optional[Decimal]
+    diff_amount: Optional[Amount]
 
 
 class Posting(NamedTuple):
@@ -228,11 +229,11 @@ class Posting(NamedTuple):
     """
 
     account: Account
-    units: Amount | None
-    cost: Cost | CostSpec | None
-    price: Amount | None
-    flag: Flag | None
-    meta: Meta | None
+    units: Optional[Amount]
+    cost: Optional[Union[Cost, CostSpec]]
+    price: Optional[Amount]
+    flag: Optional[Flag]
+    meta: Optional[Meta]
 
 
 class Transaction(NamedTuple):
@@ -261,8 +262,8 @@ class Transaction(NamedTuple):
     meta: Meta
     date: datetime.date
     flag: Flag
-    payee: str | None
-    narration: str | None
+    payee: Optional[str]
+    narration: Optional[str]
     tags: frozenset[str]
     links: frozenset[str]
     postings: list[Posting]
@@ -305,8 +306,8 @@ class Note(NamedTuple):
     date: datetime.date
     account: Account
     comment: str
-    tags: frozenset[str] | None
-    links: frozenset[str] | None
+    tags: Optional[frozenset[str]]
+    links: Optional[frozenset[str]]
 
 
 class Event(NamedTuple):
@@ -419,8 +420,8 @@ class Document(NamedTuple):
     date: datetime.date
     account: Account
     filename: str
-    tags: frozenset[str] | None
-    links: frozenset[str] | None
+    tags: Optional[frozenset[str]]
+    links: Optional[frozenset[str]]
 
 
 class Custom(NamedTuple):
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -7,10 +7,6 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        os: [ ubuntu-latest ]
-      fail-fast: false
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -34,7 +30,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: '3.12'
 
       # pip doesn't support this
       - uses: astral-sh/setup-uv@v4
--- a/beancount/core/data_test.py
+++ b/beancount/core/data_test.py
@@ -3,6 +3,7 @@ __license__ = "GNU GPLv2"
 
 import datetime
 import pickle
+import typing
 import unittest
 from datetime import date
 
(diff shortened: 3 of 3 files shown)


trim21:
We have `from __future__ import annotations` in data.py, what compatibility issue are you referring to?

Is this because some 3rd package like pydantic can't infer type correctly with `__future__.annotations` ?

dnicolodi:
> We have `from __future__ import annotations` in data.py, what compatibility issue are you referring to?

`from __future__ import annotations` only defers the evaluation of the annotation to the time at which they are accessed. Annotations still needs to be valid expressions. The `T | None` expression is not valid in Python 3.9. Anything that wants to make use of the annotations, including `mypy`, while using Python 3.9 cannot do it:
```python
>>> import typing
>>> import beancount.core.data
>>> typing.get_type_hints(beancount.core.data.Open)
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.9/3.9.19_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 1459, in get_type_hints
    value = _eval_type(value, base_globals, localns)
  File "/usr/local/Cellar/python@3.9/3.9.19_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 292, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/usr/local/Cellar/python@3.9/3.9.19_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 554, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'EnumMeta' and 'NoneType'
```

`beanquery` uses the annotations to derive columns types and as demonstrated above, they are now not valid anymore. I don't really understand what is the reason to rewrite the annotations using the new syntax, when the old annotations were perfectly fine. Also, I don't understand why the annotation style has not been kept consistent: now some modules use `T | None` and other use `Optional[T]`.

blais:
Reason was to modernize. I prefer to let go of old baggage in the latest version.

Lets:
1. Bump the required version to 3.10
2. Fix the remaining old annotations

dnicolodi:
Modernize what exactly? The functionality is identical. This is only automated code churn that is only changing syntax for the sake of change. We have been holding back useful changes for years in the name of code stability but now we are breaking compatibility for having typing annotations that look slightly nicer. I don't get it.

blais:
Stability was for v2. In branch master we want to be able to move forward no?. The python community is making this change broadly. 

I'll grant that it's not a very important change. 

I don't know. I'll take another look later (stuck at through away from my box).

dnicolodi:
> The python community is making this change broadly.

Which change?  The change I'm questioning is using `str | None` instead of `Optional[str]`. The two have the same identical meaning. Except that the latter what we used till last week and is compatible with Python 3.9 and the former is not. Even if we were to drop Python 3.9 compatibility, I would be against applying the mechanical change to code that is not otherwise touched: it is just noise in the diffs.

As I wrote a few times: using the latter form does not bring any advantage and breaks compatibility with a supported version of Python. I don't have anything against embracing Python functionality introduced in (reasonably) recent version of Python if it makes the code better or it enables functionality, but here we are talking about change that has been introduced only because who wrote the patches didn't bother checking that the code they wrote was compatible with the Python baseline.

blais:
I understand what you're saying. 

At the same time there are two equivalent syntaxes, an old one and a new one. The old one is pretty old by now so it's not unreasonable that some people want to move on to the new one, which requires dropping an old version of Python. 

It's not clear to me there's a clear best here, you can argue it both ways. If we stick with the old we have to tell new contributions to use the old syntax (and the automated test will do that too). But then our code looks old. Sure, maybe not a big deal. If we move to the new we have to drop an old but still supported version. Old version users will have to update.

I think open source should also be some significant degree of fun, and people who are excited with contributing love to love the newest of features (though we have yet to see a match statement in beancount). 

Unclear to me how to call this one. I lean towards the new.

blais:
Another consideration: moving to the new type annotations would make it more difficult to merge the latest changes into v3 (stable) without also bumping its required version. (That's an argument for the conservative side.)

dnicolodi:
I would agree with the "development should be fun" and "new language features are more fun to use" argument, if the changes I am objecting to were part of a larger contribution. But the contribution we are talking about is the change to the new syntax in itself. I may be to old school, but committing the result of `ruff check --fix` is not my kind of fun, and neither it is to have to go around fixing the breakage caused by too eagerly applying the recommendations (misconfigured) linters.

beancount/core/data.py
```diff
@@ -228,11 +229,11 @@ class Posting(NamedTuple):
     """
 
     account: Account
-    units: Amount | None
-    cost: Cost | CostSpec | None
-    price: Amount | None
-    flag: Flag | None
-    meta: Meta | None
+    units: Optional[Amount]
+    cost: Cost | Optional[CostSpec]
```
yagebu: this needs a `Union` as well

yagebu:
In my opinion two things are mixed up in this discussion. For type-checking with mypy, the newer syntax (when used with the annotations future import) is compatible with Python 3.9 (see https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-x-y-syntax-for-unions). Due to `if TYPE_CHECKING` blocks (to avoid unnecessary runtime overhead / import cycles / ...) I'd generally expect that some annotations can not be successfully evaluated at runtime and they don't need to be for the purposes of type-checking. The annotations weren't even marked as exported in 3.0 with a py.typed marker so I'd consider them an internal implementation detail so far.

Having the annotations be available for beanquery at runtime should still be supported. OTOH, I don't think that has to mean that we need to avoid the newer typing syntax for the rest of Beancount. Overall, I don't see this as a strong enough reason to drop support for 3.9 in any case.

AFAICT runtime availability of type annotations wasn't documented so far or tested for, so I'd assume this might break again in the future. @dnicolodi: Can you add a test that checks that the annotations that are used in beanquery can be evaluated at runtime? And maybe include the further assumption the beanquery logic takes that all the fields have just one possible type (to avoid the sort of breakage I fixed in #915)

trim21:
maybe we should pull beanquery trunk in ci beancount and run some tests in this repo?

dnicolodi:
> And maybe include the further assumption the beanquery logic takes that all the fields have just one possible type (to avoid the sort of breakage I fixed in #915)

I don't understand. beanquery assumes that some fields have just one possible type because that is true in practice. The patches in #915 are fixes to the type annotations regardless of the fact that you found them because beanquery assumptions. What you are asking is to add a test to verify that incorrect type annotations are not introduced. I don't know how to do that.

dnicolodi:
I think that the issue has been a bit misunderstood. It does not help that when I first reported it I was a bit pissed. What I am asking for is to go back to the old annotation syntax for a handful of annotations that it makes sense to consume at runtime.

As the old syntax is used in other parts of the codebase, I don't see a big problem in doing so. I agree that it does not make sense to restrict ourselves to the old syntax for annotations that are not consumed at runtime.

I added a test that verifies that the annotations that matter are correctly interpreted at runtime.

dnicolodi:
> maybe we should pull beanquery trunk in ci beancount and run some tests in this repo?

That is overkill. I added a test for this specific problem and I believe that any problem will anyhow be caught early enough. The beanquery CI jobs already run on the git master version of beancount and caught this issue very early.

blais:
@dnicolodi  Fine with delaying the conversion to the newer type annotation syntax.
We can do this later anyway.

Is this PR in the state you want it?

dnicolodi:
Yes, it should be everything that's needed. Thanks @blais

gnaish:
@blais  I am running into this issue after upgraded beancount from v2 to v3. It breaks beanquery and the statistics.html page of fava. I am a mac user, which xcode command line provides python3.9 only. Could you please have a new release to include this change?

blais:
Can't for a week. On the road 
Can look after

PR #926, opened 2025-01-02
build: make meson build_ext target optional

trim21:
looks like meson are being "too smart" to validate every command, for our `build_ext` target it will check if external program `cp` exists, which will fail on windows without git-bash or binutils.

Let's check `cp` with `find_program` and skip this target if it doesn't exists.

https://github.com/beancount/beancount/discussions/921#discussioncomment-11716559

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -254,18 +254,22 @@ parser = py.extension_module(
 # meson setup build
 # ninja -C build build_ext
 # ```
-custom_target(
-    'build extension and copy back to source tree',
-    input: parser,
-    output: 'build_ext',
-    depends: parser,
-    command: [
-        'cp',
-        parser.full_path(),
-        join_paths(meson.project_source_root(), 'beancount/parser/'),
-    ],
-    build_by_default: false,
-)
+
+cp = find_program('cp', required: false)
+if cp.found()
+    custom_target(
+        'build extension and copy back to source tree',
+        input: parser,
+        output: 'build_ext',
+        depends: parser,
+        command: [
+            cp,
+            parser.full_path(),
+            join_paths(meson.project_source_root(), 'beancount/parser/'),
+        ],
+        build_by_default: false,
+    )
+endif
 
 if get_option('tests').enabled()
     cc = meson.get_compiler('c')


blais:
Could we avoid the dependency by using shutil.copyfile() instead?

marvhen:
> Could we avoid the dependency by using shutil.copyfile() instead?

Sounds like the correct approach to me, as it would work across all supported platforms.

Although, there are [some caveats](https://docs.python.org/3/library/shutil.html), e.g. ACL's, file owners, and alternate data streams not getting copied.

dnicolodi:
Or we could embrace modern development practices and get rid of this. We are dropping support for old Python versions, thus I don't see the need to keep supporting the prehistoric way of doing things.

dnicolodi:
Also, I was under the impression that the Meson custom target was added to support Windows systems where installing make is cumbersome. However, it seems that this does not work on most Windows installation and actually causes trouble on that platform.

trim21:
> Could we avoid the dependency by using shutil.copyfile() instead?

good idea.

blais:
> Or we could embrace modern development practices and get rid of this. We are dropping support for old Python versions, thus I don't see the need to keep supporting the prehistoric way of doing things.

I'm not sure I follow. Explain your favorite way. 

If you mean always using uv and a virtual environment that's a tad annoying, though it's fast now. I really like to be able to just use the source from the directory without a build step. It's instant. Save, run. 

What do you recommend?

trim21:
> > Or we could embrace modern development practices and get rid of this. We are dropping support for old Python versions, thus I don't see the need to keep supporting the prehistoric way of doing things.
> 
> I'm not sure I follow. Explain your favorite way.
> 
> If you mean always using uv and a virtual environment that's a tad annoying, though it's fast now. I really like to be able to just use the source from the directory without a build step. It's instant. Save, run.
> 
> What do you recommend?

meson actually has support to build extension with `pip install --no-build-isolation -e .`, it will generate a python script as pth to load binary extension. But I really don't like the idea of hijacking python import system, so even I'm using virtual env I still copy binary extension back to source tree.

blais:
So I still have to "build". I think there's a big step between building and not building. Given the extension module hardly ever changes the convenience of the old skool way is still a nice to have.

Also nothing precludes us from using either way.

trim21:
> So I still have to "build".

You don't need. pth script will build extension on the fly.


```text
_______________________________________________ ERROR collecting beancount/utils/file_utils_test.py ________________________________________________
ImportError while importing test module 'C:\Users\Trim21\proj\beancount\beancount\utils\file_utils_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
.venv\lib\site-packages\_beancount_editable_loader.py:345: in _rebuild
    subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)
..\..\scoop\apps\python39\current\lib\subprocess.py:528: in run
    raise CalledProcessError(retcode, process.args,
E   subprocess.CalledProcessError: Command '['meson', 'compile']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:
..\..\scoop\apps\python39\current\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
.venv\lib\site-packages\_beancount_editable_loader.py:311: in find_spec
    tree = self._rebuild()
.venv\lib\site-packages\_beancount_editable_loader.py:347: in _rebuild
    raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
E   ImportError: re-building the beancount meson-python editable wheel package failed
```

PR #929, opened 2025-01-03
add back DefaultDictWithKey for beangulp compatibility

yagebu:
This is still used in the latest released version of beangulp so
beancount master is incompatible with that. This is just a couple of
lines of well-aged code, so I do not think there is a rush to remove
it.

### DIFF
--- a/beancount/utils/defdict.py
+++ b/beancount/utils/defdict.py
@@ -7,6 +7,20 @@ adding this in eventually.
 __copyright__ = "Copyright (C) 2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
+import collections
+
+
+# Deprecated and only kept around for compatibility with beangulp 0.1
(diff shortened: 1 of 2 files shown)


blais:
Could you move the code to beangulp itself?

yagebu:
> Could you move the code to beangulp itself?

That's happened already, but no release of beangulp has happened since then, leading to errors when only beancount is installed from git, see https://github.com/beancount/fava/issues/1860#issuecomment-2568961222

trim21:
make sence. If we release a new version of beancount it will break all users use that with old version of beangulp.

trim21:
also I think we should have some upper boundary on beancount in beangulp? At least `beancount>=2.3.5,<4`

blais:
Let's move forward and add to beangulp and release. I think @dnicolodi  already did that.

blais:
Won't add back.

Issue #194, opened 2017-09-16
Use of "@@" syntax (was: Use of float numbers in Amount)


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/194) by Zhuoyun Wei (Bitbucket: [wzyboy](https://bitbucket.org/wzyboy), GitHub: [wzyboy](https://github.com/wzyboy)).**

----------------------------------------

Hi,

I have always been bugged by the fact that in Fava, in almost all
transactions involving currency conversions (that is, every time I pay
with international credit cards), very long fractional
numbers are displayed in postings.

I always thought it's something about DisplayContext and did not dig
into the reasons. Today I am playing with BQL and finally it occurs to
me that float numbers, instead of Decimals, are used somewhere
internally in Beancount.

Observe the float round-off errors:

```
beancount> select date, narration, position, weight where account ~ "Expenses:" and narration ~ "Amazon web services";
   date         narration      position                weight
---------- ------------------- --------- ----------------------------------
[...]
2016-07-04 Amazon web services 12.06 USD  80.53000000000000000000000001 CNY
[...]
2017-05-05 Amazon web services 13.04 USD  90.13999999999999999999999999 CNY
```

I can see that in Beancount's docs, Martin always uses "@" syntax to
record currency conversions, like "10 USD @ 1.3 CAD". I, however,
always uses "@@" syntax to record conversions, like "10 USD @@ 65 CNY".
Because on my bank statements there are no currency exchange rates, and
only the numbers in foreign currency and local currency are provided.

Here is how the two transactions in the bean-query examples above look
like in the Beancount input files:

```
2016-07-04 * "Amazon web services"
  Liabilities:CMB:CreditCards                 -80.53 CNY
  Expenses:Communications:Internet            +12.06 USD @@ 80.53 CNY

2017-05-05 * "Amazon web services"
  Liabilities:CMB:CreditCards                 -90.14 CNY
  Expenses:Communications:Internet            +13.04 USD @@ 90.14 CNY
```

It seems that Beancoun did not use these numbers directly. Instead, it
uses the quotient of the two numbers to calculate the weight. In this
process, float round-off errors occur.

blais:
**Original comment by Zhuoyun Wei (Bitbucket: [wzyboy](https://bitbucket.org/wzyboy), GitHub: [wzyboy](https://github.com/wzyboy)).**

----------------------------------------

Hmm, sorry for my misunderstanding. This may not be relevant to float round-off errors:

```
[...]

PR #934, opened 2025-01-23
fix: create_simple_posting_with_cost produces 0001-01-01 as date

wzyboy:
After upgrading from 3.0.0 to 3.1.0, I find that my scripts which uses `create_simple_posting_with_cost()` function now produces lots with `0001-01-01` as date.

It turns out that commit 0f8798110 replaces the `None` value for date with hard-coded `date(1, 1, 1)`.

This PR fixes the issue by replacing `Cost` with `CostSpec`, which accepts `Optional[datetime.date]` as date.

### DIFF
--- a/beancount/core/convert.py
+++ b/beancount/core/convert.py
@@ -28,6 +28,7 @@ from beancount.core import prices
 from beancount.core.amount import Amount
 from beancount.core.number import MISSING
 from beancount.core.position import Cost
+from beancount.core.position import CostSpec
 from beancount.core.position import Position
 
 
@@ -88,6 +89,9 @@ def get_weight(pos):
     # It the object has a cost, use that as the weight, to balance.
     if isinstance(cost, Cost) and isinstance(cost.number, Decimal):
         weight = Amount(cost.number * pos.units.number, cost.currency)
+    # CostSpec is also an acceptable source of cost
+    elif isinstance(cost, CostSpec) and isinstance(cost.number_per, Decimal):
+        weight = Amount(cost.number_per * pos.units.number, cost.currency)
     else:
         # Otherwise use the postings.
         weight = units
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -590,7 +590,7 @@ def create_simple_posting_with_cost(
     if cost_number is not None and not isinstance(cost_number, Decimal):
         cost_number = D(cost_number)
     units = Amount(number, currency)
-    cost = Cost(cost_number, cost_currency, datetime.date(1, 1, 1), None)
+    cost = CostSpec(cost_number, None, cost_currency, None, None, None)
     posting = Posting(account, units, cost, None, None, None)
     if entry is not None:
         entry.postings.append(posting)


blais:
Oh shoot I missed that during review. That date(1,1,1) is wrong.
Your PR attempts to use CostSpec, which is not valid.
I have to do something else.

wzyboy:
I'm curious why `CostSpec` is not valid here? My understanding is that `CostSpec` is almost the same as `Cost` but allows `None` as some of the values. For the use in `create_simple_posting_with_cost()`, the user cannot customize the additional arguments provided by `CostSpec` anyway...

blais:
CostSpec is only allowed temporarily as output from the parser. They get converted to Cost in the booking code.

In the C++ or Rust version this won't be allowed anyway, the data types won't allow it, it's only because this is leveraging duck-typing in Python that there's ambiguity here.

wzyboy:
I use `create_simple_posting_with_cost()` function in my beangulp extraction scripts. Previously, it produces text like this:

```
2024-12-20 * "Exchange in: Global Innovators Class, Greater Canada"
  Assets:Fidelity:TFSA:Cash        -250.00 CAD
  Assets:Fidelity:TFSA:Investment   3.0836 FID5973 {40.5364 CAD}
  Assets:Fidelity:TFSA:Investment   4.5413 FID1246 {27.5249 CAD}
```

With you fix of using the transaction's date, it produces text like this:

```
2024-12-20 * "Exchange in: Global Innovators Class, Greater Canada"
  Assets:Fidelity:TFSA:Cash        -250.00 CAD
  Assets:Fidelity:TFSA:Investment   3.0836 FID5973 {40.5364 CAD, 2024-12-20}
  Assets:Fidelity:TFSA:Investment   4.5413 FID1246 {27.5249 CAD, 2024-12-20}
```

It does not break anything serious, but just a bit more verbose.

Is manually constructing a `CostSpec` better fitting my use case?

blais:
Oh I see, your purpose is to generate text. The answer isn't entirely straightforward.

The data structures have to be redesigned to make the distinction between parser output (i.e., the AST) and finalized interpolated and booked data structures unambiguous, and that will be done in a typed language. I don't want to bother doing that with the Python version--it would be a waste of time (a lot of things would have to be reviewed); this has been done in the Proto + C++ version that has been abandoned. I'm planning to do that in the Rust version. Moving to a typed language will force and that's the right way to do this IMHO. This ambiguity is a major problem, but in practice only has crept up in a few use cases.

The current printer is AFAICR designed to render finalized data structures, i.e., Cost not CostSpec.
It is possible that it works with CostSpec, but I just looked in the source and that looks incidental.
It does have some limited support for missing numbers though.
It could be made more explicitly to handle both cases, I don't see why not.

I think creating a posting programmatically for importers is intended to produce a finalized transaction, not an AST. (In fact, because of that, it should enforce the presence of non-optional fields; I don't think we validate the data structure too closely in beangulp? Don't remember.) Given that, I'm not too sure I have a great solution for you: a finalized transaction has postings with non-optional dates on it, it would be an error to produce a Cost without a date (and some things might break as a result).

I suppose we should review beangulp and switch to assuming the importers produce only AST data and validate that (all CostSpec, never Cost, allow missing data, etc.). That would be a major change. I don't want to do this, it'll break everyone. 

I'd much rather fix all of this up in Rust in the next version.

wzyboy:
> I think creating a posting programmatically for importers is intended to produce a finalized transaction, not an AST. (In fact, because of that, it should enforce the presence of non-optional fields

I'm not sure I follow that. My understanding is that the text produced by beangulp is intended to be piped into a text file, thus it should be able to produce an **input** format, just like manually typed b/y a huamn user. If we enforce it produce finalized text only, we can no longer use it to generate **incomplete** transactions.

Some of my beangulp scripts intentionally generate incomplete transactions so that I edit them manually. One use case is to deal with foreign transactions. My credit card provider does not provide the FX rate in the statement, instead they only tell me about the total of the foreign currency. As Beancount [does not support total price (`@@`) internally](https://github.com/beancount/beancount/issues/194#issuecomment-632817024), I cannot produce this text programtically:

```
2025-01-01 * "AWS"
  Liabilities:CreditCard    -43.00 CAD
  Expenses:Web               30.00 USD @@ 43.00 CAD
```

I can only calculate the FX rate in the script and produce this:

```
2025-01-01 * "AWS"
  Liabilities:CreditCard    -43.00 CAD
  Expenses:Web               30.00 USD @ 1.43333333333333333 CAD
```

But this is ugly to look at. So I ended up producing an **incomplete/invalid** transaction like this:

```
2025-01-01 * "AWS"
  Liabilities:CreditCard    -43.00 CAD
  Expenses:Web               30.00 USD
```

This of course shows up as an error in my editor, and I just append an `@` symbol to make a valid transaction:

```
2025-01-01 * "AWS"
  Liabilities:CreditCard    -43.00 CAD
  Expenses:Web               30.00 USD @
```

There are some other use cases where I intentionally produce incomplete transactions and then fix them up manually. My point is, importers should have the ability to produce **incomplete** transactions.

blais:
Yes, in a newer, revamped world (in Rust), you'll have very distinct data structures for each of these things. The AST will support that, and the finalized schema won't, it'll be as expected. I don't want to do a bunch of work on the Python version since this will be done way better in a typed schema (probably in Rust), it would be a waste of time. I understand it's less than ideal right now.

wzyboy:
> As Beancount does not support total price (`@@`) internally, I cannot produce this text programtically:

I posted a dirty solution for this: https://github.com/beancount/beancount/issues/194#issuecomment-2726240922

PR #937, opened 2025-02-06
Update account.py

QunixHacker:
Support Chinese Unicode As ACC_NAME

### DIFF
--- a/beancount/core/account.py
+++ b/beancount/core/account.py
@@ -35,7 +35,7 @@ sep = ":"
 #   L: All letters.
 #   Nd: Decimal numbers.
 ACC_COMP_TYPE_RE = r"[\p{Lu}][\p{L}\p{Nd}\-]*"
-ACC_COMP_NAME_RE = r"[\p{Lu}\p{Nd}][\p{L}\p{Nd}\-]*"
+ACC_COMP_NAME_RE = r"[\p{Lu}\p{Nd}]\p{Han}][\p{L}\p{Nd}\p{Han}\-]*"
 
 # Regular expression string that matches a valid account. {5672c7270e1e}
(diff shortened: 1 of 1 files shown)


QunixHacker:
This PR Maybe is not the best solution to solve locale/i18n AccountName

Dear blais, if you have purpose to do something with i18n/locale AccountName or AccountType； I Would like do something and ask you to permit me to be a committer for this project.

I will do some work like 
1. analysis diffrent language(eg. Chinese\JP) usage

beancount/core/account.py
```diff
@@ -35,7 +35,7 @@
 #   L: All letters.
 #   Nd: Decimal numbers.
 ACC_COMP_TYPE_RE = r"[\p{Lu}][\p{L}\p{Nd}\-]*"
-ACC_COMP_NAME_RE = r"[\p{Lu}\p{Nd}][\p{L}\p{Nd}\-]*"
+ACC_COMP_NAME_RE = r"[\p{Lu}\p{Nd}]\p{Han}][\p{L}\p{Nd}\p{Han}\-]*"
```
dnicolodi, line 38: From my limited understanding of Unicode properties and scripts not using the Latin alphabet, I think adding the Han character class to the regular expression is correct. However, there is a syntax error in the regular expression: there are two closing brackets `]` but just one opening one `[`. I infer that this change has not been tested. Also, this changes the regular expression for the account name components after the first, but it does not change the one for the leading component (`ACC_COMP_TYPE_RE` vs `ACC_COMP_NAME_RE`). I think both need to be changed in a similar way.

dnicolodi (changes requested):
The idea looks good, however the implementation contains a mistake. We would also need a better commit message describing what the commit does in the first sentence. The current commit message also refers to `ACC_NAME` which is an entity that does not exist. A test that verifies that things work as expected would also be nice.

PR #952, opened 2025-04-03
Support total prices in parsing and printing.

sarg:
Given such bank statement:
```csv
"Date","Payee","Account number","Transaction type","Payment reference","Amount (EUR)","Amount (Foreign Currency)","Type Foreign Currency"
"2022-08-08","SCANDIC GDANSK - RECE","","MasterCard Zahlung","-","-255.83","-1201.8","PLN"
```

I want to see such beancount transaction:
```
2022-08-08 * "SCANDIC GDANSK - RECE"
  Assets:Card:N26   -1201.8 PLN @@ 255.83 EUR
  Expenses:Travel:Hotel                        
```

Both prices in the entry should come directly from the CSV, however currently the importer has to calculate unit price and `-1201.8 PLN @ 0.2128723581 USD` would be printed.

### DIFF
--- a/beancount/core/amount.py
+++ b/beancount/core/amount.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 __copyright__ = "Copyright (C) 2013-2022, 2024  Martin Blais"
 __license__ = "GNU GPLv2"
 
+import builtins
 import re
 from decimal import Decimal
 from typing import TYPE_CHECKING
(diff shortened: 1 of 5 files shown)


blais:
Sorry but we're not changing the schema, would require a lot more changes than this.

sarg:
@blais what schema changes do you mean? `@@` syntax exists for ages. This PR intends to fix a long standing bug that `@@` can be only input manually, but not produced by an ingest script.

blais:
That's a longer discussion and the mailing-list should have detailed long threads on the topic, but basically Beancount's internal design is a little flawed in that the AST and the finalized data structures are shared; they really should be separate. In order to generate the total syntax you'd have to use the AST, and the current AST doesn't support it.  The prototype in protobuf from the cpp branch has a version that does that, and the code to generate incomplete data that matches the input would create that AST data structure (which I dub "inter" in the code, but really should rename to "ast"), not the finalized one.

Point is, if we make this change, a bunch of other code might not handle the change.

Issue #922, opened 2025-01-02
Use of `type | None` typing annotation is not supported in Python 3.9


dnicolodi:
`pyproject.toml` still states compatibility with Python 3.7 and later. However, testing with Python 3.7 and 3.8 has been removed from the CI matrix thus I assume the new baseline is Python 3.9.

It seems that the baseline has been bumped to Python 3.9 to be able to subscript base types. However, typing annotations using the `type | None` the code has been added. These are supported only starting with Python 3.10.

I don't think that dropping compatibility with older Python versions only to be able to spell typing annotations in a slightly easier form is worth the trouble caused to projects depending on Beancount, especially if the typing annotations are not used nor checked: if they were the issue with the Python 3.9 compatibility would have been caught.

josephw:
> if the typing annotations are not used nor checked

0f879811050dfee2dae2dae41 introduces `mypy` to CI, but running with Python 3.12. This is good, but would not catch an issue with an older but supported Python (such as 3.9).

Seems like this still requires _either_ a wider range of `python-version` in the `mypy` job, or a simple declaration that these types won't be compatible with older Pythons.

blais:
@kerrickstaley

kerrickstaley:
@blais my preference here would be to drop 3.9 support. Python 3.10 came out almost 4 years ago, which feels like an eternity to me :) For comparison, [the Numpy project recommends that scientific Python projects drop support for Python versions that are over 42 months (3.5 years) old](https://numpy.org/neps/nep-0029-deprecation_policy.html). Beancount would be consistent with this policy if it dropped support for 3.9 *and* 3.10.

Do you have an opinion? I can create a PR to bump the minimum Python version to 3.10.

PR #967, opened 2025-07-04
Fix typing on Transaction.flag

kerrickstaley:
This field is optional, mark it as such.

### DIFF
--- a/beancount/core/data.py
+++ b/beancount/core/data.py
@@ -261,7 +261,7 @@ class Transaction(NamedTuple):
 
     meta: Meta
     date: datetime.date
-    flag: Flag
+    flag: Flag | None
     payee: Optional[str]
     narration: Optional[str]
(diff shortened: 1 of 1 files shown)


josephw:
This PR regresses #922 due to the use of `type | None` over `Optional`. 

Could be simpler to make 3.10 mandatory than to change this.

dnicolodi:
> Could be simpler to make 3.10 mandatory than to change this.

Are you saying that changing a single line from
```python
   flag: Flag | None
```
to
```python
   flag: Optional[Flag]
```
is m,ore complex than dropping support for a Python version in Beancount and in all projects depending on Beancount?

I disagree.

blais:
More generally it's true that the friction of these tickets is costing us. It's annoying.
This isn't the first ticket caused by 3.9

blais:
it's done

PR #978, opened 2025-10-12
Fix parser.printer Balance tolerance condition

NReilingh:
Dear reviewer:

I am new to this project and have not tested this change in context, so if it is not immediately obvious why the change is correct, please feel free to send this back for me to (figure out how to) write tests.

`entry.tolerance` can be specified as `decimal.Decimal(0)`, and this should be printed but the current condition treats it as falsy. This fix changes the condition to explicitly check for `None`, which I assume will result in the correct behavior given that `tolerance` is expected to be `Optional[decimal.Decimal]`.

Please see https://github.com/beancount/beangulp/issues/192 for further context (i.e. why 0 tolerance should be printed).

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -321,7 +321,7 @@ class EntryPrinter:
 
         # Render optional tolerance.
         tolerance = ""
-        if entry.tolerance:
+        if entry.tolerance is not None:
             tolerance_fmt = self.dformat.format(entry.tolerance, entry.amount.currency)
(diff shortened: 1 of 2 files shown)


blais:
Looks right, but no tests, no action.

NReilingh:
@blais Copy that. Force pushed a prefix commit to this branch which adds a failing test to repro the issue. Checked that the tests defined in the GH Actions `tests` workflow pass when run in my local environment.

PR #980, opened 2025-11-17
lifetimes: Fix duplicate Friday records when closing positions on Friday

val3344:
Skip invalid date intervals where date_begin >= date_end to prevent generating redundant price records, particularly for positions closed on Fridays.

The issue occurred because:
- When closing a position on Friday, date_begin and date_end could be equal
- The weekday adjustment code would keep the date as Friday
- This resulted in duplicate Friday records in the price requirements

By skipping intervals where date_begin >= date_end, we ensure only valid trading periods are considered for daily price lookups.


blais:
Can you add a test?

val3344:
> Can you add a test?

Added tests in e8c544d. Please take a look, thanks!

Issue #989, opened 2025-12-15
Also in v3, documents from a directory option does not work as expected from an include


mlell:
In #596, it was projected that Beancount v3 will fix the problem that the "directory" option does not work if it is located in a file which is read in by inclusion:

> In the current implementation, the include directive is not to be intended as a literal inclusion of the content of the included file in the source file (like the C pre-processor #include) but as a way to merge the interpreted content of the files. This results in the surprising behavior reported here and other limitations. I think there is agreement that the include directive should be revisited to be a literal copy in place of the included file. This will likely be implemented in v3.

The issue is closed but the problem persists, so I thought I'd create a new one.

Currently, it is still the case that `option "documents" "/path/to/documents"` must be written in the **main** beancount file, otherwise it does not work. This behaviour is not documented.

Issue #691, opened 2022-01-04
tag in multiple file?


cnut1648:
Hi, my workflow is to have 1 `.bean` file for each year, for example I would have `2021.bean` and `2022.bean`.

And in the `2021.bean` I have pushed a tag `pushtag #2021-Winter-Beak`, and would like to pop the tag in `2022.bean`. However it seems that beancount doesn't recognize this tag across multiple file. And I had the warning as below 
```
Unbalanced pushed tag: '2021-Winter-Break'
Attempting to pop absent tag: '2021-Winter-Break'
```

I am wondering what is the best workaround for this issue? Thanks!

dnicolodi:
This is a known issue caused by how Beancount implements the `import` directive. There is no other workaround that have a `poptag` at the end of `2021.bean` and a `pushtag` at the beginning of `2022.bean`. An implementation of an `import` directive with a more intuitive semantic has been proposed in PR #657 and #660 but it has not been reviewed yet.

Issue #268, opened 2018-04-03
'pushtag' does not work with include


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/268) by Carbo Kuo (Bitbucket: [byvoid](https://bitbucket.org/byvoid), GitHub: [byvoid](https://github.com/byvoid)).**

----------------------------------------

For example:

```
#!yaml

pushtag #tag1
include 'another_file'
poptag #tag1
```

Expected behavior: all transactions in 'another_file' are tagged with #tag1

Actual behavior: no transaction is in 'another_file' tagged

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

That's a known design characteristic.
Includes aren't parsed in order.
All the directives are accumulated, then sorted, then processed for booking.

I'm not sure whether to close this, as I can see how one would assume this works, and it would be useful to have it work. I'll promote this to a "proposal" and keep it here.

PR #990, opened 2025-12-15
Document that option 'documents' must be included in main file

mlell:
... as a short-term workaround for #989

### DIFF
--- a/beancount/parser/options.py
+++ b/beancount/parser/options.py
@@ -580,7 +580,10 @@ PUBLIC_OPTION_GROUPS = [
         """
       A list of directory roots, relative to the CWD, which should be searched
       for document files. For the document files to be automatically found they
-      must have the following filename format: YYYY-MM-DD.(.*)
+      must have the following filename format: YYYY-MM-DD.(.*). Note that
+      currently, this option can not be specified in a beancount file which is
+      read using the `include` statement, it must be given in the **main**
+      beancount file.
     """,
         [Opt("documents", [], "/path/to/your/documents/archive")],
     ),


beancount/parser/options.py
```diff
@@ -580,7 +580,10 @@ def Opt(
         """
       A list of directory roots, relative to the CWD, which should be searched
       for document files. For the document files to be automatically found they
-      must have the following filename format: YYYY-MM-DD.(.*)
+      must have the following filename format: YYYY-MM-DD.(.*). Note that
+      currently, this option can not be specified in a beancount file which is
+      read using the `include` statement, it must be given in the **main**
+      beancount file.
```
dnicolodi, line 586: This is not true. The `documents` option is effective for the entries read from the file where it is specified. As all other options. Therefore, I don't see the need to specify this behavior only for the `documents` option.

dnicolodi (changes requested):

beancount/parser/options.py
mlell: I see. Also, I have found it now in the general documentation about [includes](https://beancount.github.io/docs/beancount_language_syntax.html#includes). So it is documented, yet, it seems to confuse people regulary (#268, #653, #691, and more that are referenced in the respective issue's answers). So I wonder, is there maybe a way to document it differently so that less people like me miss it? Maybe add a genereral reminder to the options page: 

>  Keep in mind that options affect only the entries of the file where it is specified

blais: I always welcome more documentation, thank you.
Maybe we should add a section in the google docs for the behavior of includes.

(Somebody else needed includes when it was implemented, I hardly use it myself, so I think it's a little underdocumented. We should fix that.)

NReilingh: > The `documents` option is effective for the entries read from the file where it is specified.

@dnicolodi The `documents` option controls behavior that synthesizes Document directives post-parse, no? How can it have an effect on any entries being read from a beancount file, whether include or parent?

I believe the key here is that since the [`process_documents` plugin](https://github.com/beancount/beancount/blob/eb8be17c750257dc693e3173c30d2cabc9ab2faa/beancount/ops/documents.py#L35) runs post-parse, it only receives the options map returned for the top-level file. Users are expecting the specification of `option "documents"` in an included file to affect the workings of `document:` metadata attributes on transactions read from that same file, but in reality the metadata attributes just need the appropriate Document directive to be defined explicitly somewhere _or_ synthesized in the post-parse `process_documents` code. Since the `documents` option in an included file never makes it to the post-parse plugin, the user's expectations are broken.

PR #997, opened 2026-01-08
ci: run pre-commit hooks in lint ci instead of latest ruff

trim21:
We now have 2 ruff in projects, one is in pre-commit hooks, with pinned version, and one is in ci, without any version specification.

So when someone running pre-commit, the files got formatted by a old version of ruff.

It should be more proper that we also use pre-commit hooks to do the linting in CI.

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -11,10 +11,36 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.13'
-      - uses: astral-sh/setup-uv@v5
-      - run: pip install ruff
-      - run: make lint
+          python-version: "3.13"
+
+      - name: Get Current Python Version
+        shell: python
+        id: py
+        run: |
+          import os
+          import sys
+          import hashlib
+
+          v = hashlib.sha3_224(sys.version.encode()).hexdigest()
+
+          with open(os.environ["GITHUB_OUTPUT"], 'a+', encoding="utf-8") as f:
+              env = f.write(f"\nPY={v}\n")
+
+      - uses: actions/cache@v5
+        id: cache
+        with:
+          path: ~/.cache/pre-commit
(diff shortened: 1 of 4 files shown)


.github/workflows/linting.yaml
```diff
@@ -11,18 +11,44 @@ jobs:
...
+          post: pipx run pre-commit gc
 
   mypy:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.13'
+          python-version: "3.13"
```
dnicolodi, line 51: Unrelated change

.github/workflows/linting.yaml
```diff
@@ -11,18 +11,44 @@ jobs:
...
+          restore-keys: |
+            pre-commit-3-${{ steps.py.outputs.PY }}-
+
+      - run: pipx run pre-commit run --all-files --show-diff-on-failure --color=always
+
+      - name: Run this action with only post
+        if: steps.cache.outputs.cache-hit != 'true'
+        uses: gacts/run-and-post-run@v1
+        with:
+          post: pipx run pre-commit gc
```
dnicolodi, line 43: Can you explain exactly which problem are you solving here? Isn't it sufficient to keep the ruff version used here and in the pre-commit configuration aligned? The caching implemented here seems to be severely over engineered.

dnicolodi (changes requested):

dnicolodi:
This only adds complexity for no benefit at all. Just fix the pre-commit configuration, or do not submit PRs containing bogus changes implemented by a wrong pre-commit configuration.

trim21:
If we do not validate the files with pre-commit hooks, add pre-commit-config.yaml would only cause trouble for developer that use them. And for users that don't use ruff, they have no idea which ruff version should they pick.

The cache is just to speedup the job and reduce the usage of github cpu/cache which is some kind or public resource for developers, indeed this is not necessary, but still recommanded.

If maintaining these caching config is not a good idea for you I can replace it with simplly `pipx run pre-commit run -a ...`. would that be good for you?

PR #998, opened 2026-01-08
ci: re-enable pre-commit

trim21:
We now have 2 ruff entry in projects, one is in pre-commit hooks, with pinned version, and one is in ci (called from `make lint`), without any version specification.

So when someone running pre-commit, the files got formatted by a old version of ruff.

Also we have pre-commit-config.yaml  and did't valid it in ci, cause it `out-of-sync` in project files and pre-commit hooks. This will cause trouble for developer that use pre-commit. And for users that don't use ruff, they have no idea which ruff version should they pick to run `make lint`, the ruff version will changes over time.

It should be more proper that we also use pre-commit hooks to do the linting in CI. (and maybe also in `make lint`?)

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -13,8 +13,18 @@ jobs:
         with:
           python-version: '3.13'
       - uses: astral-sh/setup-uv@v5
-      - run: pip install ruff
       - run: make lint
+
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+      - uses: astral-sh/setup-uv@v7
+      - run: uv tool install pre-commit --
+      - run: pre-commit run --all-files --show-diff-on-failure --color=always
 
   mypy:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -154,9 +154,10 @@ constraints dep-constraints: build/beancount.deps
 
 
 # Run the linter on all source code.
+# keep the version in sync with ruff version we have in pre-commit-config.yaml
 ruff lint:
-	NO_COLOR=1 uv tool run ruff check .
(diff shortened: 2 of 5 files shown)


.github/workflows/linting.yaml
```diff
@@ -13,9 +13,19 @@ jobs:
...
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+      - uses: astral-sh/setup-uv@v7
+      - run: uv pip install pre-commit --system
+      - uses: trim21/actions/pre-commit@master
```
yagebu: Why not just run pre-commit via uv directly? I don't think having some personally-maintained action here is great.

trim21: It has a slow repo cloning and package installing progress I want to avoid by caching.

there is another option that we can use [prek](https://github.com/j178/prek) as replacement which has a github action https://github.com/j178/prek-action

yagebu: If we want the fastest, we could just enable https://pre-commit.ci/. Even with installing it from scratch, I'd bet this would still be one of the fastest CI steps, I don't think saving 5-10 seconds is worth this overhead.

Makefile
```diff
@@ -154,9 +154,10 @@ constraints dep-constraints: build/beancount.deps
 
 
 # Run the linter on all source code.
+# keep the version in sync with ruff version we have in pre-commit-config.yaml
 ruff lint:
-	NO_COLOR=1 uv tool run ruff check .
-	NO_COLOR=1 uv tool run ruff format .
+	NO_COLOR=1 uv tool run 'ruff==0.14.10' check .
```
yagebu, line 159: Wouldn't it be better to also use pre-commit here? (with e.g. `uv tool run pre-commit...`) Then we don't have to keep versions in sync

trim21: I actually don't use this command but only pre-commit, so no opinion here .....

PR #999, opened 2026-01-08
fix: printer should insert space between tags and links

trim21:
we previous print document as `2014-06-08 document Assets:Account1 "{path}" #tag1#tag2^link1^link2` which is ugly, now it's `2014-06-08 document Assets:Account1 "{path}" #tag1 #tag2 ^link1 ^link2`

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -13,6 +13,7 @@ import re
 import sys
 import textwrap
 from decimal import Decimal
+from typing import Iterator
 from typing import Optional
 
 from beancount.core import account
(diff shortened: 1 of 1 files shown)


beancount/parser/printer.py
```diff
@@ -352,12 +352,12 @@ def Note(self, entry, oss):
 
     def Document(self, entry, oss):
         oss.write('{e.date} document {e.account} "{e.filename}"'.format(e=entry))
-        if entry.tags or entry.links:
+        for tag in sorted(entry.tags):
```
yagebu: This now diverges how the tags and links are printed for notes and transactions. Why?

I'd also find it nicer with a space between but then that should be the same for all entry types, no?

trim21: Good point, I didn't use tag and links in notes and transactions so I didn't find it.

we should also add space in these types.

yagebu: maybe unify this in a `write_tags_and_links` method?

trim21: we already has space in them when printing transactions. so only need to change for note.

blais:
I would have made this a function not a method but that's alright.

PR #1006, opened 2026-01-18
build abi3 wheels for unix platform

trim21:
meson can't build abi3 wheels for windows yet.

the abi3 wheels is disabled by default so users still got non-abi3 wheel when they try to install from source, but we still build them in CI, so users of wheels can use them.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -6,6 +6,8 @@ on:
   workflow_dispatch:
 
   pull_request:
+    branches:
+      - master
     # trigger on self change to validation config
     paths:
       - .github/workflows/wheels.yaml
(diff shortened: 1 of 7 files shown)


dnicolodi:
> meson can't build abi3 wheels for windows yet.

This is not true.

> the abi3 wheels is disabled by default so users still got non-abi3 wheel when they try to install from source, but we still build them in CI, so users of wheels can use them.

What's the point of building abi3 wheels for beancount?

trim21:
> > meson can't build abi3 wheels for windows yet.
> 
> This is not true.

I haven't got it work on windows, feel free to fix it.


> > the abi3 wheels is disabled by default so users still got non-abi3 wheel when they try to install from source, but we still build them in CI, so users of wheels can use them.
> 
> What's the point of building abi3 wheels for beancount?


At least so far we haven't provide wheel for python3.14 on pypi yet, with abi3 wheels we can build less wheel and support more python versions.

image

blais:
I'll let you guys converge and agree on something, I'm agnostic, don't care much about packaging myself.

PR #1013, opened 2026-01-27
change required_daily_prices date range to include one day ranges

stenius:
fixes half of beancount/beanprice#134


beancount/ops/lifetimes.py
```diff
@@ -214,7 +214,7 @@ def required_daily_prices(lifetimes_map, date_last, weekdays_only=False):
             # Iterate over all weekdays.
             if date_end is None:
                 date_end = date_last
-            while date < date_end:
+            while date <= date_end:
```
yagebu, line 217: interval ends are exclusive, so this doesn't look correct. Rather, this seems like a bug in the calling code.

PR #1015, opened 2026-01-28
build: use pypi distributed flex/bison

trim21:
I just create pre-compiled binary on pypi for flex/bison/winflexbison , so it will be possible to use then to build beancount without extra system tool dependency.

They don't target to any python abi, so they doesn't need to release a new version for new python.

https://pypi.org/project/flex-bin/#files
https://pypi.org/project/bison-bin/#files
https://pypi.org/project/winflexbison-bin/#files

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -56,21 +56,6 @@ jobs:
         env:
           CIBW_SKIP: '*-musllinux*'
           CIBW_ARCHS: ${{ matrix.architecture }}
-          CIBW_BEFORE_ALL_LINUX: |
-            curl -sSfO https://ftp.gnu.org/gnu/bison/bison-3.8.2.tar.xz
-            tar xJf bison-3.8.2.tar.xz
-            cd bison-3.8.2 && ./configure && make && make install
-            curl -sSfLO https://github.com/westes/flex/releases/download/v2.6.4/flex-2.6.4.tar.gz
-            tar xzf flex-2.6.4.tar.gz
-            cd flex-2.6.4 && ./configure CFLAGS="-D_GNU_SOURCE" && make && make install
-          CIBW_ENVIRONMENT_MACOS: |
-            PATH="/usr/local/opt/bison/bin:/opt/homebrew/opt/bison/bin:$PATH"
-          CIBW_BEFORE_ALL_MACOS: |
-            brew install bison
-          CIBW_ENVIRONMENT_WINDOWS: |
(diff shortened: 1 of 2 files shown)


dnicolodi:
Which problem does this solve?

To compile beancount you still need to install a C compiler (and on most Linux distributions, the Python development package). The only platform where installing these tools is not a single command away is Windows and there installing a C compiler is already complicated and something that cannot be expected of users. On macOS installing the development tools is maybe simpler, but still not trivial. However, it is very easy to build wheels for all the architectures on which Windows or macOS can be used, and we do it. The only platform for which we do not provide binary wheels for all architectures is Linux, there are too many supported. However, installing development tools on Linux is trivial and I expect it to not be a problem for anyone running Linux a non-mainstream architecture.

I have a lot of reserves on having beancount depend on packages maintaned by a single developer. Especially if these packages are used only in the build process to generate code. It seems the perfect recipe for a supply chain attack.

I have the impression that this is somehow related to the fact that beancount still does not have a release with binary wheels for Python 3.14. However, this PR does not solve this issue either: we still do not have a release that makes installing on Python 3.14 easy, for two reasons: this PR is not in a release yet, and users still have to install other development tools (which on Windows is unlikely to happen, as expressed above).

I was much more sympathetic to the idea of making beancount use the Python stable ABI, if done properly, and if it does not result in a performance regression (using the stable ABI transforms some macro accessors and inline functions into proper function calls, which may result in a performance impact).

I also think that using PyPI to distribute binaries of tools that are not Python specific is a misuse of community resources. If you want an ecosystem that packages Python packages and native packages in a seamless way, you may want to look into conda-forge. pixi https://pixi.prefix.dev/latest/ makes working with conda-forge packages very nice (I use it on Windows at $WORK).

yagebu:
I'm with @dnicolodi on this one and do not like installing opaque blobs from PyPI - in build time at particular.

If flex/bison is such a problematic dependency, I think either again committing the flex/bison output or at least including it in the distributed source distribution would be better workarounds than this supply chain attack waiting to happen.

I'm very much in favour in having abi3 wheels.

dnicolodi:
> I think either again committing the flex/bison output or at least including it in the distributed source distribution would be better workarounds

I don't like having generated files in version control. But we can definitely have the generated files included in the sdist. I'll work on a patch as soon as I'll find some time (don't hold your breath).

trim21:
great idea, I'll send a PR. But I think commit generated files in git is a good idea.

Issue #388, opened 2019-04-12
Are unknown lines ignored or do they produce erros


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/388) by Martin Michlmayr (Bitbucket: [tbm13](https://bitbucket.org/tbm13), GitHub: [tbm](https://github.com/tbm)).**

----------------------------------------

Jonathan Corbet reviewed [beancount on LWN](https://lwn.net/SubscriberLink/785553/39167e23980e7e4c/):

"It is also worth noting that beancount explicitly ignores any lines in the file that it doesn't recognize. The apparent purpose of this feature is to allow other types of markup in the file; the ability to manage it with the Emacs Org mode is said to be especially gratifying. But this behavior can also make it easy for mistakes to pass unnoticed. Similar behavior in cron has led to security vulnerabilities in the past. Your editor would feel much better knowing that a corrupted beancount file would not escape notice."

I wondered if that's true and Jonathan [referred to the manual](https://docs.google.com/document/d/1wAMVrKIA2qtRGmoVDSUBJGmYZSygUaR0uOMW1GV3YE0/edit#heading=h.9zjhwskw53j8) which says "Any line that does not begin as a valid Beancount syntax directive (e.g. with a date) is silently ignored".

It seems like the manual is wrong because:

    echo foo > x
    bean-check x
    ..../x:1:       Invalid token: 'foo'

Can you clarify, and if necessary, update the manual?

FWIW, beancount's syntax is stricter in many ways than ledger (where some unknown lines are ignored) and I agree with Jonathan that anything unknown should result in an error.

blais:
**Original comment by Martin Michlmayr (Bitbucket: [tbm13](https://bitbucket.org/tbm13), GitHub: [tbm](https://github.com/tbm)).**

----------------------------------------

Martin commented on the list:

"I think that comment stems from outdated documentation.
Org-mode / Markdown start-line characters are now ignored explicitly, see
here:
https://bitbucket.org/blais/beancount/src/fa1edde3bcd02a277fac193f460a39c9a1461161/beancount/parser/lexer.l#lines-173"

So it seems the manual needs updating.

blais:
**Original comment by Luyu Cheng.**

----------------------------------------

Thanks. I’d been crazy on why “invalid token” occurs everywhere. This issue helped me found a nondescript mistakenly typed quote.

tswast:
To be honest, I kinda got my hopes up when reading this same line in the docs,

[...]

PR #1021, opened 2026-03-07
feat: introduce I/O plugin architecture for parser

tswast:
Adds support for I/O plugins via the '.beanrc' config file. I/O plugins run just before the C parser is invoked on a file, allowing dynamic preprocessing such as decryption. To demonstrate this architecture, the existing file encryption logic has been extracted from loader.py and reimplemented as a built-in I/O plugin module:
beancount.plugins.io.encryption.

For some info on my motivation for this change, see: https://docs.google.com/document/d/16conmm3cKC4raXnFW1UiiSmMGIhs0qO2oLK2siHUBOI/edit?usp=sharing and https://github.com/beancount/beancount/issues/388

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,8 @@ CLUSTERS_REGEXPS =							\
 	beancount/parser		 	parser			\
 	beancount/plugins/.*_test\.py	 	plugins/tests		\
 	beancount/plugins		 	plugins			\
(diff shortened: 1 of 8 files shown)


tswast:
Per discussion on the mailing list, I'm going to pursue a token/ast-based approach in a fork.

Issue #398, opened 2019-06-11
Support logograms (like CJK characters) as account names.


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/398) by Alex Hou (Bitbucket: [alex-hou](https://bitbucket.org/alex-hou), GitHub: [alex-hou](https://github.com/alex-hou)).**

----------------------------------------

According to beancount language syntax, account names should follow this rule:

> Each component of the account names begin with a capital letter or a number and are followed letters, numbers or dash \(-\) characters. All other characters are disallowed.

But logograms \(Chinese characters, Japanese hiragana and katakana, Korean alphabet, etc.\) have no ‘capital letter’.

Please support these characters as account names.

[http://www.programminginkorean.com/programming/hangul-in-unicode/](http://www.programminginkorean.com/programming/hangul-in-unicode/)

> Hangul Syllables \(AC00–D7A3\)  
> Hangul Jamo \(1100–11FF\)  
> Hangul Compatibility Jamo \(3130–318F\)  
> Hangul Jamo Extended-A \(A960–A97F\)  
> Hangul Jamo Extended-B \(D7B0–D7FF\)

[https://stackoverflow.com/a/30200250/4458143](https://stackoverflow.com/a/30200250/4458143)  
[http://www.rikai.com/library/kanjitables/kanji\_codes.unicode.shtml](http://www.rikai.com/library/kanjitables/kanji_codes.unicode.shtml)

> Hiragana \(3040 - 309F\)  
> Katakana \(30A0 - 30FF\)

[https://en.wikipedia.org/wiki/Halfwidth\_and\_Fullwidth\_Forms\_\(Unicode\_block\)](https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block))

> halfwidth forms of compatibility jamo characters for Hangul \(FFA0–FFDC\)  
> half-width katakana \( FF65–FF9F \)

‌

[https://stackoverflow.com/a/1366113/4458143](https://stackoverflow.com/a/1366113/4458143)  
[https://en.wikipedia.org/wiki/CJK\_Unified\_Ideographs](https://en.wikipedia.org/wiki/CJK_Unified_Ideographs)

```
Block                                      Range           Comment
CJK Unified Ideographs                     4E00-9FFF       Common
CJK Unified Ideographs Extension A         3400-4DBF       Rare
CJK Unified Ideographs Extension B         20000-2A6DF     Rare, historic
[...]

Issue #161, opened 2017-03-12
Unicode support


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/161) by rutsky (Bitbucket: [rutsky](https://bitbucket.org/rutsky), GitHub: [rutsky](https://github.com/rutsky)).**

----------------------------------------

I believe this issue arised [several](https://groups.google.com/d/msg/ledger-cli/C6GOOj8kGtQ/6pGThn3DRTIJ) [times](https://groups.google.com/d/msg/beancount/vVzzfE3m_ow/usM6Ft6-zi0J), but haven't found ticket for it with information about current status and plans.

For example, I want to have account name with Unicode characters in it. As I see beancount currently doesn't support that:

```
$ cat test2.beancount 
option "operating_currency" "RUB"

2014-01-31 open Assets:Imbalance-RUB                            RUB
2014-01-31 open Assets:Orphan-RUB                               RUB
2014-01-31 open Assets:Активы                                   RUB
$ bean-check test2.beancount 
/home/bob/stuff/money/test2.beancount:5:       Invalid token: 'Assets:Активы'
$ pip list | grep beancount
beancount (2.0b15)
```

I'm trying to switch from Gnucash to Beancount (with [gnucash-to-beancount](https://github.com/henriquebastos/gnucash-to-beancount) converter tool), and limiting account names to ASCII is a showstopper for me.

What is the current status of Unicode support (in account names and other entities)?

What are the plans to support Unicode everywhere?

Maybe there are known workarounds?

blais:
**Original comment by vrutsky (Bitbucket: [vrutsky](https://bitbucket.org/vrutsky), GitHub: [vrutsky](https://github.com/vrutsky)).**

----------------------------------------

Related: https://bitbucket.org/blais/beancount/pull-requests/14/allow-unicode-in-account-names

/CC: @medranocalvo_

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

Not for 1.0. This will come after.
This requires a fair amount of work in order to be done well and without constant bug reports.

(I'll suggest moving forward with romanized names, you can always rename everything later easily with a sed command or text editor.)

Issue #31, opened 2015-03-09
NameError: name 'prices' is not defined


blais:
**[Original report](https://bitbucket.org/blais/beancount/issue/31) by clemente (Bitbucket: [clemente](https://bitbucket.org/clemente), GitHub: [clemente](https://github.com/clemente)).**

----------------------------------------

~~~~
beancount> select price,convert(price,'EUR')
Traceback (most recent call last):
  File "/usr/lib/python3.4/cmd.py", line 214, in onecmd
    func = getattr(self, 'do_' + cmd)
AttributeError: 'BQLShell' object has no attribute 'do_select'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/w/beancount/src/python/beancount/query/shell.py", line 225, in default
    self.dispatch(statement)
  File "/w/beancount/src/python/beancount/query/shell.py", line 215, in dispatch
    return method(statement)
  File "/w/beancount/src/python/beancount/query/shell.py", line 358, in on_Select
    self.options_map)
  File "/w/beancount/src/python/beancount/query/query_execute.py", line 220, in execute_query
    values = [c_expr(context) for c_expr in c_target_exprs]
  File "/w/beancount/src/python/beancount/query/query_execute.py", line 220, in <listcomp>
    values = [c_expr(context) for c_expr in c_target_exprs]
  File "/w/beancount/src/python/beancount/query/query_env.py", line 229, in __call__
    return prices.convert_amount(context.price_map, args[1], args[0])
NameError: name 'prices' is not defined
~~~~

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

There's an easy "fix" to this, but it won't help, the feature isn't quite ready.
It turns out correctly computing market prices requires a bit more thought.

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

To resolve this issue, complete working out the semantics of the CONVERT() function in SQL. In particular, it should do something right if no price is available in the price database.

blais:
**Original comment by Martin Blais (Bitbucket: [blais](https://bitbucket.org/blais), GitHub: [blais](https://github.com/blais)).**

----------------------------------------

I think this issue has been addressed at this point, with CONVERT() and GETPRICE() and such.

PR #1022, opened 2026-03-08
Allow a wider range of Unicode characters in account names

### LINKED ISSUE #377: account_types.is_root_account() needs update for UTF-8
**[Original report](https://bitbucket.org/blais/beancount/issue/377) by Martin Michlmayr (Bitbucket: [tbm13](https://bitbucket.org/tbm13), GitHub: [tbm](https://github.com/tbm)).**

beancount nowadays allows UTF-8 in the account type.

beancount/core/account.py defines:

    ACC_COMP_TYPE_RE = regexp_utils.re_replace_unicode(r"[\p{Lu}][\p{L}\p{Nd}\-]*")

However, account_types.is_root_account() has not been updated accordingly:

                    bool(re.match(r'([A-Z][A-Za-z0-9\-]+)$', account_name)))

Test:

    from beancount.core import account_types

    print(account_types.is_root_account("Töst"))

will print `False` but it should be `True`

### LINKED ISSUE #733: account names support utf-8 characters?
for now, it seems that the name not starting with `a-zA-Z0-9` failed to be recognized.

`1账户` or `A账户` will be recognized, but `账户` will not be recognized.

I've seen someone post a similar issue before, but utf-8 name is still not fully supported.

any plan on this kind of issue?

Zjl37:
This PR fixes #733, #398, #161, and by the way fixes #377.

I began this patch initially to fix the problem that “Chinese characters cannot be used at the start of account names, because only upper case letters are allowed but not ideographs or letters with no casing here” (#733), but then put efforts to devise a more Unicode-aware syntax for account name, taking inspirations form [UAX #31](https://www.unicode.org/reports/tr31), to allow for a codepoint range as wide as reasonably appropriate while respecting some beancount traditions, in the hope that users from different countries will not be surprised or disappointed again with their home language being banned.

Specifically, this PR proposes a new syntax for account name defined by the following regular expressions. Don't be scared -- I'll explain them in the next comment. The choice of characters/character classes is good enough up to my i18n knowledge, but of course open to discussion.

https://github.com/beancount/beancount/blob/8b74e4b994ad1736adbd70945d429bef72733dac/beancount/core/account.py#L37-L50

worthy to note that, as an effect, in the ASCII range the new syntax also allows underscore `_` after the first character in account name. The lexer has been updated to reflect this. I hope this is acceptable and would not cause significant breaks.

This PR also supersedes pr #937 and #706.

**Why not #706**:

#706 and similar historic PRs hardcodes allowed code point ranges and lists in this repo, which puts maintainance burden. The list could easily be or become incomplete in terms of either what users expect or what any standards define. it is not future-proof as new versions of Unicode may add new characters to a language.

**Why not #937**:

#937 defines the syntax using `\p{Han}`, that is, by scripts/writing systems. This is still not optimal because it only considers the Han script (Chinese ideographs). In this case we'll need to add support for each language without casing upon request, or enumerate all unicameral languages. It is not future-proof as well because new versions of Unicode might add new languages/writing systems, which could have no casing, theoretically.

This PR, on the other hand, defines the syntax using [General Category](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category) and some other Unicode character property. As per Wikipedia, "Each code point is assigned a value for General Category", so newly assigned code points will get a General Category, meaning we have a tighter grasp on what characters we want. The ranges of these character classes are implemented by the new `regex` engine, meaning they will be updated as our `regex` dependency updates to support a newer Unicode version.


Zjl37:
Let me first express myself on general approach: Why follow UAX#31, why modify it, how to modify it

Account names in beancount are very much like identifiers in programming languages. They are used to uniquely identify something. They are not some quoted literals but sit directly in the global scope of source files. Thus the [Unicode® Standard Annex #31: Unicode Identifiers and Syntax](https://www.unicode.org/reports/tr31) are of great reference significance for us. By following closely to UAX#31, we can better align with what characters other (programming) languages allows, and what international users expects to use in such context. It also helps avoiding potential i18n pitfalls that the Unicode standard have already considered thoroughly, as long as we take some effort to understand its specs and jargons. On the other hand, beancount puts some restrictions on the style of account names, which means we'll be changing the specifications in UAX#31. The python language also does modify UAX#31 to support non-ASCII identifiers in their PEP 3131.

As discussed earlier in the two "why not" questions, *General Category* covers all codepoints and languages, updates with Unicode version and is future-proof. It's a good unit for classification. So I prefer doing customization in this granularity.

---

Let's now take a look at the syntax itself.

For clarity let's break down `ACC_COMP_TYPE_RE` and `ACC_COMP_NAME_RE`, as each component will get longer later:

https://github.com/beancount/beancount/blob/8b74e4b994ad1736adbd70945d429bef72733dac/beancount/core/account.py#L46-L47

The equivalent values for these component of the original patterns are:

```py
ACC_NAME_Start = r"[\p{Lu}]"
ACC_TYPE_Start = r"[\p{Lu}\p{Nd}]"
ACC_NAME_Continue = r"[\p{L}\p{Nd}\-]*"
```

Let's compare them to UAX#31 specs. The main sections we're looking at is [2 Default Identifiers](https://www.unicode.org/reports/tr31/#Default_Identifier_Syntax), especially [Table 2. Properties for Lexical Classes for Identifiers](https://www.unicode.org/reports/tr31/#Table_Lexical_Classes_for_Identifiers). We'll not care `XID_Start` and `XID_Continue`, as we do not apply any Unicode Normalization for now. What we're doing now is to take the "UnicodeSet notation" for `ID_Start` and `ID_Continue`, transform them to regex notation, and then evaluate, take or modify each character classes.

#### Letters

Beancount requires account names to start with capital letters. The old pattern implements this as `\p{Lu}`([Uppercase Letter](https://www.compart.com/en/unicode/category/Lu) [^1]), which fails to take care of ideographs or letters with no cases. In the eyes of users speaking such a language and really wishing to book keep in their language, this rule suddenly becomes "I can't use my home letters at the start of account name", which is surprising and disappointing and might made them give up beancount. Mixing scripts like `A资产`, if the user insists, might spam their ledger with visual ugliness and increase typing difficulty. This is all the mentioned issues are about.

[Section 3.1.2 Semantics Based on Case of UTS#55](https://www.unicode.org/reports/tr55/#Case-Semantics) suggests that:

> In general, when placing requirements on case, implementations should disallow the unwanted case (e.g., disallow lowercase), rather than requiring the desired case (e.g., requiring uppercase).

So wherever the old pattern requires `\p{Lu}`, we'll use `[\p{L}--\p{Ll}]` (Letter excluding [Lowercase Letter](https://www.compart.com/en/unicode/category/Ll))

In fact, this change alone, I guess, is enough to satisfy almost 100% the need of the authors of the mentioned issues.

#### Numbers

Wherever the old pattern allows `\p{Nd}`([Decimal Number](https://www.compart.com/en/unicode/category/Nd)), I prefer using `\p{N}` (Number) in the new pattern.

Note that the UAX spec allows `\p{Nl}`([Letter Number](https://www.compart.com/en/unicode/category/Nl)) in `ID_Start`, even when ascii decimal digits (0-9) are not allowed at the start. (because they looks like letters?) So we'll keep this in our `ACC_TYPE_Start`.

#### Other categories and properties

For other character classes that beancount seems not specially care about, I prefer following what the UAX#31 spec allows and excludes.

With the above principles, we can summarize the new values for `ACC_TYPE_Start` and `ACC_NAME_Start` as follows:

```py
ACC_NAME_Start = r"[\p{L}\p{N}\p{Other_ID_Start}--\p{Pattern_Syntax}\p{Pattern_White_Space}\p{Ll}]"
ACC_TYPE_Start = r"[\p{L}\p{Nl}\p{Other_ID_Start}--\p{Pattern_Syntax}\p{Pattern_White_Space}\p{Ll}]"
```

Let's move on to look at `ID_Continue`. It first includes a `\p{ID_Start}`. (meaning every character allowed at the start is allow after) As this is a derived property and we have modified our corresponding syntax, we'll copy `ACC_NAME_Start` and build our `ACC_NAME_Continue` from there.

#### dashes

Wherever the old pattern allows ascii dash `[\-]`, I prefer extending to `\p{Pd}`([Dash Punctuation](https://www.compart.com/en/unicode/category/Pd)) in the new pattern.

However, as ascii dash is excluded by `\p{Pattern_Syntax}` (what Unicode thinks by default is reserved for "syntactic use"), we'll need to add them back at the very last. This results in the nested sets in the new pattern.

With the above principles, we can put the new `ACC_NAME_Continue` as:

```py
ACC_NAME_Continue = r"[[\p{L}\p{N}\p{Other_ID_Start}\p{Mn}\p{Mc}\p{Pc}\p{Other_ID_Continue}--\p{Pattern_Syntax}\p{Pattern_White_Space}]\p{Pd}]"
```

It's worthy to note that, the inclusion of `\p{Pc}` ([Connector Punctuation](https://www.compart.com/en/unicode/category/Pc)) allows ascii underscore _ in `ACC_NAME_Continue`. To actually make this work the lexer (`beancount/parser/lexer.l`) has also been updated to keep syntax rules in sync. I prefer to include this category because underscore has the "not breaking" semantic in many contexts, and there has been historical requests to use underscores in currency names.

The new patterns uses [Nested sets and set operations](https://github.com/mrabarnett/mrab-regex?tab=readme-ov-file#nested-sets-and-set-operations), which is Version 1 behavior of the `regex` engine. Therefore whenever these patterns are used or indirectly used it must be accompanied by `regex.V1` flag.

[^1]: These links are put here for you to quickly get a glimpse of what characters these categories contains. It's a site among top search results that seems the most accurate and complete but might not actually be so. After all we're sticking to [Unicode Character Database](https://www.unicode.org/reports/tr44) or the implementation of the `regex` engine for ground truth.

Zjl37:
Finally, let's talk about compatibility with the C-based lexer.

As we can see from the lexer rule (https://github.com/beancount/beancount/blob/eb8be17c750257dc693e3173c30d2cabc9ab2faa/beancount/parser/lexer.l#L121-L130), the lexer allows any valid multi-byte UTF-8 sequence (`UTF-8-ONLY`) in account name. So all non-ascii code points we have extended in the new syntax are within the range the lexer can recognize.

For the ascii range, we can quickly validate with a python script that:

- the allowed ASCII character by `ACC_NAME_Start` and `ACC_TYPE_Start` of the new account name syntax is exactly the same as the old lexer rules recogonizes.
- the allowed ASCII character by `ACC_NAME_Continue` of the new syntax is exactly one character more than the old lexer rules can recognize, namely underscore `_`. This PR has added it to the lexer rule.

<details>
  <summary>The python script to reveal the allowed ASCII character set for regex patterns</summary>

```py
# mrab_regex_test.py
import regex

ACC_NAME_Start = (
    r"[\p{L}\p{N}\p{Other_ID_Start}--\p{Pattern_Syntax}\p{Pattern_White_Space}\p{Ll}]"
)
ACC_TYPE_Start = (
    r"[\p{L}\p{Nl}\p{Other_ID_Start}--\p{Pattern_Syntax}\p{Pattern_White_Space}\p{Ll}]"
)
ACC_NAME_Continue = r"[[\p{L}\p{N}\p{Other_ID_Start}\p{Mn}\p{Mc}\p{Pc}\p{Other_ID_Continue}--\p{Pattern_Syntax}\p{Pattern_White_Space}]\p{Pd}]"

Pd = r"\p{Pd}"
Pc = r"\p{Pc}"
Pattern_Syntax = r"\p{Pattern_Syntax}"
Pattern_White_Space = r"\p{Pattern_White_Space}"

for pattern in [ACC_NAME_Start, ACC_TYPE_Start, ACC_NAME_Continue, Pd, Pc, Pattern_Syntax, Pattern_White_Space]:
    print("pattern:", pattern)
    print(
        "matches ASCII character:",
        "👉"
        + "".join(chr(i) for i in range(128) if regex.match(pattern, chr(i), regex.V1))
        + "👈",
    )
    print()
```

<details>
  <summary>Expected output</summary>

```
pattern: [\p{L}\p{N}\p{Other_ID_Start}--\p{Pattern_Syntax}\p{Pattern_White_Space}\p{Ll}]
matches ASCII character: 👉0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ👈

pattern: [\p{L}\p{Nl}\p{Other_ID_Start}--\p{Pattern_Syntax}\p{Pattern_White_Space}\p{Ll}]
matches ASCII character: 👉ABCDEFGHIJKLMNOPQRSTUVWXYZ👈

pattern: [[\p{L}\p{N}\p{Other_ID_Start}\p{Mn}\p{Mc}\p{Pc}\p{Other_ID_Continue}--\p{Pattern_Syntax}\p{Pattern_White_Space}]\p{Pd}]
matches ASCII character: 👉-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz👈

pattern: \p{Pd}
matches ASCII character: 👉-👈

pattern: \p{Pc}
matches ASCII character: 👉_👈

pattern: \p{Pattern_Syntax}
matches ASCII character: 👉!"#$%&'()*+,-./:;<=>?@[\]^`{|}~👈

pattern: \p{Pattern_White_Space}
matches ASCII character: 👉	
 👈

```

</details>

</details>

Overall the new syntax is lenient in non-ASCII range and conservative in ASCII range with minimal updates in terms of what new characters it introduces. I hope with these observations there will not be big breaks in the lexer and grammar system. Nevertheless please test this patch against you existing ledgers to find problems early!

PR #1023, opened 2026-03-16
Add support for printing @@ postings on request

tmartin-gh:
I have a use case where I parse some external data and use the beancount APIs to create Postings and Transaction objects then print them.  For stocks, it is easier for me to deal with the total amount of a posting then to handle multi-digit per-unit costs.  While the parser supports the @@ format for specifying the total, the printer currently only supports printing per-unit prices.

This PR provides a metadata-based method to request the printer to print using the @@ format.

An example for using this is:

```
account = ...
units = ...
cost = ...
price = ...
flag = ...
meta = {}
meta['__print_price_total_format'] = '{:.2f}'
posting = data.Posting(account=account, units=units, cost=cost, price=price, flag=flag, meta=meta)
```

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -296,7 +296,16 @@ class EntryPrinter:
             position_str = ""
 
         if posting.price is not None:
-            position_str += " @ {}".format(posting.price.to_string(self.dformat_max))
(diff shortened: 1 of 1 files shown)


blais:
No tests.

PR #1024, opened 2026-03-17
Add support for printing @@ postings on request

tmartin-gh:
I have a use case where I parse some external data and use the beancount APIs to create Postings and Transaction objects then print them. For stocks, it is easier for me to deal with the total amount of a posting then to handle multi-digit per-unit costs. While the parser supports the @@ format for specifying the total, the printer currently only supports printing per-unit prices.

This PR provides a metadata-based method to request the printer to print using the @@ format, along with a pytest unittest showing example usage.

(Related to #1023 but adds unit test, I don't appear to have permission to re-open closed PRs)

### DIFF
--- a/beancount/parser/printer.py
+++ b/beancount/parser/printer.py
@@ -296,7 +296,16 @@ class EntryPrinter:
             position_str = ""
 
         if posting.price is not None:
-            position_str += " @ {}".format(posting.price.to_string(self.dformat_max))
+            meta = posting.meta or {}
+            price_total_format = meta.get("__print_price_total_format")
+            if price_total_format is not None:
+                # Calculate the total amount by multiplying the per-unit price
+                # and print using the @@ syntax.
+                total_number = posting.price.number * abs(posting.units.number)
+                total_amount = amount.Amount(total_number, posting.price.currency)
+                position_str += " @@ {}".format(total_amount.to_string(price_total_format))
+            else:
+                position_str += " @ {}".format(posting.price.to_string(self.dformat_max))
 
         return flag_account, position_str, weight_str
 
--- a/beancount/parser/printer_test.py
+++ b/beancount/parser/printer_test.py
@@ -13,9 +13,11 @@ from decimal import Decimal
 from beancount import loader
 from beancount.core import data
 from beancount.core.amount import Amount
+from beancount.core.number import MISSING
 from beancount.ops.balance import BalanceError
 from beancount.parser import cmptest
 from beancount.parser import printer
+from beancount.core.position import CostSpec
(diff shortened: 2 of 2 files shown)


blais:
Hmm interesting take. I don't like the metadata though; why do you prefer this way?
Would it work for you to always print the total amount and have that state in the printer object instead?

tmartin-gh:
> Hmm interesting take. I don't like the metadata though; why do you prefer this way? Would it work for you to always print the total amount and have that state in the printer object instead?

When I first wrote this, I imagined the decision would be per-posting, but as I've been using it I find I pretty much exclusively use the @@ syntax.  I think a printer state variable would be fine.  Are you asking me to update the PR?

mmoya:
OP says _While the parser supports the @@ format for specifying the total, the printer currently only supports printing per-unit prices._ but more precisely: it's not a matter of printer support, it's that total_price is discarded after parsing.

I've needing to print `@@` and took the lazy way of adding an optional `Posting.price_total` and get the printer emit `@@` if `price_total` is set. Not the most elegant solution because `price_total` feels redundant with `price` in Posting but I didn't see any other obvious way to maintain the exact input value instead of deriving it from units*price.

I think the @@-information-lost limitation is indirectly acknowledged in Vnext's [Intermediate Parsed Data vs. Final List of Directives](https://beancount.github.io/docs/beancount_v3/#intermediate-parsed-data-vs-final-list-of-directives).

tmartin-gh:
I haven’t posted back on this as I’ve been working through all of my data / scenarios.  I can’t say this with 100% certainty but so far it is looking like the vast majority of my scenarios do not need the @@ price_total at all, instead what I really need is a cost_total and then I can use the format {0 # total_cost, date, label} or the syntactic sugar {{total_cost, date, label}}. 

From what I observe of existing Beancount unit tests, my use of # is strictly not per-design.  The design intent appears to be {per_unit_cost # total_comission, date, label}.  Then internally the real unit cost is re-calculated by adding the total comission divided by number of units. 

I have solved the vast majority of my scenarios by adding a total_cost field to the Cost object and modifying some of the core beancount code.  This is similar in spirit to your concept of a price_total.    I realize this will get a lot of scrutiny so I’ve been working to add additional unit tests to demonstrate all of my scenarios that aren’t handled by the existing code.  I’ll make a separate PR for that once I’m ready.

tmartin-gh:
Closing this as I no longer need it.  For my use case, I was incorrectly using the @@ total price syntax and what I really wanted was a total _cost_ syntax.  The recently merged PR https://github.com/beancount/beancount/pull/1033 solves my use case.

PR #1034, opened 2026-06-01
Fix insecure pickle cache placement (CWE-502)

alanturing881:
### Summary

The default pickle cache is stored alongside the `.beancount` file as
`.{filename}.picklecache`. Since `pickle.load()` is called without any integrity
verification, an attacker who can write to the same directory as a user's beancount
file can plant a malicious cache file that executes arbitrary Python code the next
time any beancount command (`bean-check`, `bean-report`, `bean-query`, etc.) processes
that file. The cache is enabled by default.

**Attack vector**: supply a `.beancount` file alongside a crafted `.picklecache` — e.g.,
in a shared directory, a downloaded demo archive, or a collaborative repository.

### Root Cause

```python
# loader.py:231 — no signature or HMAC check before deserialization
result = pickle.load(file)   # executes __reduce__ in attacker-controlled payload
```

The `needs_refresh()` freshness check happens *after* `pickle.load()` returns, so the
deserialization (and any embedded payload) already ran regardless of whether the cache is
considered fresh.

**Vulnerable code:**
```python
PICKLE_CACHE_FILENAME = ".{filename}.picklecache"   # same dir as data file

if exists:
    with open(cache_filename, "rb") as file:
        result = pickle.load(file)   # ← RCE if file is attacker-controlled
```

### Fix

Move the default cache location to the user's OS-specific cache directory
(`XDG_CACHE_HOME/beancount` on Linux, `~/Library/Caches/beancount` on macOS,
`LOCALAPPDATA/beancount` on Windows). A 16-char SHA-256 prefix of the source
directory is incorporated in the filename to prevent collisions between files with
the same basename in different directories.

```python
# After fix — cache stored where only the user can write
# /home/alice/.cache/beancount/.finances.beancount-a1b2c3d4e5f6a7b8.picklecache
```

The `BEANCOUNT_LOAD_CACHE_FILENAME` env-var override and the explicit
`cache_filename` argument to `initialize()` still accept the legacy relative
pattern for backwards compatibility.

### Impact
- **Type**: CWE-502 — Deserialization of Untrusted Data → RCE
- **Auth required**: No (file system access to data directory)
- **Affected versions**: All versions with pickle cache (default)
- **CVSS:3.1**: AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H = **7.8 High**

### Verification

Dynamically confirmed on beancount v3.2.3 (python:3.12-slim Docker) — `touch /tmp/BEANCOUNT_PWNED_3_2_3` executed silently via `bean-check test.beancount` before fix; no longer executes after fix because the malicious `.test.beancount.picklecache` in the data directory is not loaded.

### References
- CWE-502: https://cwe.mitre.org/data/definitions/502.html
- Similar: CVE-2022-21699 (Jupyter — arbitrary code via pickle)

### DIFF
--- a/beancount/loader.py
+++ b/beancount/loader.py
@@ -72,6 +72,35 @@ RENAMED_MODULES: dict[str, str] = {}
 
 
 # Filename pattern for the pickle-cache.
+# Stored in the user's cache directory (not alongside the data file) to prevent
+# an attacker who has write access to the data directory from injecting a
+# malicious pickle payload. See CWE-502.
+def _get_default_cache_dir() -> str:
+    """Return a user-specific cache directory that the data-file owner controls."""
+    if sys.platform == "win32":
+        base = os.environ.get("LOCALAPPDATA") or os.path.expanduser("~")
+    elif sys.platform == "darwin":
+        base = os.path.expanduser("~/Library/Caches")
+    else:
+        base = os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache")
+    return os.path.join(base, "beancount")
+
+
+def _make_cache_filename(filename: str) -> str:
+    """Return a collision-free cache path for *filename* under the user cache dir.
+
+    Incorporates a hash of the file's absolute directory so that two files with
+    the same basename in different directories map to distinct cache files.
+    """
+    abs_filename = path.abspath(filename)
+    dir_hash = hashlib.sha256(path.dirname(abs_filename).encode()).hexdigest()[:16]
+    cache_name = f".{path.basename(abs_filename)}-{dir_hash}.picklecache"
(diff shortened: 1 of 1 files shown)


yagebu:
> Attack vector: supply a .beancount file alongside a crafted .picklecache — e.g.,
in a shared directory, a downloaded demo archive, or a collaborative repository.

Operating on untrusted data (or an directory with untrusted data) is not really an attack vector that Beancount is designed against. With normal Beancount features, this "attack" is already possible: you can provide a custom plugin, put that next to the beancount file and run arbitrary Python code in the plugin.

alanturing881:
Thank you for the information.

PR #1049, opened 2026-08-29
booking: return an error for total cost on zero units

### LINKED ISSUE #1048: Zero units with a total cost raises an uncaught decimal.DivisionByZero out of load_file
A posting with zero units and a total cost divides by the unit count without guarding it, so `load_file` propagates `decimal.DivisionByZero` instead of returning an error. `bean-check` prints a traceback.

## Reproduction

```beancount
2024-01-01 open Assets:S HOOL
2024-01-01 open Assets:Cash USD

2024-02-01 * "zero units, total cost"
  Assets:S      0 HOOL {{20.00 USD}}
  Assets:Cash  -20.00 USD
```

```
$ bean-check min.beancount
  ...
  File ".../beancount/parser/booking_full.py", line 769, in convert_costspec_to_cost
    unit_cost = cost_total / units_number
                ~~~~~~~~~~~^~~~~~~~~~~~~~
decimal.DivisionByZero: [<class 'decimal.DivisionByZero'>]
$ echo $?
1
```

The `{0 # 20.00 USD}` spelling of the same thing fails identically, at the same line. Reproduced on beancount 3.2.3 with Python 3.13, and the division is still unguarded on master.

## Cause

`convert_costspec_to_cost` divides the total cost by the unit count with no zero check:

```python
# booking_full.py:761
if number_total is not None:
    units_number = abs(posting.units.number)
    cost_total = number_total
    if number_per is not MISSING:
        cost_total += number_per * units_number
    unit_cost = cost_total / units_number     # line 769
```

`compute_cost_number` (same module, line 718) contains the same unguarded division, but it is not on this input's path. Instrumenting it shows it is never called for this file, so treat it as a separate thing worth checking rather than a second confirmed exposure here.

## Why this looks like a bug rather than a rejected input

The input is not meaningful and should be rejected. The issue is how. `load_file` is documented to return `(entries, errors, options_map)`, and callers rely on parse and booking problems arriving as entries in `errors`. An exception escaping the loader takes down anything built on it rather than surfacing a diagnostic.

Zero units are already treated as needing a guard in this same arithmetic elsewhere in the module. `interpolate_group` checks before dividing by the unit count:

```python
# booking_full.py:963, MissingType.COST_PER
if units.number != ZERO:
    number_per = (weight - (cost.number_total or ZERO)) / units.number
```

The analogous zero-unit case on the total-PRICE path is also handled gracefully:

```beancount
2024-02-01 * "zero units, total price"
  Assets:S      0 HOOL @@ 20.00 USD
  Assets:Cash  -20.00 USD
```

resolves to `price = 0 USD` and reports `ValidationError: Transaction does not balance: (-20.00 USD)`, no traceback. Line 769 is the spot that misses the equivalent handling.

For reference, rustledger reports `E3001 Transaction does not balance: residual -20.00 USD` on the same input, which seems like the natural verdict here too.

## Suggested fix

Guard the division and emit a booking error for a total cost against zero units, matching what line 963 and the `@@` path already do.

Found while differential-testing cost spec forms between beancount and an alternative implementation.

ryanduguid:
Fixes #1048

`convert_costspec_to_cost` (and `compute_cost_number`) divided a total cost by the unit count with no zero check. `0 HOOL {{20.00 USD}}` — and the `{0 # 20.00 USD}` spelling — therefore escaped `load_file` / `bean-check` as `decimal.DivisionByZero` instead of a booking error.

The `@@` total-price path already survives this input and reports a balance error. The zero-units-at-cost check in `interpolate_group` already knows how to emit `Amount is zero`, but it never ran because conversion crashed first.

This guards both divisions. `compute_cost_number` returns `None` when the per-unit cost is undefined (its documented fallback). `convert_costspec_to_cost` uses a zero unit cost so interpolation can report `Amount is zero: "0 HOOL"`, then validation reports the same residual as the `@@` path.

Verification:

```
pytest beancount/parser/booking_full_test.py
```

121 passed (20 skipped). Includes the existing `TestComputeCostNumber` cases plus both spellings of the #1048 reproducer.

I reviewed, understand, and take responsibility for this change.

### DIFF
--- a/beancount/parser/booking_full.py
+++ b/beancount/parser/booking_full.py
@@ -734,6 +734,9 @@ def compute_cost_number(costspec, units):
         # component involved.
         cost_total = number_total
         units_number = abs(units.number)
+        if units_number == ZERO:
+            # Per-unit cost is undefined when there are no units.
+            return None
         if number_per is not None:
             cost_total += number_per * units_number
         unit_cost = cost_total / units_number
@@ -766,7 +769,12 @@ def convert_costspec_to_cost(posting):
                 cost_total = number_total
                 if number_per is not MISSING:
                     cost_total += number_per * units_number
-                unit_cost = cost_total / units_number
+                # Zero units cannot be converted to a per-unit cost; use
+                # zero so interpolation can report the invalid amount.
+                if units_number == ZERO:
+                    unit_cost = ZERO
+                else:
+                    unit_cost = cost_total / units_number
             else:
                 unit_cost = number_per
             new_cost = Cost(unit_cost, cost.currency, cost.date, cost.label)
--- a/beancount/parser/booking_full_test.py
+++ b/beancount/parser/booking_full_test.py
@@ -1340,6 +1340,52 @@ class TestComputeCostNumber(unittest.TestCase):
             ),
         )
 
+    def test_zero_units_total(self):
+        self.assertIsNone(
+            bf.compute_cost_number(
+                position.CostSpec(None, D("20"), "USD", None, None, False),
+                amount.from_string("0 HOOL"),
+            )
+        )
+
+
+class TestZeroUnitsTotalCost(cmptest.TestCase):
+    @loader.load_doc(expect_errors=True)
+    def test_double_brace_total_cost(self, entries, errors, options_map):
+        """
+        2024-01-01 open Assets:Stock HOOL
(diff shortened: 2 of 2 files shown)


ryanduguid:
The pull-request workflow runs are currently waiting for maintainer approval. Please approve them when convenient so CI can run against this revision.

